PR #23, opened 2021-06-16
rework for go-chi

bigangryrobot:
there seems to be a few reasons to not use iris, here is a chi implementation

https://www.reddit.com/r/golang/comments/b481q7/a_warning_about_githubcomkatarasiris/
https://github.com/the-benchmarker/web-frameworks/issues/1129
https://www.reddit.com/r/golang/comments/57w79c/why_you_really_should_stop_using_iris/


pkg/api/server.go
```diff
@@ -0,0 +1,166 @@
...
+			logger.Errorf("Could not listen on", zap.String("addr", s.opt.APIAddr), zap.Error(err))
+		}
+	}()
+
+	logger.Infof("Server is ready to handle requests", zap.String("addr", s.opt.APIAddr))
+	s.Close()
+}
+
+// Close closes Server.
+func (s *Server) Close() {
```
xxx7xxxx, line 108: We do the signal handling in the `cmd/server/main.go` in one place.

cmd/server/main.go
```diff
@@ -91,8 +91,8 @@ func main() {
 
 	closeCls := func() {
 		wg := &sync.WaitGroup{}
-		wg.Add(2)
-		apiServer.Close(wg)
+		wg.Add(1)
+		apiServer.Close()
```
xxx7xxxx, line 95: Please use the consistent pattern for closing each component.

pkg/api/api.go
```diff
@@ -160,37 +73,37 @@ func (s *Server) RegisterAPIs(apis []*APIEntry) {
...
 		{
 			// https://stackoverflow.com/a/43381061/1705845
-			Path:    "/healthz",
-			Method:  "GET",
-			Handler: func(iris.Context) { /* 200 by default */ },
+			Path:   "/healthz",
+			Method: "GET",
+			Handler: func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "text/plain")
+				w.Write([]byte("ok"))
```
xxx7xxxx, line 105: Please eliminate this line.

xxx7xxxx (commented):
Thanks for your contribution. It's totally OK to use a new web framework to handle admin API. But there's one thing I'm considering. The `chi` looks like a little low level by comparing with echo[1], which tries to be compatible with stdlib. How do you consider it?

[1] https://github.com/labstack/echo

pkg/api/middlewares.go
```diff
@@ -23,70 +23,48 @@ import (
...
+		ww := middleware.NewWrapResponseWriter(w, r.ProtoMajor)
 
-		logger.APIAccess(method, remoteAddr, path, code,
-			bodyBytesReceived, bodyBytesSent,
-			startTime, processTime)
-	}
+		t1 := time.Now()
+		defer func() {
+			logger.APIAccess(r.Method, r.RemoteAddr, r.URL.Path, ww.Status(),
+				r.Response.ContentLength, int64(ww.BytesWritten()),
```
benja-wu, line 37: Directly using `r.Response` will cause panic in runtime. Here is the stack
```bash
 ^[[31mERROR^[[0m        api/middlewares.go:48   recover from /apis/v1/status/members, err: runtim
e error: invalid memory address or nil pointer dereference, stack trace:
goroutine 2603 [running]:
runtime/debug.Stack(0xc00300b650, 0x2035320, 0x2c9c6f0)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x9f
github.com/megaease/easegress/pkg/api.(*Server).newRecoverer.func1.1(0xc000221300, 0x24266a0, 0xc00055a9a0)
        /Users/benja/programmer/github/tmp/easegress/pkg/api/middlewares.go:49 +0xba
panic(0x2035320, 0x2c9c6f0)
        /usr/local/go/src/runtime/panic.go:965 +0x1b9
github.com/megaease/easegress/pkg/api.(*Server).newAPILogger.func1.1(0xc000221300, 0x24389b8, 0xc00020b9c0, 0xc02a7defc8b09818, 0
x317fa75a8a, 0x2cbf420)
        /Users/benja/programmer/github/tmp/easegress/pkg/api/middlewares.go:37 +0xc4
github.com/megaease/easegress/pkg/api.(*Server).newAPILogger.func1(0x24266a0, 0xc00055a9a0, 0xc000221300)
        /Users/benja/programmer/github/tmp/easegress/pkg/api/middlewares.go:41 +0x17b
net/http.HandlerFunc.ServeHTTP(0xc0005c0600, 0x24266a0, 0xc00055a9a0, 0xc000221300)
        /usr/local/go/src/net/http/server.go:2069 +0x44
github.com/megaease/easegress/pkg/api.(*Server).newRecoverer.func1(0x24266a0, 0xc00055a9a0, 0xc000221300)
        /Users/benja/programmer/github/tmp/easegress/pkg/api/middlewares.go:58 +0x83
net/http.HandlerFunc.ServeHTTP(0xc0005c0618, 0x24266a0, 0xc00055a9a0, 0xc000221300)
        /usr/local/go/src/net/http/server.go:2069 +0x44
github.com/megaease/easegress/pkg/api.(*Server).newConfigVersionAttacher.func1(0x24266a0, 0xc00055a9a0, 0xc000221300)
        /Users/benja/programmer/github/tmp/easegress/pkg/api/middlewares.go:68 +0x1f2
net/http.HandlerFunc.ServeHTTP(0xc00063c520, 0x24266a0, 0xc00055a9a0, 0xc000221300)
        /usr/local/go/src/net/http/server.go:2069 +0x44
github.com/go-chi/chi/v5.(*Mux).ServeHTTP(0xc0003dcb40, 0x24266a0, 0xc00055a9a0, 0xc000221200)
        /Users/benja/programmer/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.3/mux.go:88 +0x310
net/http.serverHandler.ServeHTTP(0xc0007c81c0, 0x24266a0, 0xc00055a9a0, 0xc000221200)
        /usr/local/go/src/net/http/server.go:2887 +0xa3
net/http.(*conn).serve(0xc00079ca00, 0x242bb38, 0xc0027830c0)
        /usr/local/go/src/net/http/server.go:1952 +0x8cd
created by net/http.(*Server).Serve
        /usr/local/go/src/net/http/server.go:3013 +0x39b
```
Cause `r.Response` can only be used in the `HTTP client redirect`[1] scenario. 
Plz form a cluster by using the scripts in the `easegress/example` directory and check if the new feature is runnable locally.

# Reference
[1] https://pkg.go.dev/net/http#Request.Response

pkg/api/object.go
```diff
@@ -214,17 +214,18 @@ func (s *Server) listObjects(ctx iris.Context) {
...
-	ctx.Write(buff)
+	w.Header().Set("Content-Type", "text/vnd.yaml")
+
+	w.Write(buff)
 }
 
-func (s *Server) getStatusObject(ctx iris.Context) {
-	name := ctx.Params().Get("name")
+func (s *Server) getStatusObject(w http.ResponseWriter, r *http.Request) {
+	name := r.URL.Query().Get("name")
```
benja-wu, line 223: Actually, the object's name is in the RESTful path's `name` section.  Using `r.URL.Queyr()` can **only** get the parameters from URL after the `?` and separated by `&` token. [1] It you test the `get/update` API locally, you will get a `404` return. 

**This problem exists in every RESTful APIs which get `name` or another parameter from the URL path via a placeholder.** Plz fix them. 

## Reference
[1] https://pkg.go.dev/net/url#URL.Query

pkg/api/object.go
```diff
@@ -20,11 +20,11 @@ package api
 import (
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"sort"
 
 	"github.com/megaease/easegress/pkg/supervisor"
 
-	"github.com/kataras/iris"
```
benja-wu, line 27: Inline #63, #68, #78..., the ` ObjectPrefix + "/{name:string}"` this iris path grammar won't work in go-chi. Plz read the go-chi's example in [README](https://github.com/go-chi/chi/blob/master/README.md) carefully!

It should be `ObjectPrefix +"/{name}". The other RESTful APIs in this directory or in `mesh controller`  also need to be fixed.

pkg/api/object.go
bigangryrobot: ah i missed a few of these, i realized midway though writing this that i wrote them wrong and tried to correct haha. ok will take care of this

pkg/api/middlewares.go
```diff
@@ -34,7 +34,7 @@ func (s *Server) newAPILogger(next http.Handler) http.Handler {
 		t1 := time.Now()
 		defer func() {
 			logger.APIAccess(r.Method, r.RemoteAddr, r.URL.Path, ww.Status(),
-				r.Response.ContentLength, int64(ww.BytesWritten()),
+				0, int64(ww.BytesWritten()),
```
benja-wu, line 37: In `logger.APIAccess` function, the fifth parameter is `bodyBytedReceived`.  So the original implementation, this parameter is valued with `ctx.ContentLgnth` which returns the request's header value of "Content-Length".[1]
So I think using `r.ContentLength` here is definitely better than just valued it with `0`. 

## Reference
[1] https://github.com/megaease/easegress/blob/main/pkg/api/middlewares.go#L53

bigangryrobot: ya, i couldnt yet find a way to grab that without blowing things up. Let me take a stab at it again

xxx7xxxx (approved):

benja-wu (approved):
LGTM

benja-wu:
> its worth mentioning that I did roll through 80% of adding the echo framework as well, just to see how it went and it was a bit harder to implement due mainly to how the easegress framework is currently structured. It might be something to return to after some time.

@bigangryrobot  Thank you so much for contributing this PR and providing a performance quick look. The `echo` is also a cool web framework but I think `go-chi` looks good enough for Easegress API management currently. 

Feel free to open an issue about design discussions or new feature proposals.

xxx7xxxx:
Sure, I just think echo got higher abstraction than chi in some ways. But chi is fine in this simple scenario. Thanks for your contribution!

PR #39, opened 2021-06-18
Fix: main package redeclaration in backend-service dir

Loyalsoldier:


benja-wu:
Well, actually, in the `demoing/testing` scenario, we prefer to use `go run example/backend-service/mirror.go` or `go run example/backend-service/remote.go` directly for dedicated demo/test case. So we don't need to make the `backend-service` dir compilable. 
Still thanks for your concern and contribution. But I will close this PR as it's not necessary for our `demoing/testing` secnario.

Loyalsoldier:
If not to change this, running `go vet ./...` will be failed with `main package redeclaration` error.

benja-wu:
Ok, I misunderstood your purpose here. Let me reopen and review it. 

And maybe we can add more description in the PR comment to avoid such kind of communication problem. Thx :-)

benja-wu (approved):
LGTM

PR #45, opened 2021-06-18
Docker: add tzdata package for latest alpine

Loyalsoldier:


build/package/Dockerfile
```diff
@@ -8,14 +8,14 @@ RUN make clean && make build
 
 # ---
 
-FROM alpine:3.13
+FROM alpine:latest
```
xxx7xxxx, line 11: latest sounds cool, but can it guarantee stability?

Loyalsoldier: The release strategy of Alpine Linux seems promising: https://alpinelinux.org/releases/

bigangryrobot: latest should be fine for now for our needs especially using a built golang executable. If we were using python or java on the other hand, beyond me running for the hills, we would have a much larger dependency on the libs provided by the base image

xxx7xxxx (approved):

PR #79, opened 2021-06-28
Upgrade MeshController to use TrafficController

xxx7xxxx:


benja-wu (approved):
Mesh related logic had been tested in KM env, LGTM

pkg/object/meshcontroller/meshcontroller.go
```diff
@@ -76,15 +79,18 @@ func (mc *MeshController) DefaultSpec() interface{} {
...
 
 // Inherit inherits previous generation of MeshController.
-func (mc *MeshController) Inherit(spec *supervisor.Spec,
+func (mc *MeshController) Inherit(superSpec *supervisor.Spec,
 	previousGeneration supervisor.Object, super *supervisor.Supervisor) {
 
-	previousGeneration.Close()
-	mc.Init(spec, super)
+	mc.superSpec, mc.spec, mc.super = superSpec, superSpec.ObjectSpec().(*spec.Admin), super
+	mc.reload()
```
localvar, line 93: `mc.api` is not set as in function `Init`.

pkg/object/meshcontroller/worker/worker.go
```diff
@@ -119,23 +116,25 @@ func New(superSpec *supervisor.Spec, super *supervisor.Supervisor) *Worker {
...
 	store := storage.New(superSpec.Name(), super.Cluster())
-	_service := service.New(superSpec, store)
+	_service := service.New(superSpec, super)
 	registryCenterServer := registrycenter.NewRegistryCenterServer(spec.RegistryType,
 		serviceName, applicationIP, applicationPort, instanceID, serviceLabels, _service)
-	ingressServer := NewIngressServer(superSpec, super, serviceName)
-	egressEvent := make(chan string, egressEventChanSize)
-	egressServer := NewEgressServer(superSpec, super, serviceName, _service, egressEvent)
+
+	// FIXME: check service Name
```
localvar, line 123: remove if serviceName is correct

pkg/object/meshcontroller/worker/worker.go
```diff
@@ -362,115 +359,45 @@ func (wrk *Worker) updateHearbeat() error {
...
-
-	wrk.informer.Close()
-	wrk.registryServer.Close()
-	wrk.apiServer.Close()
+func (worker *Worker) Close() {
+	close(worker.done)
+
+	worker.egressServer.Close()
+	worker.ingressServer.Close()
+	worker.informer.Close()
```
localvar, line 400: `informer` should be closed first, this is to ensure no notifications could be sent after `egressServer` & `ingressServer` were closed, otherwise, they will be re-initialized.

also need to update informer implementation, to make sure no notification be sent after it is closed.

pkg/object/meshcontroller/worker/worker.go
benja-wu: Ok

pkg/object/meshcontroller/worker/worker.go
benja-wu: The `Informoer`'s implementation update I recommend to put it into the next PR. 
And let me update the sequence of `close` callings.

pkg/object/meshcontroller/meshcontroller.go
benja-wu: @xxx7xxxx ， Is that necessary to introduce the `api` into MeshController?  Since `master` and `worker` have handled the API registration inside. And I can't find any use of `m.api` in MeshController.

xxx7xxxx: It's expected. Otherwise it will cause repeating registering apis.

localvar: if so, in `Init`, can we change
```go
mc.api = api.New(superSpec, super)
```
to
```go
api.New(superSpec, super)
```
and remove `mc.api`?

it is weird to only initialize the variable in `Init`.

pkg/object/meshcontroller/worker/worker.go
localvar: I've made the changes to `Informer`, but I agree with you to hold it on to the next PR, as it introduces new `wait` operations and may result in deadlocks.
I'll create an issue to track this.

pkg/object/meshcontroller/meshcontroller.go
xxx7xxxx: I don't see the difference. IMO, the standalone `api.New` is weirder to me. I updated it to `api.Register`

localvar (approved):

PR #96, opened 2021-06-30
Support dynamic admin API && correct syncer && make interface cleaner

xxx7xxxx:
- Support dynamic(register/unregister) mux in admin api
- Don't send data if not change in syncer(leverage it in supervisor)
- Make the interface Object&Filter cleaner
- Fix some minor bugs


localvar:
There's a potential issue with the new behavior of `syncer` in this PR, take the mesh ingress as an example:

When receiving the resource update event,  the mesh ingress controller tasks it just as a notification and ignores data in the event, it then reloads all data itself and rebuilds all data structures, and it uses the old data if it fails to reload data or rebuild data structures for some reason. So, if the `syncer` only sends an event at a resource change, there's a possibility that the latest data can be applied in time.

xxx7xxxx:
>  it uses the old data if it fails to reload data or rebuild data structures for some reason.

Acceptable if the cluster failed is the reason.

> So, if the syncer only sends an event at a resource change, there's a possibility that the latest data can be applied in time.

Do you mean `not in time`. If so, it's also acceptable.

localvar:
> Do you mean `not in time`. If so, it's also acceptable.

It should be "cannot be applied in time", sorry for the typo.

localvar (approved):

benja-wu (approved):
LGTM

PR #100, opened 2021-07-02
Use default spec coorectly & fix httppipeline.Validate without recursively checking

xxx7xxxx:
A bugfix next to https://github.com/megaease/easegress/pull/96. 

It fixed:

- object&filter won't use default spec to cover empty fields.
- HTTPPipeline.Validate won't recursively check subfields.

But here left an issue:

`required` string type in jsonschema allow empty string[1], for example, the field `Name` in `ratelimiter.Policy` can be an empty string. We need to fix it.

[1] https://stackoverflow.com/a/54426110/1705845

### DIFF
--- a/pkg/object/httppipeline/httppipeline.go
+++ b/pkg/object/httppipeline/httppipeline.go
@@ -216,7 +216,6 @@ func (meta *FilterMetaSpec) Validate() error {
 // Validate validates Spec.
 func (s Spec) Validate() (err error) {
 	errPrefix := "filters"
-	_ = errPrefix
 	defer func() {
 		if r := recover(); r != nil {
 			err = fmt.Errorf("%s: %s", errPrefix, r)
--- a/pkg/object/httppipeline/httppipeline.go
+++ b/pkg/object/httppipeline/httppipeline.go
@@ -24,13 +24,12 @@ import (
 	"sync"
 	"time"
 
-	yaml "gopkg.in/yaml.v2"
-
 	"github.com/megaease/easegress/pkg/context"
 	"github.com/megaease/easegress/pkg/logger"
 	"github.com/megaease/easegress/pkg/protocol"
 	"github.com/megaease/easegress/pkg/supervisor"
 	"github.com/megaease/easegress/pkg/util/stringtool"
+	"github.com/megaease/easegress/pkg/util/yamltool"
 )
 
 const (
@@ -69,7 +68,7 @@ type (
 	// Spec describes the HTTPPipeline.
 	Spec struct {
 		Flow    []Flow                   `yaml:"flow" jsonschema:"omitempty"`
-		Filters []map[string]interface{} `yaml:"filters" jsonschema:"-"`
+		Filters []map[string]interface{} `yaml:"filters" jsonschema:"required"`
 	}
 
 	// Flow controls the flow of pipeline.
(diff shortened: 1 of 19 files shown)


pkg/object/meshcontroller/spec/spec.go
```diff
@@ -85,10 +85,10 @@ type (
 		RegistryType string `yaml:"registryType" jsonschema:"required"`
 
 		// APIPort is the port for worker's API server
-		APIPort int `yaml:"apiPort" jsonschema:"omitempty"`
+		APIPort int `yaml:"apiPort" jsonschema:"required"`
 
 		// IngressPort is the port for http server in mesh ingress
-		IngressPort int `yaml:"ingressPort" jsonschema:"omitempty"`
+		IngressPort int `yaml:"ingressPort" jsonschema:"required"`
```
localvar, line 91: this port should be optional as it is only used in mesh ingress controller

pkg/object/httppipeline/httppipeline.go
```diff
@@ -69,7 +68,7 @@ type (
 	// Spec describes the HTTPPipeline.
 	Spec struct {
 		Flow    []Flow                   `yaml:"flow" jsonschema:"omitempty"`
-		Filters []map[string]interface{} `yaml:"filters" jsonschema:"-"`
+		Filters []map[string]interface{} `yaml:"filters" jsonschema:"required"`
```
localvar, line 71: `flow` should be `required` as `filters` is `required`

pkg/object/meshcontroller/spec/spec.go
xxx7xxxx: Yes, but the `Admin` spec if for the MeshController, so every instance got the same spec.

pkg/object/httppipeline/httppipeline.go
xxx7xxxx: Nope, `flow` is allowed to be empty, which cause the all `filters` will be a normal pipeline without jumping.

pkg/util/yamltool/yamltool.go
```diff
@@ -0,0 +1,23 @@
...
+
+func Marshal(in interface{}) []byte {
+	buff, err := yaml.Marshal(in)
+	if err != nil {
+		panic(fmt.Errorf("marshal %s to yaml string failed: %v", in, err))
+	}
+	return buff
+}
+
+func Unmarshal(in []byte, out interface{}) {
```
benja-wu, line 17: Proposal to return an error instead of using `panic`, I don't think using `panic` in a `util` usage package is a good idea.  The `util` package will be widely used in the project and a toolkit purpose function calls should not cause an `application immediately crash` consequence when the caller forgets to use `recover`[1]. 

## Reference
1. https://stackoverflow.com/questions/44504354/should-i-use-panic-or-return-error/44505268

pkg/object/httppipeline/httppipeline.go
```diff
@@ -230,17 +214,20 @@ func (meta *FilterMetaSpec) Validate() error {
 }
 
 // Validate validates Spec.
-func (s Spec) Validate(config []byte) (err error) {
+func (s Spec) Validate() (err error) {
 	errPrefix := "filters"
+	_ = errPrefix
```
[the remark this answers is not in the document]
xxx7xxxx: redundant one line, deleted.

pkg/util/yamltool/yamltool.go
xxx7xxxx: Here's the reason:

This utility aims to fix so many time calling `marshal && if err != nil` and `unmarshal && err !=  nil` in httppipeline/spec.go, httppipeline/httppipeline.go and supervisor/spec.go.  It saves a lot of repeated code.

And I've added comments to them, can you accept it after that?

benja-wu: * I got it. Then propose to add the panic behavior comment for `yamltool. Marshal`, `yamltool. Unmarshal` as well. To make sure the caller is able to read this designed behavior through IDE's remind.
* I believe it's the caller's duty to read the toolkit functions' usage comment carefully before using it.

xxx7xxxx: Hmm, I had done it.

benja-wu: Hmmm, can't see the `Outdated` tag in this GitHub conversation card. (So I thought u haven't updated this file)
My bad.

benja-wu (approved):
LGTM

localvar (approved):

PR #105, opened 2021-07-05
Add missing `return err` after `if err == <ErrorType>` statement.

vimfun:
If missing `return err` after `if` statement, some errors will be ignored. It will cause hidden the reason of `err`. That is a bad practice.


benja-wu (approved):
LGTM

benja-wu:
@vimfun 
* After @localvar‘s reminding and confirming the former logic,  this PR should not be merged. Because the `AlreadyWatched` type error is the exception error type which is not needed to be handle by the caller. So the former logic is correct.
*  We will add some comments to clarify this behavior and reserves this PR.

Thank you for opening this PR as well.

PR #108, opened 2021-07-06
[v]: Use newer json schema generator & fix spec test failure

xxx7xxxx:
The failure reason of unit test of mesh spec path are:

1. `github.com/megaease/jsonschema` is outdated
2. There was an outdated config field in the proxy filter

But the ecosystem of json schema in golang is poor, there left an issue(not a problem for now) which is `github.com/alecthomas/jsonschema` can't generate the correct json schema with `$ref` and `definitions`. I put it in the code comment.


localvar (approved):

benja-wu (approved):
LGTM

Issue #114, opened 2021-07-10
http server cannot accept any connection


localvar:
Steps to reproduce:

1. Setup a http server and a pipeline
2. Send request to the http server

zhao-kun:
Was it introduced by #110 ?

localvar:
> Was it introduced by #110 ?

Yes, found after a git rebase in the development of wasm.

PR #115, opened 2021-07-10
fix bug #114

localvar:


pkg/object/httpserver/listen.go
```diff
@@ -78,7 +84,7 @@ func (l *LimitListener) SetMaxConnection(n uint32) {
 // Close closes LimitListener.
```
xxx7xxxx, line 84: Comment for the line 81: Do we need to monitor the returning channel of `l.sem.SetMaxCount` or not? If not, please eliminate it.

localvar: It is not monitored currently.
But I propose to keep it as `l.sem.SetMaxCount` is an async operation, and `sem` is a utility package, it may be useful later.

xxx7xxxx: If so, please comment it because I got confused.

localvar: updated.

benja-wu: I agree with keeping this `done` chan, cause in some scenarios, the user may want to make sure the acquirement is done successfully.

pkg/object/httpserver/listen.go
```diff
@@ -27,46 +28,51 @@ import (
 // NewLimitListener returns a Listener that accepts at most n simultaneous
 // connections from the provided Listener.
 func NewLimitListener(l net.Listener, n uint32) *LimitListener {
+	ctx, cancel := context.WithCancel(context.Background())
```
benja-wu, line 31: The `context.Background()` will return a none-deadline, forever waiting's context, should we limit this wait time here?

xxx7xxxx: It's a canceling function do the similar work as `done`

xxx7xxxx (approved):

pkg/object/httpserver/listen.go
localvar: No, we don't need to do this, the `ctx` is only to ensure `sem.Acquire` could be cancelled at the close of the listener.

benja-wu: * Yes, the `cancel function does close this context, but one client request may be hanged at the connection stage forever if it doesn't close this connection manually and firstly (If it can't acquire sem successfully due to poor luck).
Although the HTTPServer can eventually close this connection when it shutdown.

* This HTTPSever behavior design can be continued, I will figure it out at another new issue if necessary.

benja-wu (approved):
LGTM

PR #116, opened 2021-07-10
add filter WasmHost (close #1)

### LINKED ISSUE #1: WASM:The new plugin mechanism for the Easegress
### Backgroud

Our flagship product Easegress(EaseGateway) has many splendid features, it is fascinating, especially our `pipeline/plugin` mechanism which empower customers to achieve their specific goal with the Easegress customizing way
But the current `pipeline/plugin` mechanism still has too many barriers to use  If a user really wants to extend the Easegress he needs to conquer the following issues:

1. Master the Go language
2. Master and understand low level `pipeline/plugin` mechanism
3. Commit changes to the Easegress repository and rebuild the Easegress server
4. Deploy Easegress, need to reboot the  Easegress server 

I think the last two of these barriers are the biggest obstacles for users to extend the  Easegress. So I think we need another `pipeline/plugin` mechanism for the EG customization.

### Goal

Compare with other gateway productions, we can found they are all choosing a solution that is embedding a weak language to enhance the power of extensibility.  but there are serval cons in these implementations.

- Weak language (general the embedded language is the Lua), it's difficult to solve complicated business logic in users' scenarios
- Performance penalty, as the Lua is a lightweight interpreted programming language， although the Lua introduces the JIT mechanism, it can't be high performance. 

If we want to provide a more flexible customization mechanism, we must solve the above disadvantages.

### Proposal

After several days of study, I found we can leverage WebAssembly to solve the above problems.(被打脸了……), because the WebAssembly has the following feature:

- Near-native performance.
- Wasm expands to any language.
- Wasm runs in an isolated VM and can dynamically update without EG restarts.
- Eliminate the need to recompile and maintain a build of EG.

Golang has rich ecology, I found an open-source Golang WebAssembly runtime library at [1]. 

> PS: I don't want to deprecate the current pipeline/plugin mechanism, but I think we need multiple customized abstraction, the different way to process the different scene.  This solution has been adopted by Envoy as its filter's extensibility [2]. 

[1] https://github.com/wasmerio/wasmer-go
[2] https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/wasm-cc

localvar:
add the `WasmHost` filter, it is disabled by default in the build because of the requirements of CGo, and can be enabled by the build command below:

```bash
$ make GOTAGS=wasmhost
```

and this PR only includes the filter itself, we still need to develop SDKs for different languages to empower users to write their own code for the filter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,8 @@ ENABLE_CGO= CGO_ENABLED=0
 # Check Go build tags, the tags are from command line of make
 ifdef GOTAGS
   GO_BUILD_TAGS= -tags ${GOTAGS}
-  # Must enable Cgo when wasmfilter is included
-  ifeq ($(findstring wasmfilter,${GOTAGS}), wasmfilter)
+  # Must enable Cgo when wasmhost is included
+  ifeq ($(findstring wasmhost,${GOTAGS}), wasmhost)
 	ENABLE_CGO= CGO_ENABLED=1
   endif
 endif
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,18 @@ endif
 # Build Flags
 GO_LD_FLAGS= "-s -w -X github.com/megaease/easegress/pkg/version.RELEASE=${RELEASE} -X github.com/megaease/easegress/pkg/version.COMMIT=${GIT_COMMIT} -X github.com/megaease/easegress/pkg/version.REPO=${GIT_REPO_INFO}"
 
+# Cgo is disabled by default
+ENABLE_CGO= CGO_ENABLED=0
+
+# Check Go build tags, the tags are from command line of make
+ifdef GOTAGS
+  GO_BUILD_TAGS= -tags ${GOTAGS}
+  # Must enable Cgo when wasmfilter is included
+  ifeq ($(findstring wasmfilter,${GOTAGS}), wasmfilter)
+	ENABLE_CGO= CGO_ENABLED=1
+  endif
+endif
+
 # Targets
 TARGET_SERVER=${RELEASE_DIR}/easegress-server
 TARGET_CLIENT=${RELEASE_DIR}/egctl
@@ -38,7 +50,7 @@ build_client:
 build_server:
 	@echo "build server"
 	cd ${MKFILE_DIR} && \
-	CGO_ENABLED=0 go build -v -trimpath -ldflags ${GO_LD_FLAGS} \
+	${ENABLE_CGO} go build ${GO_BUILD_TAGS} -v -trimpath -ldflags ${GO_LD_FLAGS} \
 	-o ${TARGET_SERVER} ${MKFILE_DIR}cmd/server
 
 dev_build: dev_build_client dev_build_server
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -48,6 +48,8 @@ var (
 	apisMutex      = sync.Mutex{}
 	apis           = make(map[string]*APIGroup)
 	apisChangeChan = make(chan struct{}, 10)
+
+	appendAddonAPIs []func(s *Server, group *APIGroup)
 )
 
 type apisbyOrder []*APIGroup
@@ -97,6 +99,10 @@ func (s *Server) registerAPIs() {
 	group.Entries = append(group.Entries, s.metadataAPIEntries()...)
 	group.Entries = append(group.Entries, s.healthAPIEntries()...)
 	group.Entries = append(group.Entries, s.aboutAPIEntries()...)
+
+	for _, fn := range appendAddonAPIs {
+		fn(s, group)
+	}
 
 	RegisterAPIs(group)
 }
--- a/pkg/filter/wasm/hostfunc.go
+++ b/pkg/filter/wasm/hostfunc.go
@@ -35,8 +35,10 @@ import (
 ////////////////////////////////////////////////////////////////////////////////
 // helper functions
 
+const wasmMemory = "memory"
+
 func (vm *WasmVM) readDataFromWasm(addr int32) []byte {
-	mem := vm.inst.GetExport(vm.store, "memory").Memory().UnsafeData(vm.store)
+	mem := vm.inst.GetExport(vm.store, wasmMemory).Memory().UnsafeData(vm.store)
 	reader := bytes.NewReader(mem[addr:])
 	var size int32
 	if e := binary.Read(reader, binary.LittleEndian, &size); e != nil {
@@ -52,7 +54,7 @@ func (vm *WasmVM) readDataFromWasm(addr int32) []byte {
 }
 
 func (vm *WasmVM) writeDataToWasm(data []byte) int32 {
-	mem := vm.inst.GetExport(vm.store, "memory").Memory().UnsafeData(vm.store)
+	mem := vm.inst.GetExport(vm.store, wasmMemory).Memory().UnsafeData(vm.store)
 	vaddr, e := vm.fnAlloc.Call(vm.store, len(data)+4)
 	if e != nil {
 		panic(e)
@@ -67,7 +69,7 @@ func (vm *WasmVM) writeDataToWasm(data []byte) int32 {
 }
 
 func (vm *WasmVM) readStringFromWasm(addr int32) string {
-	mem := vm.inst.GetExport(vm.store, "memory").Memory().UnsafeData(vm.store)
+	mem := vm.inst.GetExport(vm.store, wasmMemory).Memory().UnsafeData(vm.store)
 
 	start := addr
 	for mem[addr] != 0 {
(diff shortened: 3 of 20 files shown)


pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+		syncer, err = c.Syncer(time.Minute)
+		if err == nil {
+			chWasm, err = syncer.Sync(c.Layout().WasmCodeEvent())
+			if err == nil {
+				break
+			}
+		}
+		logger.Errorf("failed to watch Wasm code event: %v", err)
+		select {
+		case <-time.After(10 * time.Second):
```
benja-wu, line 189: Need to add `continue` here?

pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+func (f *WasmFilter) readWasmCode() ([]byte, error) {
+	if isURL(f.spec.Code) {
+		return readWasmCodeFromURL(f.spec.Code)
+	}
+	if _, e := os.Stat(f.spec.Code); e == nil {
+		return readWasmCodeFromFile(f.spec.Code)
+	}
+	return base64.StdEncoding.DecodeString(f.spec.Code)
+}
+
```
benja-wu, line 148: So we don't support providing an API for storing the user's WASM code into Easegress' Etcd, Right? 
Propose to support this feature in the future version.

pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+	var pool *WasmVMPool
+	if p := f.vmPool.Load(); p == nil {
+		ctx.AddTag("Wasm VM pool is not initialized")
+		return resultOutOfVM
+	} else {
+		pool = p.(*WasmVMPool)
+	}
+
+	// get a free WASM VM and attach the ctx to it
+	vm := pool.Get()
```
benja-wu, line 251: `pool.Get()` may create a new VM by calling `newWasmVM` which will invoke `panic` if something goes wrong. So we may also need to use `recover` here if we want to avoid the whole request cllapsed.

pkg/api/api.go
```diff
@@ -98,6 +100,10 @@ func (s *Server) registerAPIs() {
 	group.Entries = append(group.Entries, s.healthAPIEntries()...)
 	group.Entries = append(group.Entries, s.aboutAPIEntries()...)
 
+	for _, fn := range appendAddonAPIs {
```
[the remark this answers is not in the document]
localvar: wasm requires cgo and add about 20MB to the binary, so I make it an optional feature, `RegisterAPIs` does not fit this case.

pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+		return nil, e
+	}
+	defer resp.Body.Close()
+	return io.ReadAll(resp.Body)
+}
+
+func isURL(str string) bool {
+	// we only check the first a few bytes as str could be
+	// the LONG base64 encoded wasm code
+	for _, p := range []string{"http://", "https://"} {
```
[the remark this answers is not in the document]
localvar: if `str` is the base64 encoded wasm code, it can be very long, but we only need to check the first a few bytes.

pkg/filter/wasm/wasm.go
localvar: Yes, we support that, `f.spec.Code` could be a file path, an URL, or the base64 encoded wasm code, the filter detects 
 its type automatically.

pkg/filter/wasm/wasm.go
localvar: no, it won't fall through to the next `case`.

pkg/filter/wasm/wasm.go
localvar: where the `panic` come from?
if it comes from `importHostFuncs`, then that's fine because it should never happen and must be fixed at the development phase.

pkg/filter/wasm/wasm.go
benja-wu: Great! Seems that I had missed the base64 encoded wasm code in the spec solution.

pkg/filter/wasm/wasm.go
benja-wu: Yes, such as `writeStringToWasm`'s `memory allocate panic`. Since we had discussed the background clearly, it's only a registration behavior instead of actual code execution, I am Ok with it.

pkg/api/api.go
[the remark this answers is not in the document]
benja-wu: My concern is reuse the `RegisterAPIs/UnRegisterAPIs` for a more clean api package.(without adding other business logic file into this pkg)
But WASM feature is implemented as a filter instread of controller, current solution seems acceptable to me.

pkg/filter/wasm/wasm.go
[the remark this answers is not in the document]
localvar: > it will try to split the URL with `#`

this is the case, `#` is not a valid character in base64 encoding, so the split has to go through the whole string before it finds out that the string does need a split.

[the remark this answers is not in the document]
benja-wu: Yes, if we care about the parsing cost especially here, your solution's performance is better than the standard lib.

benja-wu (approved):
LGTM

Makefile
```diff
@@ -22,6 +22,18 @@ endif
 # Build Flags
 GO_LD_FLAGS= "-s -w -X github.com/megaease/easegress/pkg/version.RELEASE=${RELEASE} -X github.com/megaease/easegress/pkg/version.COMMIT=${GIT_COMMIT} -X github.com/megaease/easegress/pkg/version.REPO=${GIT_REPO_INFO}"
 
+# Cgo is disabled by default
+ENABLE_CGO= CGO_ENABLED=0
+
+# Check Go build tags, the tags are from command line of make
+ifdef GOTAGS
```
xxx7xxxx, line 29: Actually, I don't think it's a big deal to include more than 20MB in EG with wasm. I think it brings more concern in compilation, which gives more confusing stuff to developers, operators.

pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+
+	"github.com/megaease/easegress/pkg/cluster"
+	"github.com/megaease/easegress/pkg/context"
+	"github.com/megaease/easegress/pkg/logger"
+	"github.com/megaease/easegress/pkg/object/httppipeline"
+)
+
+const (
+	// Kind is the kind of WasmFilter.
+	Kind        = "WasmFilter"
```
xxx7xxxx, line 42: There is a filter naming `WasmFilter` that sounds weird/redundant to me, can we rename it to `WasmRuntime` or something else.

pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+	return "WasmFilter implements a filter which runs web assembly"
+}
+
+// Results returns the results of WasmFilter.
+func (f *WasmFilter) Results() []string {
+	return results
+}
+
+func readWasmCodeFromFile(path string) ([]byte, error) {
+	f, e := os.Open(path)
```
xxx7xxxx, line 108: https://pkg.go.dev/os#ReadFile could replace the 2 system call.

pkg/filter/wasm/wasm.go
```diff
@@ -0,0 +1,317 @@
...
+	if !ok || code > maxWasmCode {
+		panic(fmt.Errorf("invalid Wasm result: %v", r))
+	}
+
+	return wasmCodeToResult(code)
+}
+
+// Status returns Status genreated by Runtime.
+func (f *WasmFilter) Status() interface{} {
+	return nil
```
xxx7xxxx, line 311: Do we really have nothing to report?

pkg/filter/wasm/hostfunc.go
```diff
@@ -0,0 +1,449 @@
...
+	"io"
+	"net/http"
+	"net/textproto"
+	"strings"
+
+	"github.com/bytecodealliance/wasmtime-go"
+	"github.com/megaease/easegress/pkg/logger"
+)
+
+////////////////////////////////////////////////////////////////////////////////
```
xxx7xxxx, line 35: hmm, please eliminate this line

pkg/filter/wasm/wasm.go
localvar: wasmtime and wasmer are wasm runtimes, so `WasmRuntime` is not a suitable name for the filter, in fact, I tried to find another one, but none seems better.

benja-wu: `WasmVM `？

localvar: there's `WasmVM` already in our code...

pkg/filter/wasm/wasm.go
localvar: updated.

pkg/filter/wasm/wasm.go
xxx7xxxx: How about `Wasm`

localvar: `Wasm` often refers to many other things like the wasm code and may bring additional confusion.
I'll rename it to `WasmHost` as it is a host environment of WebAssembly, any idea?

xxx7xxxx: I'm fine with it

xxx7xxxx (approved):

PR #124, opened 2021-07-16
golint warning fixes

haoel:
major fix the following golint warnings:

- The incorrect comments
- Convert ALL_CAPS variable name to Camels style
- Inconsistent of receiver name
- Stuttering struct name - `IngressController.IngressControllerSpec` -> `IngressController.Spec`
- If block ends with a return statement, so drop this else and outdent its block. Refer to golang/lint#17
- Typo of the variable name. `protocal` -> `protocol`
- Rename some variables. `HttpClient`  -> `HTTPClient`,  `JsonToKVMap` -> `JSONToKVMap`

### DIFF
--- a/pkg/object/meshcontroller/service/service.go
+++ b/pkg/object/meshcontroller/service/service.go
@@ -60,7 +60,7 @@ func (s *Service) Lock() {
 	}
 }
 
-// Unlock - unlocks all store, it will do cluster panic if failed.
+// Unlock unlocks all store, it will do cluster panic if failed.
 func (s *Service) Unlock() {
 	err := s.store.Unlock()
 	if err != nil {
--- a/pkg/object/meshcontroller/service/service.go
+++ b/pkg/object/meshcontroller/service/service.go
@@ -52,7 +52,7 @@ func New(superSpec *supervisor.Spec) *Service {
 	return s
 }
 
-// Lock - locks all store, it will do cluster panic if failed.
+// Lock locks all store, it will do cluster panic if failed.
 func (s *Service) Lock() {
 	err := s.store.Lock()
 	if err != nil {
--- a/pkg/object/meshcontroller/service/service.go
+++ b/pkg/object/meshcontroller/service/service.go
@@ -52,7 +52,7 @@ func New(superSpec *supervisor.Spec) *Service {
 	return s
 }
 
-// Lock locks all store, it will do cluster panic if failed.
+// Lock - locks all store, it will do cluster panic if failed.
 func (s *Service) Lock() {
 	err := s.store.Lock()
 	if err != nil {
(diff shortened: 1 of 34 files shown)


pkg/object/trafficcontroller/trafficcontroller.go
```diff
@@ -100,7 +100,7 @@ func newNamespace(namespace string) *Namespace {
 	}
 }
 
-// Space gets handler within the namspace of it.
+// GetHandler - Space gets handler within the namspace of it.
```
localvar, line 103:
```suggestion
// GetHandler gets handler within the namespace
```

pkg/object/meshcontroller/service/service.go
```diff
@@ -52,15 +52,15 @@ func New(superSpec *supervisor.Spec) *Service {
 	return s
 }
 
-// Lock locks all store, it will do cluster panic if failed.
+// Lock - locks all store, it will do cluster panic if failed.
```
localvar, line 55: why add a `-`?

pkg/object/meshcontroller/service/service.go
```diff
@@ -52,15 +52,15 @@ func New(superSpec *supervisor.Spec) *Service {
...
 func (s *Service) Lock() {
 	err := s.store.Lock()
 	if err != nil {
 		api.ClusterPanic(err)
 	}
 }
 
-// Lock unlocks all store, it will do cluster panic if failed.
+// Unlock - unlocks all store, it will do cluster panic if failed.
```
localvar, line 63: Ditto

pkg/object/meshcontroller/service/service.go
haoel: golint requires the comment must be lead by function name. because Lock and Unlock is a pair function. And the `-` just separate the function name and the comments. otherwise, it seems a bit stuttering.

localvar: The official go source code for `Mutex.Lock` & `Mutex.Unlock` doesn't have it.

```go
// Lock locks m.
// If the lock is already in use, the calling goroutine
// blocks until the mutex is available.
func (m *Mutex) Lock() {


// Unlock unlocks m.
// It is a run-time error if m is not locked on entry to Unlock.
//
// A locked Mutex is not associated with a particular goroutine.
// It is allowed for one goroutine to lock a Mutex and then
// arrange for another goroutine to unlock it.
func (m *Mutex) Unlock() {

```

localvar:
```suggestion
// Lock locks all store, it will do cluster panic if failed.
```

pkg/object/meshcontroller/service/service.go
localvar:
```suggestion
// Unlock unlocks all store, it will do cluster panic if failed.
```

localvar (approved):

xxx7xxxx (approved):

PR #133, opened 2021-07-18
supply some unit tests to increase the code coverage

haoel:
The unit tests for 

- pkg/util/ratelimiter
- pkg/util/circuitbreaker
- pkg/util/agent_controller
- pkg/util/texttemplate
- pkg/common/utils
- pkg/common/strings
- pkg/common/options
- pkg/common/time
- pkg/common/stat_aggregator
- pkg/common/go
- pkg/common/callback

### DIFF
--- a/pkg/common/option_test.go
+++ b/pkg/common/option_test.go
@@ -18,128 +18,186 @@
 package common
 
 import (
+	"math"
 	"regexp"
+	"strconv"
 	"testing"
 )
 
 func TestUint8(t *testing.T) {
 	var u8 uint8
-	u := NewUint8Value(128, &u8)
-	if u8 != 128 {
-		t.Errorf("expected %d, result %d", 128, u8)
+	u := NewUint8Value(math.MaxUint8, &u8)
+	if u8 != math.MaxUint8 {
+		t.Errorf("expected %d, result %d", math.MaxUint8, u8)
 	}
 
-	str := "100"
-	u8 = 100
+	var x uint8 = 100
+	u = NewUint8Value(x, nil)
+	if u.Get() != x {
+		t.Errorf("expected %d, result %d", x, u.Get())
+	}
+
+	x = 200
+	str := strconv.FormatUint(uint64(x), 10)
 	u.Set(str)
-	if u.Get() != uint8(u8) {
-		t.Errorf("expected %d, result %d", u8, u.Get())
+	if u.Get() != x {
+		t.Errorf("expected %d, result %d", x, u.Get())
 	}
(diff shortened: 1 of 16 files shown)


benja-wu (approved):
LGTM

pkg/common/option_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+func TestUint8(t *testing.T) {
+	var u8 uint8
+	u := NewUint8Value(128, &u8)
+	if u8 != 128 {
+		t.Errorf("expected %d, result %d", 128, u8)
+	}
+
+	str := "100"
+	u8 = 100
```
localvar, line 33:
```suggestion
	u8 = 200
```

pkg/util/circuitbreaker/circuitbreaker.go
```diff
@@ -307,20 +307,27 @@ var stateStrings = []string{
 }
 
 // NewPolicy create and initialize a policy with default configuration
```
localvar, line 309:
```suggestion
// NewPolicy create and initialize a policy
```

pkg/util/jmxtool/agent_controller_test.go
```diff
@@ -0,0 +1,82 @@
...
+		fmt.Fprintf(w, "Hello, %q", html.EscapeString(r.URL.Path))
+	})
+	if err := s.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		fmt.Println(err)
+	}
+	fmt.Println("Finished")
+	finished <- true
+}
+func TestAgentClient(t *testing.T) {
+
```
localvar, line 49:
```suggestion

func TestAgentClient(t *testing.T) {
```

pkg/common/option_test.go
haoel: Those unit tests could confuse people, I am going to rewrite them.

localvar: I propose to remove `option.go`, it is not used in our code anymore.
Also propose to remove `go.go` (tricky and never used) and `time.go` (tricky, only referenced in an unused logger function, can be replaced by standard library).

benja-wu: Agree with removing `option.go` in `pkg/common`, since we had changed to use `pkg/option` already. 
The same to `go.go`.

haoel: I suggest another dedicated PR to clean up the source code.  this one is just only for the unit tests.

localvar (approved):

PR #138, opened 2021-07-21
add user-guide for WasmHost

### LINKED ISSUE #119: AssemblyScript SDK
As the [WebAssembly filter](https://github.com/megaease/easegress/pull/116) is almost done, users will be able to leverage WebAssembly to extend the ability of Easegress soon.

Easegress has provided a set of host functions for the interoperation between Easegress and user-developed WebAssembly code, and because it is really painful to write WebAssembly code directly, Easegress also needs to provide SDKs for different high-level languages to wrap the obscure things.

The 1st language we want to support is JavaScript, the most popular one of today. But we cannot do that because there's still [not a good compiler to compile JavaScript to WebAssembly](https://github.com/WebAssembly/design/issues/219).

The alternative choice for JavaScript is [AssemblyScript](https://www.assemblyscript.org/), which is a variant of [TypeScript](https://www.typescriptlang.org/), has similar syntax with JavaScript and was designed for WebAssembly.

localvar:
add user-guide for WasmHost (close #119)

### DIFF
--- a/doc/filters.md
+++ b/doc/filters.md
@@ -603,7 +603,7 @@ oauth2:
 
 ## WasmHost
 
-The WasmHost filter implements a host environment for user-developed [WebAssembly](https://webassembly.org/) code. Below is an example configuration that loads wasm code from a file.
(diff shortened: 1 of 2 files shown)


xxx7xxxx (approved):

benja-wu (approved):
LGTM

PR #140, opened 2021-07-22
fix compile error

localvar:
APIGroup and APIEntry were renamed to Group and Entry, but some of their usages haven't been updated simultaneously.


xxx7xxxx (approved):

benja-wu (approved):
LGTM

PR #143, opened 2021-07-24
add more wasm host functions and support `wasm_init`

localvar:


megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 143 Deploy Test failed

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 143 Deploy Test Success

benja-wu (approved):
LGTM

pkg/filter/wasmhost/hostfunc.go
```diff
@@ -380,6 +382,14 @@ func (vm *WasmVM) hostLog(level int32, addr int32) {
 	}
 }
 
+func (vm *WasmVM) hostNow() int64 {
+	return time.Now().UnixNano() / 1e6
```
xxx7xxxx, line 386: Why not just export nanosecond.

localvar: there are two reasons:
1. The host functions need to support many languages, so we have to choose the lowest precision. 
2. Nanosecond requires 64 bits integers, but in some languages,  all numeric values are IEEE754 doubles and the max safe integer value is `2^53-1`, which is not larger enough to represent time in nanosecond (about 104 days).

xxx7xxxx: The put the time unit into the function name of document.

localvar: updated

xxx7xxxx (approved):

PR #150, opened 2021-07-27
Jwt pkg version update

benja-wu:
As for Github dependantbot's alert

* jwt-go allows attackers to bypass intended access restrictions in situations with []string{} for m["aud"] (which is allowed by the specification). Because the type assertion fails, "" is the value of aud. This is a security problem if the JWT token is presented to a service that lacks its own audience check. There is no patch available and users of jwt-go are advised to migrate to golang-jwt at version 3.2.1

1. Update jwt version to 3.2.1
2. Replace pkg with community management repo.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	github.com/Shopify/sarama v1.29.1
 	github.com/alecthomas/jsonschema v0.0.0-20210526225647-edb03dcab7bc
 	github.com/bytecodealliance/wasmtime-go v0.28.0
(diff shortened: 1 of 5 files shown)


localvar (approved):

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 150 Deploy Test failed

xxx7xxxx (approved):

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 150 Deploy Test Success

PR #202, opened 2021-08-20
create objects from spec files at startup (close #147)

### LINKED ISSUE #147: create objects from spec files at startup
# Background

Easegress now could be configured as a K8s ingress controller, however, users must manually send a command to Easegress to apply the spec, this is not a user-friendly fashion and greatly increases the difficulty of the K8s ingress controller deployment process.

# Proposal

Add new command-line arguments to direct Easegress to load more spec files at startup, and create objects according to these specs.

To make specs of an Easegress cluster consistent, they need to be saved into Etcd as if they are applied by command.

Because after an Easegress instance started, object specs could be updated dynamically, which result in Etcd contains newer copies of the specs than the files, Easegress should use the copy from Etcd if the spec of an object exists both in specs files and Etcd.

localvar:


megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 202 Deploy Test Success

PR #234, opened 2021-09-03
add tcp/udp proxy feature

jxd134:
Add tcp/udp proxy feature
Show it for everyone to comment or correct.


megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

pkg/object/tcpserver/tcpserver.go
```diff
@@ -0,0 +1,93 @@
...
+	panic("implement me")
+}
+
+// Init initializes HTTPServer.
+func (T *TCPServer) Init(superSpec *supervisor.Spec, muxMapper protocol.MuxMapper) {
+	panic("implement me")
+}
+
+// Inherit inherits previous generation of HTTPServer.
+func (T *TCPServer) Inherit(superSpec *supervisor.Spec, previousGeneration supervisor.Object, muxMapper protocol.MuxMapper) {
```
jxd134, line 91: @benja-wu `Init` and `Inherit` is designed for http handler.
If i use traffic controller to manage tcp proxy server, may be i need change `Init`/`Inherit` or add new function.

jxd134 (commented):
I tend to add new function in interface.

pkg/object/tcpserver/tcpserver.go
benja-wu: Any reasonable updates or modifications are welcome! 
It's cool to bring new and better implementation into Easegress. :-)

haoel:
@jxd134 There are some warnings in your code which reported by glint, could you please help to fix them?

jxd134:
@haoel I will fix it in future commit.

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

jxd134:
The main work is almost complete.
Subsequent changes will have an impact on the overall design, so further discussion is needed.

For example:
```golang
// TrafficObject is the object of Traffic
	TrafficObject interface {
		Object

		// Init initializes the Object.
		Init(superSpec *Spec, muxMapper protocol.MuxMapper)

		// Inherit also initializes the Object.
		// But it needs to handle the lifecycle of the previous generation.
		// So its own responsibility for the object to inherit and clean the previous generation stuff.
		// The supervisor won't call Close for the previous generation.
		Inherit(superSpec *Spec, previousGeneration Object, muxMapper protocol.MuxMapper)
}
```

`TrafficObject` only support http `MuxMapper`.
The `Layer4MuxMapper` is different with `MuxMapper`。
I have no good ideas to solve this problem at this stage.

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

pkg/object/udpproxy/runtime.go
```diff
@@ -0,0 +1,255 @@
...
+		return
+	}
+
+	var cp *connPool
+	if r.spec.HasResponse {
+		cp = newConnPool()
+	}
+
+	go func() {
+		defer cp.close()
```
localvar, line 101: `cp` could be nil.

pkg/object/udpproxy/session.go
```diff
@@ -0,0 +1,204 @@
...
+		buf.Release()
+		return nil
+	case <-t.C:
+		buf.Release()
+		return fmt.Errorf("write data to channel timeout")
+	}
+}
+
+// ListenResponse session listen upstream connection response and send to downstream
+func (s *session) ListenResponse(sendTo *net.UDPConn) {
```
localvar, line 142: if two clients send requests to one server, how to know whom the response is for?

pkg/object/udpproxy/runtime.go
```diff
@@ -0,0 +1,255 @@
...
+
+	r.mu.Lock()
+	for k, s := range r.sessions {
+		delete(r.sessions, k)
+		s.Close()
+	}
+	r.sessions = nil
+	r.mu.Unlock()
+
+	r.pool.Close()
```
localvar, line 79: we should do these cleanup operations until we are sure all other goroutines have stopped, using a lock here won't make the code goroutine safe in most cases.

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,441 @@
...
+	mu               sync.Mutex
+	startOnce        sync.Once
+	connStopChan     chan struct{} // use for connection close
+	listenerStopChan chan struct{} // use for listener close
+
+	onRead  func(buffer *iobufferpool.StreamBuffer) // execute read filters
+	onClose func(event ConnectionEvent)
+}
+
+// NewDownstreamConn wrap connection create from client
```
localvar, line 61: same as the comment in UDP, please make the meaning of downstream/upstream clear.

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,441 @@
...
+// @param remoteAddr client addr for udp proxy use
```
localvar, line 62:
```suggestion
// @param remoteAddr client addr for tcp proxy use
```

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,441 @@
...
+
+// State get connection running state
+func (c *Connection) State() ConnState {
+	if atomic.LoadUint32(&c.closed) == 1 {
+		return ConnClosed
+	}
+	if atomic.LoadUint32(&c.connected) == 1 {
+		return ConnActive
+	}
+	return ConnInit
```
localvar, line 125: propose to use one variable for `c.closed` and `c.connected`, using atomic functions like this is not correct in most cases.

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,441 @@
...
+	return c.rawConn.(io.Reader).Read(c.readBuffer)
+}
+
+func (c *Connection) doWrite() (int64, error) {
+	bytesSent, err := c.doWriteIO()
+	if err != nil && atomic.LoadUint32(&c.closed) == 1 {
+		return 0, nil
+	}
+
+	if bytesBufSize := c.writeBufLen(); bytesBufSize != c.lastWriteSizeWrite {
```
localvar, line 343: what if `c.closed` becomes 1 here?
please be very very careful with `atomic` functions, it is much easier to make mistakes than looks.

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,441 @@
...
+	Connection
+	connectTimeout time.Duration
+	connectOnce    sync.Once
+}
+
+// NewUpstreamConn construct tcp upstream connection
+func NewUpstreamConn(connectTimeout uint32, upstreamAddr net.Addr, listenerStopChan chan struct{}) *UpstreamConnection {
+	conn := &UpstreamConnection{
+		Connection: Connection{
+			connected:  1,
```
localvar, line 387: connected?

pkg/object/tcpproxy/listener.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func (l *listener) setMaxConnection(maxConn uint32) {
+	l.tcpListener.SetMaxConnection(maxConn)
+}
+
+func (l *listener) close() (err error) {
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
```
localvar, line 111: seems the lock is only used here.

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 234 Deploy Test failed

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+						c.localAddr.String(), c.remoteAddr.String())
+					tcpBufferPool.Put(c.readBuffer)
+					return
+				}
+
+				if te, ok := err.(net.Error); ok && te.Timeout() {
+					if n == 0 {
+						continue // continue read data, ignore timeout error
+					}
+				}
```
localvar, line 176: why n==0 && timeout is not an error? and why n >0 && timeout is an error?

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+			return
+		case <-c.listenerStopChan:
+			return
+		default:
+			n, err := c.doReadIO()
+			if n > 0 {
+				c.onRead(iobufferpool.NewStreamBuffer(c.readBuffer[:n]))
+			}
+
+			if err != nil {
```
localvar, line 164: early continue to reduce nesting.

```suggestion
			if err == nil {
			    continue
			}
```

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+
+		u.rawConn, err = net.DialTimeout("tcp", addr.String(), timeout)
+		if err != nil {
+			if err == io.EOF {
+				err = errors.New("server has been closed")
+			} else if te, ok := err.(net.Error); ok && te.Timeout() {
+				err = errors.New("connect to server timeout")
+			} else {
+				err = errors.New("connect to server failed")
+			}
```
localvar, line 385: do we need to distinguish these errors and overwrite the low-level errors?

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+	c.ioBuffers = c.ioBuffers[:0]
+	c.writeBuffers = c.writeBuffers[:0]
+	return
+}
+
+// ServerConnection wrap tcp connection to backend server
+type ServerConnection struct {
+	Connection
+	connectTimeout time.Duration
+	connectOnce    sync.Once
```
localvar, line 343: why we need to use sync.Once here? do we expect `Connect` to be called multiple times and concurrently?

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+					}
+					c.appendBuffer(buf)
+				default:
+					break OUTER
+				}
+			}
+			_, err = c.doWrite()
+		}
+
+		if err != nil {
```
localvar, line 224: early continue to reduce nesting.

```suggestion
		if err == nil {
		    continue
		 }
```

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+	lastBytesSizeRead  int
+	lastWriteSizeWrite int
+
+	readBuffer      []byte
+	writeBuffers    net.Buffers
+	ioBuffers       []*iobufferpool.StreamBuffer
+	writeBufferChan chan *iobufferpool.StreamBuffer
+
+	mu               sync.Mutex
+	startOnce        sync.Once
```
localvar, line 59: why we need to use sync.Once here? do we expect Start to be called multiple times and concurrently?

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+				logger.Errorf("tcp connection goroutine panic: %v\n%s\n", r, string(debug.Stack()))
+				if recoverHandler != nil {
+					// it is not needed to wrap recoverHandler with go func in the current scenario
+					recoverHandler(r)
+				}
+			}
+		}()
+		handler()
+	}()
+}
```
localvar, line 121: It is good to use some patterns, but it is also important to keep code short, plain, and easy to read, we need to balance.

```suggestion
	fnRecover := func() {
		if r := recover(); r != nil {
			logger.Errorf("tcp connection goroutine panic: %v\n%s\n", r, string(debug.Stack()))
			c.Close(NoFlush, LocalClose)
		}
	}
		
	c.startOnce.Do(func() {
		go func() {
			defer fnRecover()
			c.startReadLoop()
		}()

		go func() {
			defer fnRecover()
			c.startWriteLoop()
		}()
	})
}
```

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,395 @@
...
+				case buf, ok := <-c.writeBufferChan:
+					if !ok {
+						return
+					}
+					c.appendBuffer(buf)
+				default:
+					break OUTER
+				}
+			}
+			_, err = c.doWrite()
```
localvar, line 221: please try to refactor this piece of code (or maybe the whole function).

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

pkg/object/tcpproxy/connection.go
jxd134: No need, I have corrected it in the new commit.

pkg/object/tcpproxy/connection.go
jxd134: The code is inspired by `getty`.
But there is a problem with my implementation, i will fix it.

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 234 Deploy Test Success

pkg/object/tcpproxy/connection.go
```diff
@@ -0,0 +1,482 @@
...
+			logger.Errorf("tcp connection close panic, err: %+v\n%s", r, string(debug.Stack()))
+		}
+	}()
+
+	if ccType == FlushWrite {
+		_ = c.Write(iobufferpool.NewIoBufferEOF())
+		return nil
+	}
+
+	if !atomic.CompareAndSwapUint32(&c.closed, 0, 1) {
```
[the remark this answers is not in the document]
jxd134: I removed most scenarios where atomic variables are used.
Now atomic variables are mainly used to protect `Connection.Close` function from being repeatedly called.

pkg/object/udpproxy/session.go
```diff
@@ -0,0 +1,224 @@
...
+		clientIdleTimeout: clientIdleTimeout,
+		writeBuf:          make(chan *iobufferpool.Packet, 512),
+
+		stopped:      false,
+		stopChan:     make(chan struct{}),
+		listenerStop: listenerStop,
+		onClose:      onClose,
+	}
+
+	go func() {
```
samutamm, line 67: nit: This `newSession` constructor is a bit long. I would consider refactoring lines 67-120 to a private function, for example `startSession` etc and then here call `go s.startSession()`.

samutamm (approved):

pkg/object/udpproxy/session.go
jxd134: Good idea.
I will refractor it.

pkg/util/ipfilter/layer4ipfilters.go
```diff
@@ -0,0 +1,84 @@
...
+	m.rules.Store(&ipFiltersRules{
+		spec:     spec,
+		ipFilter: New(spec),
+	})
+	return m
+}
+
+// AllowIP check whether the IP is allowed to pass
+func (i *Layer4IpFilters) AllowIP(ip string) bool {
+	rules := i.rules.Load().(*ipFiltersRules)
```
samutamm, line 53: UDP Server seems to expect that `i.rules` is not empty. When I create `UDPServer` without **ipFilters** specified:
```yaml
kind: UDPServer
name: udp-server-demo
port: 10070
maxConns: 1024
hasResponse: true
clientIdleTimeout: 30000
serverIdleTimeout: 30000
pool:
  servers:
    - url: '127.0.0.1:9095'
  loadBalance:
    policy: roundRobin
```
and then open UDP connection, I got this error:

```
panic: interface conversion: interface {} is nil, not *ipfilter.ipFiltersRules

goroutine 5880 [running]:
github.com/megaease/easegress/pkg/util/ipfilter.(*Layer4IpFilters).AllowIP(0xc000b272cc, {0xc00080f7d5, 0xc0014b0000})
	github.com/megaease/easegress/pkg/util/ipfilter/layer4ipfilters.go:53 +0xa5
github.com/megaease/easegress/pkg/object/udpproxy.(*runtime).startServer.func1()
	github.com/megaease/easegress/pkg/object/udpproxy/runtime.go:118 +0x12c
created by github.com/megaease/easegress/pkg/object/udpproxy.(*runtime).startServer
	github.com/megaease/easegress/pkg/object/udpproxy/runtime.go:92 +0x245
```

This does not happen with TCPServer, only in the case of UDPServer.

jxd134: It's my mistake.
I will fix it.

samutamm: I tried the fix in https://github.com/megaease/easegress/pull/234/commits/af9a795ef79cae3608f635743fca527c97153e4a and it seems to help with `UDPProxy` not having ipFilters. But now the error comes when creating `TCPProxy` without **ipFilters**:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1b3bdea]

goroutine 327 [running]:
github.com/megaease/easegress/pkg/util/ipfilter.(*Layer4IpFilters).ReloadRules(0x0, 0x0)
	github.com/megaease/easegress/pkg/util/ipfilter/layer4ipfilters.go:63 +0x12a
github.com/megaease/easegress/pkg/object/tcpproxy.(*runtime).reload(0xc002a909c0, 0xc0007b0680)
	github.com/megaease/easegress/pkg/object/tcpproxy/runtime.go:128 +0x68
github.com/megaease/easegress/pkg/object/tcpproxy.(*runtime).handleEventReload(...)
	github.com/megaease/easegress/pkg/object/tcpproxy/runtime.go:261
github.com/megaease/easegress/pkg/object/tcpproxy.(*runtime).fsm(0xc002a909c0)
	github.com/megaease/easegress/pkg/object/tcpproxy/runtime.go:112 +0x9c
created by github.com/megaease/easegress/pkg/object/tcpproxy.newRuntime
	github.com/megaease/easegress/pkg/object/tcpproxy/runtime.go:91 +0x187
```
I think it's better to always create the `Layer4IpFilters` AND `ipFiltersRules` structs, like this

```go
func NewLayer4IPFilters(spec *Spec) *Layer4IpFilters {
       m := &Layer4IpFilters{}
       rules := &ipFiltersRules{}
       if spec != nil {
               rules = &ipFiltersRules{
                       spec:     spec,
                       ipFilter: New(spec),
               }
        }
        m.rules.Store(rules)
        return m
 }
```
Now it should avoid the both nil pointer exceptions and work as expected... you can verify if that's the case.

jxd134: So sad for my modification.
I will take your suggestions and complete the repair as soon as possible.

jxd134:
I am now confused。
The pr put the socket read/write operations in two goroutines.  When I actively close the socket, the read/write operation will report an exception. The key is that the thrown exception is difficult to distinguish whether it is caused by closing the socket or caused by a network exception.
The general practice is to set a timeout for socket read/write, and when a timeout exception is thrown, determine whether the connection is closed at this time. But which goroutine will close the socket after that, I don’t have a good idea now.

samutamm:
> I am now confused. The pr put the socket read/write operations in two goroutines. When I actively close the socket, the read/write operation will report an exception. The key is that the thrown exception is difficult to distinguish whether it is caused by closing the socket or caused by a network exception. The general practice is to set a timeout for socket read/write, and when a timeout exception is thrown, determine whether the connection is closed at this time. But which goroutine will close the socket after that, I don’t have a good idea now.

Are you experiencing some specific reproducible error or is this more like a general question about the design? For directional channels the sender should close the channel but here the connections are bidirectional, right?

jxd134:
> 



> > I am now confused. The pr put the socket read/write operations in two goroutines. When I actively close the socket, the read/write operation will report an exception. The key is that the thrown exception is difficult to distinguish whether it is caused by closing the socket or caused by a network exception. The general practice is to set a timeout for socket read/write, and when a timeout exception is thrown, determine whether the connection is closed at this time. But which goroutine will close the socket after that, I don’t have a good idea now.
> 
> Are you experiencing some specific reproducible error or is this more like a general question about the design? For directional channels the sender should close the channel but here the connections are bidirectional, right?

yes, that's it

PR #248, opened 2021-09-10
make `docker run` with default config with not param 

benja-wu:
* Remove `--` in bash command
* Modify ENTRYPOINT.sh, if not parameter provided, run `/opt/easegress/bin/easegress-server` directly, else run the fully command provided by user.

### DIFF
--- a/build/package/Dockerfile
+++ b/build/package/Dockerfile
@@ -20,4 +20,4 @@ RUN apk --no-cache add tini tzdata && \
 
 ENV PATH /opt/easegress/bin:$PATH
 
-ENTRYPOINT ["/sbin/tini", "--", "/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "/entrypoint.sh"]
--- a/build/package/Dockerfile.goreleaser
+++ b/build/package/Dockerfile.goreleaser
@@ -10,4 +10,4 @@ RUN apk --no-cache add tini tzdata && \
 
 ENV PATH /opt/easegress/bin:$PATH
 
-ENTRYPOINT ["/sbin/tini", "--", "/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini",  "/entrypoint.sh"]
--- a/build/package/entrypoint.sh
+++ b/build/package/entrypoint.sh
@@ -4,6 +4,6 @@ set -e
 if [ $# != 0  ] ; then
   exec "$@"
 else
-  exec /opt/easegress/bin/easegress-server "$@"
+  exec /opt/easegress/bin/easegress-server
 fi
 
--- a/build/package/entrypoint.sh
+++ b/build/package/entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
+set -e
 
-if [ "$(echo $1 | head -c 1)" != "-" ] ; then
+if [ $# != 0  ] ; then
   exec "$@"
 else
   exec /opt/easegress/bin/easegress-server "$@"


megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 248 Deploy Test Success

localvar (approved):

build/package/entrypoint.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/sh
+set -e
 
-if [ "$(echo $1 | head -c 1)" != "-" ] ; then
+if [ $# != 0  ] ; then
   exec "$@"
```
haoel, line 5: this would allow running any command, it's fine but not good.

this image does not like other base images - such as alpine, ubuntu. this image is dedicated to easegress.

So, it's better for people can easy add the easegress parameters in command line instead ask user run something like this:

```
docker run easegress /opt/easegress/bin/easegress-serve foo bar
```

the better way should be:

```
docker run easegress easegress  foo bar
```

Please check the Docker official best practice - [EntryPoint](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint) , and take a look the postgress example.

haoel (changes requested):

build/package/Dockerfile
```diff
@@ -20,4 +20,4 @@ RUN apk --no-cache add tini tzdata && \
 
 ENV PATH /opt/easegress/bin:$PATH
 
-ENTRYPOINT ["/sbin/tini", "--", "/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "/entrypoint.sh"]
```
haoel, line 23: How about adding a default CMD ["easegress"] after ENTRYPOINT?

benja-wu: Sure, let me update it.

build/package/entrypoint.sh
benja-wu: * The original docker image file can let user run default configuration with 
`docker run easegress --`
  but it supports adding param in this format
`dock run easegress --foo --bar`
* My first consideration is to support running in 
`docker run easegress` with default configuration. And yes, `docker run easegress /opt/easegress/bin/easegress-serve foo bar` is a bit weird, let me align with other popular product's image

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 248 Deploy Test Success

build/package/entrypoint.sh
benja-wu: @haoel 
Update it according to the best docker best practice guide.
Now we support 
``` bash
docker run -it  megaease/easegress:latest  easegress-sever --debug  // with whatever command u want

docker run -it  megaease/easegress:latest  --debug   // only with easegress parameters

docker run -it  megaease/easegress:latest    // with all default parameters
```

build/package/Dockerfile
```diff
@@ -20,4 +20,6 @@ RUN apk --no-cache add tini tzdata && \
 
 ENV PATH /opt/easegress/bin:$PATH
 
-ENTRYPOINT ["/sbin/tini", "--", "/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "--" ,"/entrypoint.sh"]
```
xxx7xxxx, line 23: What's the purpose to adjust the place of `,`?

build/package/Dockerfile.goreleaser
```diff
@@ -10,4 +10,6 @@ RUN apk --no-cache add tini tzdata && \
 
 ENV PATH /opt/easegress/bin:$PATH
 
-ENTRYPOINT ["/sbin/tini", "--", "/entrypoint.sh"]
+ENTRYPOINT ["/sbin/tini", "--" ,"/entrypoint.sh"]
```
xxx7xxxx, line 13: Ditto.

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 248 Deploy Test Success

PR #272, opened 2021-09-24
fix multi-level custom object json marshal error

localvar:


megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 272 Deploy Test failed

megaeasex (changes requested):
[TASK:easegress-pr-test FAILED]megaease/easegress Pull Request 272 Deploy Test failed

pkg/object/meshcontroller/spec/spec.go
```diff
@@ -312,6 +313,39 @@ func (cr CustomResource) Kind() string {
 	return ""
 }
 
+// MarshalJSON implements json.Marshaler
```
xxx7xxxx, line 316: You can do a research of https://github.com/json-iterator/go , which might already have done it.

localvar: yes, `json-iterator` can handle this case, thanks for the advice.

btw, I think the document of `json-iterator` is inaccurate as it says "100% compatible with `encoding/json`", at least in this case.

xxx7xxxx: Compatible means it includes standard lib here. it depends on how you understand it. Golang1.17 is compatible with Golang 1.0 with bringing a lot of new features.

localvar: After thinking the issue twice, I think it is better to keep this function: 

This function is always a correct solution, no matter we are using `encoding/json` or `json-iterator`. 
If we remove it, then we are forcing future code to use `json-iterator`, as this requirement is an implicit one, we may meet this bug again if `encoding/json` is used in new code.

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 272 Deploy Test Success

pkg/object/meshcontroller/spec/spec.go
xxx7xxxx: Sure, I have no opposite opinion

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 272 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 272 Deploy Test Success

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 272 Deploy Test Success

pkg/object/meshcontroller/spec/spec.go
localvar: Because I found other issues with this implementation, I've moved the type conversion stuff to `UnmarshalYAML`.

megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 272 Deploy Test Success

PR #278, opened 2021-09-26
update cluster example, no std-log-level in option

suchen-sci:
No `std-log-level` in Easegress option, the default log level is INFO. To use log level of DEBUG, set `debug: true` in Easegress yaml. To replace `std-log-level: INFO` with `debug: false` will hint the people who want to use debug mode.

### DIFF
--- a/doc/benchmark.md
+++ b/doc/benchmark.md
@@ -151,7 +151,7 @@ cpu-profile-file:
 memory-profile-file:
 log-dir: ./log
 member-dir: ./member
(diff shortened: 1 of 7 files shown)


megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 278 Deploy Test Success

benja-wu (approved):

localvar (approved):

PR #302, opened 2021-10-14
fix MQTT client close bugs

suchen-sci:
This pr mainly fix two bugs. 
1. When MQTT client and broker close at the same time, the client session may be closed twice, then double close channel will cause panic. (call `SessionManager.delLocal` method).
2. When close client, the broker does not rm client from its map. Remove client from map is easy, but it may cause race condition. For example, when old client close and remove from broker, at same time, new client with same clientID come and store it to broker. Then broker may remove new client. 

I hope this pr fix these bugs.

### DIFF
--- a/pkg/object/mqttproxy/broker.go
+++ b/pkg/object/mqttproxy/broker.go
@@ -287,6 +287,16 @@ func (b *Broker) getClient(clientID string) *Client {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


haoel:
>  I hope this pr fix these bugs.

You need to make sure this patch fixes the bugs. not just hope...

PR #327, opened 2021-11-01
refactor template for performance

localvar:
* avoid some memory allocation
* early return instead of always process all input


megaeasex (approved):
[TASK:easegress-pr-test SUCCESS]megaease/easegress Pull Request 327 Deploy Test Success

xxx7xxxx:
I'm fine with this change, but along with the new pipeline jumping in this PR is useless, so for stability, we could suspend it.

PR #374, opened 2021-11-15
add pipeline for mqtt disconnect packet

suchen-sci:


pkg/object/mqttproxy/client.go
```diff
@@ -77,6 +80,18 @@ func (c *Client) ClientID() string {
...
+func (c *Client) Load(key interface{}) (value interface{}, ok bool) {
+	return c.kvMap.Load(key)
+}
+
+func (c *Client) Store(key interface{}, value interface{}) {
+	c.kvMap.Store(key, value)
+}
+
+func (c *Client) Delete(key interface{}) {
+	c.kvMap.Delete(key)
```
zhao-kun, line 92: please fix all complaints of the go lint

pkg/context/mqttcontext.go
```diff
@@ -88,8 +93,11 @@ const (
 	// MQTTPublish is mqtt packet type of publish
 	MQTTPublish MQTTPacketType = 2
 
+	// MQTTDisconnect is mqtt packet type of disconnect
+	MQTTDisconnect = 3
+
 	// MQTTOther is all other mqtt packet type
-	MQTTOther MQTTPacketType = 3
+	MQTTOther MQTTPacketType = 4
```
zhao-kun, line 100: Let MQTTOther's value equal a large value, for example, `999` or `9999`, to avoid value conflict in the future.

suchen-sci: Good idea!

zhao-kun (approved):

PR #375, opened 2021-11-16
mqtt context add backend method to publish message to backend

suchen-sci:


pkg/object/mqttproxy/broker.go
```diff
@@ -200,7 +202,7 @@ func (b *Broker) handleConn(conn net.Conn) {
 			logger.Errorf("get pipeline %v failed, %v", b.pipeline, err)
 			authFail = true
 		} else {
-			ctx := context.NewMQTTContext(stdcontext.Background(), client, connect)
+			ctx := context.NewMQTTContext(stdcontext.Background(), b, client, connect)
```
zhao-kun, line 205: Broker or backend, which should implement the Backend interface? it's confusing people.

suchen-sci: I implement it in `Broker` because we may support multiple backends and broker can send them all at once. `Backend` here is just general backend. But change to `broker.backend` is also fine for me....

suchen-sci: The other reason to use `Broker` is in MQTTContext, `Backend` is general, in future, if we add some other features, `Broker` may be flexible than `broker.backend`. Maybe change `MQTTContext.Backend` to `MQTTContext.Broker`?

zhao-kun: `Broker`.backend is really the backend, it's literally backend for MQTT proxy. IMHO,  I would say the sense that a broker isn't a backend is the instinct of people.

suchen-sci: Make sense, I will change it now!

zhao-kun (approved):

PR #376, opened 2021-11-16
feature: global filter

brzyangg:
Add business controller global filter

### DIFF
--- a/pkg/object/globalfilter/globalfilter.go
+++ b/pkg/object/globalfilter/globalfilter.go
@@ -198,7 +198,6 @@ func (gf *GlobalFilter) afterHandle(ctx context.HTTPContext) string {
 }
 
 // Close closes GlobalFilter itself.
-// Supervisor won't call Close for previous generation in Update.
 func (gf *GlobalFilter) Close() {
 
 }
--- a/pkg/object/globalfilter/globalfilter.go
+++ b/pkg/object/globalfilter/globalfilter.go
@@ -19,7 +19,9 @@ package globalfilter
 
 import (
 	"fmt"
-	"sync"
+	"sync/atomic"
+
+	"github.com/megaease/easegress/pkg/protocol"
 
 	"github.com/megaease/easegress/pkg/context"
 	"github.com/megaease/easegress/pkg/object/httppipeline"
@@ -33,9 +35,6 @@ const (
 
 	// Kind is the kind of GlobalFilter.
 	Kind = "GlobalFilter"
-
-	beforePipelineKey = "before"
-	afterPipelineKey  = "after"
 )
 
 type (
@@ -46,9 +45,8 @@ type (
 		superSpec *supervisor.Spec
 		spec      *Spec
 
-		// pipelines map[string]*httppipeline.HTTPPipeline
-		// only contains two key: before and after
-		pipelines *sync.Map
+		beforePipeline atomic.Value
+		afterPipeline  atomic.Value
 	}
 
 	// Spec describes the GlobalFilter.
@@ -73,29 +71,45 @@ func init() {
 func (gf *GlobalFilter) CreateAndUpdateBeforePipelineForSpec(spec *Spec, previousGeneration *httppipeline.HTTPPipeline) error {
 	beforePipeline := &pipelineSpec{
 		Kind: httppipeline.Kind,
-		Name: beforePipelineKey,
+		Name: "before",
 		Spec: spec.BeforePipeline,
 	}
-	return gf.CreateAndUpdatePipeline(beforePipeline, previousGeneration)
+	pipeline, err := gf.CreateAndUpdatePipeline(beforePipeline, previousGeneration)
+	if err != nil {
+		return err
+	}
+	if pipeline == nil {
+		return fmt.Errorf("before pipeline is nil, spec: %v", beforePipeline)
+	}
+	gf.beforePipeline.Store(pipeline)
+	return nil
 }
 
 // CreateAndUpdateAfterPipelineForSpec ...
 func (gf *GlobalFilter) CreateAndUpdateAfterPipelineForSpec(spec *Spec, previousGeneration *httppipeline.HTTPPipeline) error {
 	afterPipeline := &pipelineSpec{
 		Kind: httppipeline.Kind,
-		Name: beforePipelineKey,
+		Name: "after",
 		Spec: spec.AfterPipeline,
 	}
-	return gf.CreateAndUpdatePipeline(afterPipeline, previousGeneration)
+	pipeline, err := gf.CreateAndUpdatePipeline(afterPipeline, previousGeneration)
+	if err != nil {
+		return err
+	}
+	if pipeline == nil {
+		return fmt.Errorf("after pipeline is nil, spec: %v", afterPipeline)
+	}
+	gf.afterPipeline.Store(pipeline)
+	return nil
 }
 
 // CreateAndUpdatePipeline create and update globalFilter`s pipelines from pipeline spec
-func (gf *GlobalFilter) CreateAndUpdatePipeline(spec *pipelineSpec, previousGeneration *httppipeline.HTTPPipeline) error {
+func (gf *GlobalFilter) CreateAndUpdatePipeline(spec *pipelineSpec, previousGeneration *httppipeline.HTTPPipeline) (*httppipeline.HTTPPipeline, error) {
 	// init config
 	config := yamltool.Marshal(spec)
 	specS, err := supervisor.NewSpec(string(config))
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	// init or update pipeline
@@ -105,8 +119,7 @@ func (gf *GlobalFilter) CreateAndUpdatePipeline(spec *pipelineSpec, previousGene
 	} else {
 		pipeline.Init(specS, nil)
 	}
-	gf.pipelines.Store(spec.Name, pipeline)
-	return nil
+	return pipeline, nil
 }
 
 // Category returns the object category of itself.
@@ -135,7 +148,6 @@ func (gf *GlobalFilter) Status() *supervisor.Status {
 // Init initializes GlobalFilter.
 func (gf *GlobalFilter) Init(superSpec *supervisor.Spec) {
 	gf.superSpec, gf.spec = superSpec, superSpec.ObjectSpec().(*Spec)
-	gf.pipelines = &sync.Map{}
 	gf.reload(nil)
 }
 
@@ -145,44 +157,49 @@ func (gf *GlobalFilter) Inherit(superSpec *supervisor.Spec, previousGeneration s
 	gf.reload(previousGeneration.(*GlobalFilter))
 }
 
-// BeforeHandle before handler logic for beforePipeline spec
-func (gf *GlobalFilter) BeforeHandle(ctx context.HTTPContext) {
-	handler, ok := gf.getPipeline(beforePipelineKey)
-	if !ok {
+func (gf *GlobalFilter) Handle(ctx context.HTTPContext, httpHandle protocol.HTTPHandler) {
+	result := gf.beforeHandle(ctx)
+	if result == httppipeline.LabelEND {
 		return
 	}
-	handler.Handle(ctx)
+	result = httpHandle.Handle(ctx)
+	if result == httppipeline.LabelEND {
+		return
+	}
+	gf.afterHandle(ctx)
+	return
+}
+
+// BeforeHandle before handler logic for beforePipeline spec
+func (gf *GlobalFilter) beforeHandle(ctx context.HTTPContext) string {
+	value := gf.beforePipeline.Load()
+	if value == nil {
+		return ""
+	}
+	handler, ok := value.(*httppipeline.HTTPPipeline)
+	if !ok {
+		return ""
+	}
+	return handler.Handle(ctx)
 }
 
 // AfterHandle after handler logic for afterPipeline spec
-func (gf *GlobalFilter) AfterHandle(ctx context.HTTPContext) {
-	handler, ok := gf.getPipeline(afterPipelineKey)
+func (gf *GlobalFilter) afterHandle(ctx context.HTTPContext) string {
+	value := gf.afterPipeline.Load()
+	if value == nil {
+		return ""
+	}
+	handler, ok := value.(*httppipeline.HTTPPipeline)
 	if !ok {
-		return
+		return ""
 	}
-	handler.Handle(ctx)
-}
-
-func (gf *GlobalFilter) getPipeline(key string) (*httppipeline.HTTPPipeline, bool) {
-	value, ok := gf.pipelines.Load(key)
-	if !ok || value == nil {
-		return nil, false
-	}
-	pipe, ok := value.(*httppipeline.HTTPPipeline)
-	return pipe, ok
+	return handler.Handle(ctx)
 }
 
 // Close closes itself. It is called by deleting.
 // Supervisor won't call Close for previous generation in Update.
 func (gf *GlobalFilter) Close() {
-	gf.pipelines.Range(func(key, value interface{}) bool {
-		if v, ok := value.(*httppipeline.HTTPPipeline); ok {
-			v.Close()
-		}
-		return true
-	})
-	gf.pipelines.Delete(beforePipelineKey)
-	gf.pipelines.Delete(afterPipelineKey)
+
 }
 
 // Validate validates Spec.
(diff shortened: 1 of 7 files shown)


pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,217 @@
+package globalfilter
```
suchen-sci, line 1: Please add license header in this file. The `Copyright` stuff to pass License checker.

pkg/object/httpserver/mux.go
```diff
@@ -454,7 +456,9 @@ func (m *mux) handleRequestWithCache(rules *muxRules, ctx context.HTTPContext, c
 			path = ci.path.pathRE.ReplaceAllString(path, ci.path.rewriteTarget)
 			ctx.Request().SetPath(path)
 		}
+		m.beforeGlobalHTTPFilterHandle(ctx, rules)
```
suchen-sci, line 459: It seems `beforeGlobalHTTPFilterHandle` `jumpIf: END` will only jump to the end of `beforePipeline`. Which means if `beforePipeline` is used to check authentication, then requests without authentication can still go to following pipelines. 

One way to solve it is to change `HTTPPipeline.Handle` to return if it `jumpif` to end. Another way may be check if context if canceled. I will check the related codes. 
More advices need here.

brzyangg: This is a problem. I think we should discuss how to solve it. 
First of all, I also think that if `jumpIf: END` is triggered in `beforePipeline`, `pipeline` and `afterPipeline` should not be executed.
According to the current logic, it is a better choice to check whether the context is cancelled.
More advices need here.

samutamm: I wonder if canceling context is a good idea if the context is also used to send the response after failed authentication? Or could there be a flag `beforePipelineCancelled`?

Here's a bit detailed proposal how to modify `HTTPPipeline.Handle` to pass the `jumpIf: END` information. Currently `HTTPPipeline.Handle` does not make difference between arriving to the end of pipeline from `jumpIf: END` [1](https://github.com/megaease/easegress/blob/main/pkg/object/httppipeline/httppipeline.go#L422-L428) or by just finishing the last filter [2](https://github.com/megaease/easegress/blob/main/pkg/object/httppipeline/httppipeline.go#L407-L409). Maybe the `HTTPPipeline.getNextFilterIndex` could return tuple (int, bool) where the boolean indicates if it jumped to the end. Then this info could be passed all way to the `pkg/object/httpserver/mux.go` when calling `m.beforeGlobalHTTPFilterHandle(ctx, rules)` and the following pipelines could be skipped, if jumped to the END in beforePipeline.

suchen-sci: I prefer `HTTPPipeline.Handle` to return `LabelEND`, `LabelEND` only used in `HTTPPipeline` to control filter execution flow. And in `HTTPPipeline` it don't check if context is canceled or call cancel function to cancel it. Depend on the filter implementation, in some case it may not call cancel function. So, if we use `HTTPPipeline.Handle`, it can 100% reach end.

brzyangg: I agree with you. 
I think `HTTPPipeline` should return a string `lastResult` or `labelEND`. 
And, modify the return value of `HTTPPipeline.getNextFilterIndex`.

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,234 @@
...
+	// GlobalFilter is a business controller
+	// provide handler before and after pipeline in HTTPServer
+	GlobalFilter struct {
+		super     *supervisor.Supervisor
+		superSpec *supervisor.Spec
+		spec      *Spec
+
+		// pipelines map[string]*httppipeline.HTTPPipeline
+		// only contains two key: before and after
+		pipelines *sync.Map
```
localvar, line 51: As there are only two possible keys, propose to use two `sync.Value` instead of `sync.Map`, this is also to align with the `Spec` definition below, where `BeforePipeline` and `AfterPipeline` are used instead of a map.

brzyangg: ok, i get it.

pkg/object/globalfilter/globalfilter.go
```diff
@@ -145,44 +157,49 @@ func (gf *GlobalFilter) Inherit(superSpec *supervisor.Spec, previousGeneration s
 	gf.reload(previousGeneration.(*GlobalFilter))
 }
 
-// BeforeHandle before handler logic for beforePipeline spec
-func (gf *GlobalFilter) BeforeHandle(ctx context.HTTPContext) {
-	handler, ok := gf.getPipeline(beforePipelineKey)
-	if !ok {
+func (gf *GlobalFilter) Handle(ctx context.HTTPContext, httpHandle protocol.HTTPHandler) {
```
suchen-sci, line 160: Please fix this Github action warning... Except this one, I think it looks great to me!

brzyangg: Thanks, I dealt with this warning

suchen-sci (approved):

localvar (approved):

zhao-kun (approved):

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		Name              string `yaml:"name" jsonschema:"omitempty"`
+		httppipeline.Spec `yaml:",inline"`
+	}
+)
+
+func init() {
+	supervisor.Register(&GlobalFilter{})
+}
+
+// CreateAndUpdateBeforePipelineForSpec ...
```
xxx7xxxx, line 70: Complete the comment.

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		return err
+	}
+	if pipeline == nil {
+		return fmt.Errorf("before pipeline is nil, spec: %v", beforePipeline)
+	}
+	gf.beforePipeline.Store(pipeline)
+	return nil
+}
+
+// CreateAndUpdateAfterPipelineForSpec ...
```
xxx7xxxx, line 88: Ditto.

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+const (
+	// Category is the category of GlobalFilter.
+	Category = supervisor.CategoryBusinessController
+
+	// Kind is the kind of GlobalFilter.
+	Kind = "GlobalFilter"
+)
+
+type (
+	// GlobalFilter is a business controller
```
xxx7xxxx, line 41:
```suggestion
	// GlobalFilter is a business controller.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+	// provide handler before and after pipeline in HTTPServer
```
xxx7xxxx, line 42:
```suggestion
	// It provides handler before and after pipeline in HTTPServer.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		afterPipeline  atomic.Value
+	}
+
+	// Spec describes the GlobalFilter.
+	Spec struct {
+		BeforePipeline httppipeline.Spec `yaml:"beforePipeline" jsonschema:"omitempty"`
+		AfterPipeline  httppipeline.Spec `yaml:"afterPipeline" jsonschema:"omitempty"`
+	}
+
+	// pipelineSpec define httppipeline spec to create a httppipeline entity
```
xxx7xxxx, line 58:
```suggestion
	// pipelineSpec defines httppipeline spec to create an httppipeline entity.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	gf.afterPipeline.Store(pipeline)
+	return nil
+}
+
+// CreateAndUpdatePipeline create and update globalFilter`s pipelines from pipeline spec
+func (gf *GlobalFilter) CreateAndUpdatePipeline(spec *pipelineSpec, previousGeneration *httppipeline.HTTPPipeline) (*httppipeline.HTTPPipeline, error) {
+	// init config
+	config := yamltool.Marshal(spec)
+	specS, err := supervisor.NewSpec(string(config))
```
xxx7xxxx, line 110: `specS` is weird, please use `specs`.

pkg/object/httpserver/mux.go
```diff
@@ -473,6 +481,17 @@ func (m *mux) appendXForwardedFor(ctx context.HTTPContext) {
...
 
+func (m *mux) getGlobalFilter(rules *muxRules) *globalfilter.GlobalFilter {
+	if rules.spec.GlobalFilter == "" {
+		return nil
+	}
+	globalFilter, ok := rules.superSpec.Super().GetBusinessController(rules.spec.GlobalFilter)
+	if globalFilter == nil || !ok {
+		return nil
+	}
+	return globalFilter.Instance().(*globalfilter.GlobalFilter)
```
xxx7xxxx, line 492: It will panic if it got the wrong type. Please do checking type assertion.

pkg/object/httppipeline/httppipeline.go
```diff
@@ -402,10 +402,11 @@ func (hp *HTTPPipeline) reload(previousGeneration *HTTPPipeline) {
 	hp.runningFilters = runningFilters
 }
 
-func (hp *HTTPPipeline) getNextFilterIndex(index int, result string) int {
+// getNextFilterIndex return index and whether end
```
samutamm, line 405:
```suggestion
// getNextFilterIndex return filter index and whether jumped to the end of the pipeline.
```

pkg/object/httppipeline/httppipeline.go
```diff
@@ -417,31 +418,32 @@ func (hp *HTTPPipeline) getNextFilterIndex(index int, result string) int {
...
 }
 
 // Handle is the handler to deal with HTTP
-func (hp *HTTPPipeline) Handle(ctx context.HTTPContext) {
+func (hp *HTTPPipeline) Handle(ctx context.HTTPContext) string {
 	ctx.SetTemplate(hp.ht)
 
 	filterIndex := -1
 	filterStat := newFilterStat()
+	isEnd := false
```
samutamm, line 446: Nit: No need to initialize this here as `isEnd` is used only on lines 472 and 473

pkg/object/httppipeline/httppipeline.go
```diff
@@ -467,7 +469,10 @@ func (hp *HTTPPipeline) Handle(ctx context.HTTPContext) {
 			filterStat = lastStat
 		}()
 
-		filterIndex = hp.getNextFilterIndex(filterIndex, lastResult)
+		filterIndex, isEnd = hp.getNextFilterIndex(filterIndex, lastResult)
```
samutamm, line 472:
```suggestion
		filterIndex, isEnd := hp.getNextFilterIndex(filterIndex, lastResult)
```
and remove line 446

pkg/object/globalfilter/globalfilter.go
brzyangg: I have completed the comment. thank you.

samutamm (approved):

xxx7xxxx:
Integration test failed, please check it.

pkg/object/httppipeline/httppipeline.go
brzyangg: This symbol `:=` cannot be used here for assignment. It will create a new `filterIndex`. It will cause the outside `filterIndex` assignment to fail

brzyangg:
> 

This problem has been fixed, mainly because filterIndex uses `:=` assignment

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		Name              string `yaml:"name" jsonschema:"omitempty"`
+		httppipeline.Spec `yaml:",inline"`
+	}
+)
+
+func init() {
+	supervisor.Register(&GlobalFilter{})
+}
+
+// CreateAndUpdateBeforePipelineForSpec If beforPipeline is nil, it will be created by the spec, otherwise updated.
```
xxx7xxxx, line 70:
```suggestion
// CreateAndUpdateBeforePipelineForSpec creates beforPipeline if the spec is nil, otherwise it updates by the spec.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		return err
+	}
+	if pipeline == nil {
+		return fmt.Errorf("before pipeline is nil, spec: %v", beforePipeline)
+	}
+	gf.beforePipeline.Store(pipeline)
+	return nil
+}
+
+// CreateAndUpdateAfterPipelineForSpec If afterPipeline is nil, it will be created by the spec, otherwise updated.
```
xxx7xxxx, line 88:
```suggestion
// CreateAndUpdateAfterPipelineForSpec creates afterPipeline if the spec is nil, otherwise it updates with the spec.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		return err
+	}
+	if pipeline == nil {
+		return fmt.Errorf("after pipeline is nil, spec: %v", afterPipeline)
+	}
+	gf.afterPipeline.Store(pipeline)
+	return nil
+}
+
+// CreateAndUpdatePipeline create and update globalFilter`s pipelines from pipeline spec.
```
xxx7xxxx, line 106:
```suggestion
// CreateAndUpdatePipeline creates and updates GlobalFilter's pipelines.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+		return ""
+	}
+	handler, ok := value.(*httppipeline.HTTPPipeline)
+	if !ok {
+		return ""
+	}
+	return handler.Handle(ctx)
+}
+
+// Close closes itself. It is called by deleting.
```
xxx7xxxx, line 200:
```suggestion
// Close closes GlobalFilter itself.
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+// Supervisor won't call Close for previous generation in Update.
```
xxx7xxxx, line 201:
```suggestion
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+func (gf *GlobalFilter) Close() {
+
+}
+
+// Validate validates Spec.
+func (s *Spec) Validate() (err error) {
+
+	err = s.BeforePipeline.Validate()
+	if err != nil {
+		return fmt.Errorf("before pipeline is invalidate err: %v", err)
```
xxx7xxxx, line 211:
```suggestion
		return fmt.Errorf("before pipeline is invalid: %v", err)
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	err = s.AfterPipeline.Validate()
+	if err != nil {
+		return fmt.Errorf("after pipeline is invalidate err: %v", err)
```
xxx7xxxx, line 215:
```suggestion
		return fmt.Errorf("after pipeline is invalid: %v", err)
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+	if len(gf.spec.BeforePipeline.Flow) != 0 {
+		if previousGeneration != nil {
+			previous := previousGeneration.beforePipeline.Load()
+			if previous != nil {
+				beforePreviousPipeline = previous.(*httppipeline.HTTPPipeline)
+			}
+		}
+		err := gf.CreateAndUpdateBeforePipelineForSpec(gf.spec, beforePreviousPipeline)
+		if err != nil {
+			panic(fmt.Sprintf("create before pipeline error %v", err))
```
xxx7xxxx, line 233:
```suggestion
			panic(fmt.Errorf("create before pipeline failed: %v", err))
```

pkg/object/globalfilter/globalfilter.go
```diff
@@ -0,0 +1,249 @@
...
+	if len(gf.spec.AfterPipeline.Flow) != 0 {
+		if previousGeneration != nil {
+			previous := previousGeneration.afterPipeline.Load()
+			if previous != nil {
+				afterPreviousPipeline = previous.(*httppipeline.HTTPPipeline)
+			}
+		}
+		err := gf.CreateAndUpdateAfterPipelineForSpec(gf.spec, afterPreviousPipeline)
+		if err != nil {
+			panic(fmt.Sprintf("create after pipeline error %v", err))
```
xxx7xxxx, line 246:
```suggestion
			panic(fmt.Errorf("create after pipeline failed: %v", err))
```

xxx7xxxx (approved):

PR #381, opened 2021-11-22
the informer waits for all goroutines to exit before it closes (#88)

### LINKED ISSUE #88: Informer send notification after closed
As discussed in https://github.com/megaease/easegress/pull/79#discussion_r660204660, it is possible for the current mesh informer implementation to send notifications after it is closed, this is unexpected and may cause bugs.

To fix this, we can add a wait operation in `informer.Close` to ensure all notifications are sent before the function returns, the change itself is simple but we need to check the existing code to ensure the `wait` does not cause a deadlock.

If the above solution is not possible, we can document the behavior and let client code handle the delayed notifications.

u2386:
close #88

### DIFF
--- a/pkg/cluster/syncer.go
+++ b/pkg/cluster/syncer.go
@@ -49,7 +49,7 @@ func (c *cluster) Syncer(pullInterval time.Duration) (*Syncer, error) {
 		return nil, err
 	}
 	return &Syncer{
-		WaitGroup: sync.WaitGroup{},
+		WaitGroup:    sync.WaitGroup{},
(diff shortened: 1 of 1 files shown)


localvar:
Thanks for the PR, as we have mentioned in #88 , we need to ensure this change does not cause a deadlock, so it will take some time for us to review and test, please be patient.

megaeasex (changes requested):
[TASK:easegress-PR-test FAILED]megaease/easegress Pull Request 381 Deploy Test failed

Issue #233, opened 2021-09-03
[proposal] Cluster operation


benja-wu:
## Background
* Easegress embeds Etcd[1] inside. It uses Raft[2] for distributing consistency. 
* Easegress has its own cluster-start-up routine, which is quite different from original Etcd's.

### Concepts
1. Writer: Same as the Etcd's Leader/Candidate/Follower. 
2. Reader: The Etcd client, for scaling reading purposes.

### How Easegress form a cluster
1. Prepare an odd number of writer instances, pick one as the chosen `Leader`.
2. Valuing the `cluster-join-urls` in configuration by starting with the chosen `Leader`'s address first.
3. Starting the chosen `Leader` firstly.
4. Starting other `writers` one by one. (the sequence doesn't matter)
5. The known cluster member will be stored into local `member.yaml`, for saving communication cost at next time restarting.

### Dive in
* In the original Etcd start-up routine, there is an election behavior, which means not so-called chosen `Leader` is needed.
* The current Easegress cluster behavior is called a `Join` instead of an `Election`. It can save the initial competing cost. But during the daily operation, the failover election can still be processed automatically.

### Pros/Cons
Pros:  Saving initial election costing by directly join the `exist` cluster instead. 
Cons: Startup sequence required, local member YAML file removed needed at some situation. 

## Questions & Discussion  
1. Should we keep this cluster design going or just roll back to use Etcd's cluster behavior?
2. Are there any more efficient cluster start-up routines can we align with, but **still has Etcd inside**? (zookeepers? consul? ...)
3. Feel free to tell us your concerns and prefers, thx

> Please focus on cluster start-up routine discussion, Etcd inside is still Easegress' first choice. 

## References
1. https://etcd.io
2. https://en.wikipedia.org/wiki/Raft_(algorithm)

xxx7xxxx:
There are more Pros/Cons in the current and native-etcd cluster operation:

Pros of the current cluster:

- It saves one step(`etcdctl member add`) to add members manually

Cons of the current cluster:

- The member file will easily confuse operators in the long-run maintaining including adding, restarting, pruning.
- The member adding behavior doesn't fit in k8s statefulsets startup, and causes some weird failure.

Pros of the native-etcd cluster:

- Most operation problems and solutions can be found in the huge etcd/k8s community.
- It is clearer for every newcomer only if they know etcd.

Cons of the native-etcd cluster:

[...]

PR #382, opened 2021-11-22
Etcd based cluster config

samutamm:
Close https://github.com/megaease/easegress/issues/233. Based on this design https://github.com/megaease/easegress/issues/233#issuecomment-971414551.

- Minor refactoring of URL parsing in [pkg/cluster/config.go](pkg/cluster/config.go)
- Add alternative way to configure cluster so that it start's all members at once ([example config](example/config/cluster/cluster-1.yaml)) VS the current way, which starts members one by one ([example legacy config](example/config/cluster/legacy-cluster-1.yaml))
- Add unittests for new configuration


pkg/option/option.go
```diff
@@ -225,75 +260,77 @@ func (opt *Options) adjust() {
 	}
 }
 
+// ParseURLs parses list of strings to url.URL objects.
+func ParseURLs(urlStrings []string) ([]url.URL, error) {
+	urls := make([]url.URL, len(urlStrings))
+	for i, urlString := range urlStrings {
+		parsedUrl, err := url.Parse(urlString)
+		if err != nil {
+			fmt.Println(urlString)
```
localvar, line 269: a log message?

pkg/option/option.go
```diff
@@ -225,75 +260,77 @@ func (opt *Options) adjust() {
...
 			}
-		}
-		for _, peerURL := range opt.ClusterInitialAdvertisePeerURLs {
-			_, err := url.Parse(peerURL)
-			if err != nil {
-				return fmt.Errorf("invalid cluster-initial-advertise-peer-urls: %s: %v", peerURL, err)
+			if len(opt.ClusterJoinURLs) > 0 {
+				err := strings.Join(strings.Fields(`
+					%s and %s are both defined.
+					Please provide only one of them. %s is the recommended way.`), " ")
```
localvar, line 321: please keep this as a string, it takes me some time to realize the purpose (maybe we can refactor this function to avoid the long code line).
and we need to find another variable name because we use `err` as an `error` in all other cases.

localvar (commented):
please also fix the issues reported by Github actions.

xxx7xxxx:
I don't go into detail about the code in the first review, I would like to specify my vision about this PR along with reasons.

1. Support older and newer roles by changing older names to newer ones in startup options. (You have already fixed it.)
2. The layout of the option is not a good design IMHO, I'd like to give a clean layout. Example without the trivial:

Primary member:

```yaml
name: primary-1
cluster-name: cluster-test
cluster-role: primary
cluster-primary:
  listen-peer-urls: [http://127.0.0.1:2380]
  listen-client-urls: [http://127.0.0.1:2379]
  advertise-client-urls: [http://127.0.0.1:2379]
  initial-advertise-peer-urls: [http://127.0.0.1:2380]
  initial-cluster-state: new
  initial-cluster:
   - primary-1: http://127.0.0.1:2380
   - primary-2: http://127.0.0.1:2378
   - primary-3: http://127.0.0.1:2376
```

Secondary member:

```yaml
name: secondary-1
cluster-name: cluster-test
cluster-role: secondary
cluster-secondary:
  primary-listen-peer-urls: [http://127.0.0.1:2380]
```

The reason I designed them like that is to make it cleaner to write/read config by splitting them up.

3. Please adapt the reader config to new rules.

4. Please don't add things to `cluster/member.go` which all will be deleted. What you changed in it will break existing cluster with member file.

samutamm:
@xxx7xxxx Thanks for the comment! I think that the example you proposed makes it simpler to add new reader/secondary members. I would develop it further like this:

Almost the same config for primary member.
```yaml
name: primary-1
cluster-name: cluster-test
cluster-role: primary # writer
cluster: # only primary member defines cluster entry 
  listen-peer-urls: [http://127.0.0.1:2380]
  listen-client-urls: [http://127.0.0.1:2379]
  advertise-client-urls: [http://127.0.0.1:2379]
  initial-advertise-peer-urls: [http://127.0.0.1:2380]
  initial-cluster-state: new
  initial-cluster:
   - primary-1: http://127.0.0.1:2380
   - primary-2: http://127.0.0.1:2378
   - primary-3: http://127.0.0.1:2376
```
and then secondary
```yaml
name: secondary-1
cluster-name: cluster-test
cluster-role: secondary # reader
# secondary member defines, where to connect to primary
primary-listen-peer-urls: [http://127.0.0.1:2380]
```

samutamm:
@xxx7xxxx I removed the extra logic from members.go, so it's easier to remove in future. The configuration is also as described above.

pkg/cluster/cluster.go
```diff
@@ -793,11 +814,32 @@ func (c *cluster) updateMembers() error {
...
 
+// UpdatePeerURLs updates peerURLs according the cluster role.
+func (c *cluster) UpdatePeerURLs(pbMembers []*pb.Member) {
+	primaryMembers := pbMembersToMembersSlice(pbMembers)
+	if c.opt.ClusterRole == "secondary" {
+		newPeerURLs := make([]string, 0)
+		for _, member := range primaryMembers {
+			newPeerURLs = append(newPeerURLs, member.PeerURL)
+		}
+		c.opt.SetClusterPrimaryListenPeerURLs(newPeerURLs)
```
xxx7xxxx, line 833: I don't think it's a good idea to store back to option, which should honestly react to the input from the command line or config file.

pkg/cluster/cluster.go
```diff
@@ -793,11 +814,32 @@ func (c *cluster) updateMembers() error {
...
+	} else {
+		newInitialCluster := make(map[string]string)
+		for _, member := range primaryMembers {
+			newInitialCluster[member.Name] = member.PeerURL
+		}
+		c.opt.SetInitialCluster(newInitialCluster)
```
xxx7xxxx, line 839: Ditto.

xxx7xxxx (commented):
1. As I commented last time, please adapt older examples of reader-004/reader-005.
2. Have you tested adding members, then shutdown all members, then restart all nodes, will it restart successfully?

pkg/cluster/cluster.go
samutamm: I do agree that the options is not the best place to store the cluster node's PeerURLs. I did this, to support adding and removing the primary members, as the set of primary members (and their URLs) change.

Another way to support adding new primary members is described [here](https://github.com/megaease/easegress/issues/233#issuecomment-977359212) in case 3. So basically instead of members listening for cluster modifications, user should explicitly call `egctl member add NAME --peer-urls=NEW-HOSTS` and same for removal. Then this `egctl` call could modify the _options_ of each member. I think this the way etcd advertise in their [tutorial](https://etcd.io/docs/v3.5/tutorials/how-to-deal-with-membership/).

pkg/option/option.go
```diff
@@ -132,6 +174,26 @@ func (opt *Options) YAML() string {
 	return opt.yamlStr
 }
 
+// UseInitialCluster returns true if the cluster.initial-cluster is defined. If it is, the ClusterJoinUrls is ignored.
+func (opt *Options) UseInitialCluster() bool {
+	return len(opt.Cluster.InitialCluster) > 0
```
localvar, line 179: is it goroutine safe? a mutex is used in `SetInitialCluster` to update `opt.cluster.InitialCluster`.

pkg/cluster/cluster.go
```diff
@@ -732,9 +750,12 @@ func (c *cluster) defrag() {
 				defragInterval = defragFailedInterval
 				logger.Errorf("defrag failed: get client failed: %v", err)
 			}
-
+			defragmentURL := c.opt.ClusterAdvertiseClientURLs[0]
+			if c.opt.UseInitialCluster() {
+				defragmentURL = c.opt.Cluster.AdvertiseClientURLs[0]
+			}
```
localvar, line 756: is it possible that `ClusterAdvertiseClientURLs` be nil?

localvar (approved):

xxx7xxxx (approved):

PR #385, opened 2021-11-26
add doc for websocket

suchen-sci:


doc/cookbook/websocket.md
```diff
@@ -0,0 +1,77 @@
...
+
+    ```none
+    +--------------+                +--------------+                +--------------+  
+    |              |  1             |              |   2            |              | 
+    |   client     +--------------->|  Easegress   +--------------->|  websocket   |
+    |              |<---------------+(WebSocketSvr)|<---------------+  backend     | 
+    |              |  4             |              |   3            |              |
+    +--------------+                +--------------+                +--------------+
+    ```
+
```
samutamm, line 66: Could we include a minimal demo that shows one request and output? For example using `curl`, like in some of the cookbook chapters.

Also what is `websocket backend` in the graph above? Does every object and filter support websocket Server or is it only some of them?

suchen-sci: Use `curl` to do demo may be not enough, because we also need a backend server. Our `WebSocketServer` is proxy used to connect client and their server. But we do provide test cases for demo. Which can be find in `websocket_test.go` `TestWebSocket`. I will update this to doc. 

`Websocket backend` is the custom's server. And our proxy connect their client and their backend. I think the name `WebSocketServer ` is confusing. Maybe change it to be `WebSocketProxy`. But change the name of business controller may not downward compatibility.

samutamm: Ah ok, I got it now! The example looks good

samutamm (approved):

localvar (approved):

PR #389, opened 2021-11-30
update session delete http method

suchen-sci:


pkg/object/mqttproxy/mqtt_test.go
```diff
@@ -1895,7 +1895,7 @@ func TestHTTPDeleteSession(t *testing.T) {
 	if err != nil {
 		t.Errorf("marshal http session %v failed, %v", data, err)
 	}
-	req, _ := http.NewRequest(http.MethodPost, "http://localhost:8888/delete/session", bytes.NewBuffer(jsonData))
+	req, _ := http.NewRequest(http.MethodDelete, "http://localhost:8888/delete/session", bytes.NewBuffer(jsonData))
```
zhao-kun, line 1898: `http://localhost:8888/delete/session` is a typo?

suchen-sci: In test, path is different. In MQTTProxy, we registry by using Easegress api. In test, I just write these path bu their meaning. I will update it, make path in test and mqttproxy same.

zhao-kun (approved):

PR #391, opened 2021-12-01
support acme (let's encrypt) (close #2)

### LINKED ISSUE #2: Support ACME of Let's Encrypt
## Background

As we discussed in the technical board[1], we decide to support automating certificates in let's encrypt.

## Proposal

We should support features:

- Importing existed ACME certificates
- Automatically Request, renew ACME certificates
- Distributed storage for all certificates

### Config

```yaml
kind: ACME
name: acme-example

```

The `certBase64` and `keyBase64` is for imported ACME certificates.

---

```yaml
kind: HTTPServer
name: http-server-example
acme: 
  email: example@megaease.com # required while certBase64 and keyBase64 are empty
  renewTimeout: 60d #optional
  caServer: https://acme-v02.api.letsencrypt.org/directory # opional [2]
  domains: ['megaease.com'] # required
  httpChallenge: {}
  tlsChanllege: {} # conflict with httpChallenge
  certBase64: ... # optional
  keyBase64: ... # optional
  keyType: RSA4096
# ...
```

### Expected Behavior

The priority for certificates of HTTPServer:

1. ACME imported certificates
2. ACME generated certificates
3. Own certificates

## Reference

[1] https://docs.google.com/document/d/1gKM6uV3zzjPdPjSdhhQ0eqROaJppgg38CduY-BOjVNU/edit#
[2] https://letsencrypt.org/docs/acme-protocol-updates/#acme-v2-rfc-8555

localvar:


pkg/object/httpserver/mux.go
```diff
@@ -352,6 +353,14 @@ func (m *mux) reloadRules(superSpec *supervisor.Spec, muxMapper protocol.MuxMapp
 }
 
 func (m *mux) ServeHTTP(stdw http.ResponseWriter, stdr *http.Request) {
+	// HTTP-01 challenges requires HTTP server to listen on port 80, but we don't
+	// known which HTTP server listen on this port (consider there's an nginx sit in
```
samutamm, line 357:
```suggestion
	// known which HTTP server listen on this port (consider there's an nginx sitting in
```

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,417 @@
...
+
+// Init initializes AutoCertManager.
+func (acm *AutoCertManager) Init(superSpec *supervisor.Spec) {
+	acm.superSpec = superSpec
+	acm.spec = superSpec.ObjectSpec().(*Spec)
+	acm.super = superSpec.Super()
+
+	// TODO: remove this check after converting AutoCertManager to a system controller
+	if globalACM.Load() != nil {
+		logger.Warnf("an AutoCertManager instance is already exist")
```
suchen-sci, line 170: Is here use panic better? Force custom to use `egctl object update` not `egctl object create` when `AutoCertManager` already created?

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,417 @@
...
+		}
+	}
+	return nil
+}
+
+func (acm *AutoCertManager) reload() {
+	acm.stopCtx, acm.cancel = context.WithCancel(context.Background())
+	acm.storage = newStorage(acm.super.Cluster())
+
+	acm.renewBefore, _ = time.ParseDuration(acm.spec.RenewBefore)
```
suchen-sci, line 208: log an error here? maybe custom want their renew time but write it in wrong format...

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,417 @@
...
+		status.Domains = append(status.Domains, CertificateStatus{
+			Name:       d.Name,
+			ExpireTime: d.certExpireTime(),
+		})
+	}
+	return &supervisor.Status{ObjectStatus: status}
+}
+
+// Close closes AutoCertManager.
+func (acm *AutoCertManager) Close() {
```
suchen-sci, line 256: When close, do we need to delete the database?

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,417 @@
...
+		if err != nil {
+			name = spec.Name
+			logger.Warnf("domain name contains invalid characters: %s", name)
+		}
+
+		d := &acm.domains[i]
+		d.DomainSpec = spec
+		d.nameInPunyCode = name
+
+		cert, err := acm.storage.getCert(d.nameInPunyCode)
```
suchen-sci, line 229: In `reload`, is it a good idea to `putCert` before `getCert`?

pkg/object/autocertmanager/autocertmanager.go
localvar: this issue is tricky.
suppose we have `autocertmanager1` and want to create `autocertmanager2`, 
according to current logic, even if we `panic` here, the spec of `autocertmanager2` will be stored into Etcd,
because we don't guarantee object creation order during startup, when we do restart, it may successfully create `autocertmanager2` but panic when creating `autocertmanager1`.

as both behaviors are incorrect and we will change `AutoCertManager` to a system controller later, we can just keep this warning message for now.

pkg/object/autocertmanager/autocertmanager.go
suchen-sci: In `storage.go` we put some data into cluster (etcd). When do `inherit`, it's ok. But when customer delete the object, `Close` method will be called, do we need to delete all data that put into etcd?

pkg/object/autocertmanager/autocertmanager.go
suchen-sci: Cool! The code is a little bit complicated, I miss this part....

pkg/object/autocertmanager/autocertmanager.go
suchen-sci: Make sense!

pkg/object/autocertmanager/autocertmanager.go
localvar: `Close` is also called at program exit. And some CAs may have a limit of certificate renewal rate, so better to keep the data as it may be reused later.

pkg/object/autocertmanager/autocertmanager.go
suchen-sci: Make sense.... Another little question. I see that `globalACM` only store value, not store nil when close. Will this cause problem?

localvar: per the doc of [atomic.Value](https://golang.google.cn/pkg/sync/atomic/#Value.Store), putting nil into it will result in a panic.

samutamm (approved):

pkg/object/autocertmanager/autocertmanager.go
localvar: updated, will clear the `globalACM` when there's no AutoCertManager instance now.

suchen-sci (approved):

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,435 @@
...
+
+	"github.com/megaease/easegress/pkg/logger"
+	"github.com/megaease/easegress/pkg/supervisor"
+	"golang.org/x/crypto/acme"
+	"golang.org/x/net/idna"
+)
+
+const (
+	// Category is the category of AutoCertManager.
+	// It is a business controller by now, but should be a system controller
```
xxx7xxxx, line 40:
```suggestion
	// It is a business controller by now, but should be a system controller.
```

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,435 @@
...
+	Category = supervisor.CategoryBusinessController
+
+	// Kind is the kind of AutoCertManager.
+	Kind = "AutoCertManager"
+)
+
+type (
+	//AutoCertManager is the controller for Automated Certificate Management
```
xxx7xxxx, line 48:
```suggestion
	//AutoCertManager is the controller for Automated Certificate Management.
```

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -0,0 +1,435 @@
...
+		DirectoryURL    string       `yaml:"directoryUrl" jsonschema:"required,format=url"`
+		Email           string       `yaml:"email" jsonschema:"required,format=email"`
+		RenewBefore     string       `yaml:"renewBefore" jsonschema:"required,format=duration"`
+		EnableHTTP01    bool         `yaml:"enableHTTP01"`
+		EnableTLSALPN01 bool         `yaml:"enableTLSALPN01"`
+		EnableDNS01     bool         `yaml:"enableDNS01"`
+		Domains         []DomainSpec `yaml:"domains" jsonschema:"required"`
+	}
+
+	// DomainSpec is the automate certificate management spec for a domain
```
xxx7xxxx, line 74:
```suggestion
	// DomainSpec is the automated certificate management spec for a domain.
```

pkg/object/httpserver/mux.go
```diff
@@ -352,6 +353,14 @@ func (m *mux) reloadRules(superSpec *supervisor.Spec, muxMapper protocol.MuxMapp
 }
 
 func (m *mux) ServeHTTP(stdw http.ResponseWriter, stdr *http.Request) {
+	// HTTP-01 challenges requires HTTP server to listen on port 80, but we don't
+	// know which HTTP server listen on this port (consider there's an nginx sitting
+	// in front of Easegress), so all HTTP servers need to handle HTTP-01 challenges.
+	if strings.HasPrefix(stdr.URL.Path, "/.well-known/acme-challenge/") {
```
xxx7xxxx, line 359: Actually, we can get the port from `rules.spec.Port`.

localvar: We cannot use this configuration here. The port needs to be accessible from the CA server,  so even `rules.spec.Port` is 80, it does not mean the CA server could send a request to this HTTP server, because other reverse proxies like Nginx could get the request and forward it to another HTTP server who does not listen on port 80.

doc/controllers.md
```diff
@@ -301,6 +303,36 @@ servers:
...
+  - name: "*.megaease.com"
+    dnsProvider:
+      name: alidns
+      zone: megaease.com
+```
+
+| Name            | Type                                       | Description                                                                          | Required                           |
+| --------------- | ------------------------------------------ | ------------------------------------------------------------------------------------ | ---------------------------------- |
+| email           | string                                     | An email address for CA account                                                      | Yes                                |
+| directoryUrl    | string                                     | The endpoint of the CA directory                                                     | No (default to use Let's Encrypt)  |
```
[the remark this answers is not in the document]
localvar: updated.

pkg/object/httpserver/mux.go
xxx7xxxx: Oh we discussed this topic before, sure.

xxx7xxxx (approved):

PR #392, opened 2021-12-01
Add service canary

xxx7xxxx:
(Will make actions correct after https://github.com/megaease/easemesh-api/pull/16 merged)

- Support multiple canaries (deprecate the first version)
- Change label separator `&` to more idiomatic `,`.

### DIFF
--- a/pkg/object/meshcontroller/master/master.go
+++ b/pkg/object/meshcontroller/master/master.go
@@ -18,7 +18,9 @@
 package master
 
 import (
+	"fmt"
 	"runtime/debug"
+	"strings"
 	"time"
 
 	"gopkg.in/yaml.v2"
@@ -54,9 +56,20 @@ type (
 		done chan struct{}
 	}
 
+	serviceInstances []*spec.ServiceInstanceSpec
+
 	// Status is the status of mesh master.
 	Status struct{}
 )
+
+func (s serviceInstances) String() string {
+	instances := make([]string, len(s))
+	for i, instance := range s {
+		instances[i] = fmt.Sprintf("%s/%s", instance.ServiceName, instance.InstanceID)
+	}
+
+	return fmt.Sprintf("[%s]", strings.Join(instances, ","))
+}
 
 // New creates a mesh master.
 func New(superSpec *supervisor.Spec) *Master {
@@ -197,27 +210,34 @@ func (m *Master) scanInstances() (failedInstances []*spec.ServiceInstanceSpec,
 			}
 			gap := now.Sub(lastHeartbeatTime)
 			if gap > m.maxHeartbeatTimeout {
-				// This instance record's time gap is beyond our tolerance, needs to be clean immediately.
-				// For freeing storage space
-				if gap > defaultDeadRecordExistTime {
-					logger.Errorf("%s/%s expired for %s, need to be deleted", _spec.ServiceName, _spec.InstanceID, gap.String())
-					deadInstances = append(deadInstances, _spec)
-				} else if _spec.Status != spec.ServiceStatusOutOfService {
-					logger.Errorf("%s/%s expired for %s", _spec.ServiceName, _spec.InstanceID, gap.String())
+				if _spec.Status != spec.ServiceStatusOutOfService {
 					failedInstances = append(failedInstances, _spec)
+				}
+
+				if gap > defaultDeadRecordExistTime {
+					deadInstances = append(deadInstances, _spec)
 				}
 			} else {
 				if _spec.Status == spec.ServiceStatusOutOfService {
-					logger.Infof("%s/%s heartbeat recovered, make it UP", _spec.ServiceName, _spec.InstanceID)
 					rebornInstances = append(rebornInstances, _spec)
 				}
 			}
 		} else {
-			logger.Errorf("status of %s/%s not found, need to delete", _spec.ServiceName, _spec.InstanceID)
 			failedInstances = append(failedInstances, _spec)
 			deadInstances = append(deadInstances, _spec)
 		}
 	}
+
+	if len(failedInstances) > 0 {
+		logger.Warnf("failed instances: %s", serviceInstances(failedInstances))
+	}
+	if len(deadInstances) > 0 {
+		logger.Warnf("dead instances: %s", serviceInstances(deadInstances))
+	}
+	if len(rebornInstances) > 0 {
+		logger.Warnf("reborn instances: %s", serviceInstances(rebornInstances))
+	}
+
 	return
 }
 
--- a/pkg/object/meshcontroller/master/master.go
+++ b/pkg/object/meshcontroller/master/master.go
@@ -235,7 +235,7 @@ func (m *Master) scanInstances() (failedInstances []*spec.ServiceInstanceSpec,
 		logger.Warnf("dead instances: %s", serviceInstances(deadInstances))
 	}
 	if len(rebornInstances) > 0 {
-		logger.Warnf("reborn instances: %s", serviceInstances(rebornInstances))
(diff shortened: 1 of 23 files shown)


pkg/object/meshcontroller/spec/spec.go
```diff
@@ -629,114 +672,144 @@ func (b *pipelineSpecBuilder) appendTimeLimiter(tl *timelimiter.Spec) *pipelineS
...
 
-	if lb == nil {
-		lb = &proxy.LoadBalance{
-			Policy: proxy.PolicyRoundRobin,
+	makeServer := func(instance *ServiceInstanceSpec) *proxy.Server {
+		var url string
+		if needMTLS {
+			url = fmt.Sprintf("https://%s:%d", instance.IP, instance.Port)
+		} else {
+			url = fmt.Sprintf("http://%s:%d", instance.IP, instance.Port)
```
samutamm, line 713: Nit: only difference between line 711 and 713 is `s`, maybe this if-else could select protocol and then construct url like `fmt.Sprintf("%s://%s:%d", protocol, instance.IP, instance.Port)`

pkg/object/meshcontroller/worker/worker.go
```diff
@@ -70,28 +70,24 @@ type (
...
-		logger.Errorf("query unescape: %s failed: %v ", labels, err)
-		return mLabels
+func decodeLabels(labelStr string) map[string]string {
+	labelMap := make(map[string]string)
+	if len(labelStr) == 0 {
+		return labelMap
 	}
 
-	arrLabel := strings.Split(strLabel, "&")
+	labelSlice := strings.Split(labelStr, ",")
```
samutamm, line 79: This will not break existing configs, right?

xxx7xxxx: The older one is not idiomatic and weird in the config, this aims to fix it. I had thought about replacing `&` with `,` in the beginning of the function. But I didn't do it to bring more historic stuff.

xxx7xxxx:
Updated as comment @samutamm

pkg/object/meshcontroller/master/master.go
```diff
@@ -197,27 +210,34 @@ func (m *Master) scanInstances() (failedInstances []*spec.ServiceInstanceSpec,
...
 	}
+
+	if len(failedInstances) > 0 {
+		logger.Warnf("failed instances: %s", serviceInstances(failedInstances))
+	}
+	if len(deadInstances) > 0 {
+		logger.Warnf("dead instances: %s", serviceInstances(deadInstances))
+	}
+	if len(rebornInstances) > 0 {
+		logger.Warnf("reborn instances: %s", serviceInstances(rebornInstances))
```
localvar, line 238: should this log message be a `info` ?

pkg/object/meshcontroller/service/service.go
```diff
@@ -942,3 +943,90 @@ func (s *Service) PutTrafficTarget(tt *spec.TrafficTarget) {
...
+}
+
+// GetServiceCanary gets the service canary.
+func (s *Service) GetServiceCanary(serviceCanaryName string) *spec.ServiceCanary {
+	serviceCanary, _ := s.GetServiceCanaryWithInfo(serviceCanaryName)
+	return serviceCanary
+}
+
+// GetServiceCanaryWithInfo gets the service canary with raw info.
+func (s *Service) GetServiceCanaryWithInfo(serviceCanaryName string) (*spec.ServiceCanary, *mvccpb.KeyValue) {
```
localvar, line 967: this function is only called in `GetServiceCanary` and `WithInfo` is not good name, I propose to remove this function and only keep `GetServiceCanary`

pkg/object/meshcontroller/spec/spec.go
```diff
@@ -175,6 +181,22 @@ type (
 		URLs                  []*urlrule.URLRule              `yaml:"urls" jsonschema:"required"`
 	}
 
+	// ServiceCanary is the service canary entry.
+	ServiceCanary struct {
+		Name string `yaml:"name" jsonschema:"required"`
+		// Priority must be [1, 9], the default is 5 if user does not set it.
+		// The smaller number get higher priority.
+		// The order is sorted by name alphabetically in the same priority.
```
localvar, line 189: why limit priority to [1, 9], is the range large enough?

pkg/object/meshcontroller/spec/spec.go
```diff
@@ -414,6 +436,27 @@ type (
...
 
+func (sc ServiceCanary) Validate() error {
+	if sc.Priority < 0 || sc.Priority > 9 {
+		return fmt.Errorf("invalid priority (range is [0, 9], the default 0 will be set to 5)")
+	}
+
+	return nil
+}
+
+// Clone clonse TrafficRules.
```
localvar, line 447:
```suggestion
// Clone clones TrafficRules.
```

pkg/util/httpfilter/httpfilter.go
```diff
@@ -113,23 +114,35 @@ func (hf *HTTPFilter) Filter(ctx context.HTTPContext) bool {
 
 func (hf *HTTPFilter) filterHeader(ctx context.HTTPContext) bool {
 	h := ctx.Request().Header()
-	headerMatch := false
-	for key, vf := range hf.spec.Headers {
-		if headerMatch {
-			break
-		}
-
+	headerMatchNum := 0
```
localvar, line 117: the change in this function impact performance.
Now, when `MatchAllHeaders` is false, the code will continue to match other headers when one header is matched.

pkg/object/meshcontroller/spec/spec.go
xxx7xxxx: Enough for our scenarios. We leverage the OS scheduling system plus explicit two-layer sorting. Infinite priority with integer causes a bad user experience.

pkg/util/httpfilter/httpfilter.go
xxx7xxxx: Updated with a pre-checking.

pkg/object/meshcontroller/service/service.go
xxx7xxxx: Updated

pkg/object/meshcontroller/master/master.go
xxx7xxxx: Sure

pkg/object/meshcontroller/spec/spec.go
xxx7xxxx: Corrected

localvar (approved):

xxx7xxxx:
I resolved all conflicts, please review @localvar @samutamm @zhao-kun

pkg/object/meshcontroller/worker/egress.go
```diff
@@ -226,6 +234,13 @@ func (egs *EgressServer) reloadByTrafficTargets(value map[string]*spec.TrafficTa
 	return true
 }
 
+func (egs *EgressServer) reloadByServiceCanaries(value map[string]*spec.ServiceCanary) bool {
+	select {
+	case egs.chReloadEvent <- struct{}{}:
+	}
```
suchen-sci, line 240:
```suggestion
	egs.chReloadEvent <- struct{}{}
```

suchen-sci: Got it. My static checker just report this. FYI, https://staticcheck.io/docs/checks#S1000 "should use a simple channel send/receive instead of select with a single case (S1000) go-staticcheck".

xxx7xxxx: Updated with correct select statement

suchen-sci (approved):

samutamm (approved):

PR #394, opened 2021-12-02
update mqtt watcher disconnect problem

suchen-sci:
`Cluster.Watcher().Watch()` may disconnect due to multiple problems. 
So, when disconnect, we reconnect watcher and update events during reconnect period.


pkg/logger/export.go
```diff
@@ -141,3 +142,22 @@ func NginxHTTPAccess(remoteAddr, proto, method, path, referer, agent, realIP str
...
+
+func getSpanTemplate(context *model.SpanContext, template string) string {
+	if context == nil {
+		return "tid=" + "<nil>" + " sid=" + "<nil>" + " " + template
+	}
+	return "tid=" + context.TraceID.String() + " sid=" + context.ID.String() + " " + template
+}
+
+// SpanDebugf is the wrapper of default logger Debugf to log tracing message
+func SpanDebugf(context *model.SpanContext, template string, args ...interface{}) {
```
xxx7xxxx, line 154: Why don't you create a dedicated log file, the span storm will flood the standard logger.

suchen-sci: The `SpanDebugf ` should behavior just like `Debugf` which use `defaultlogger`. You can see this as a wrapper of `Debugf`, we need these logs print to same place. If don't write it in this way, we may use `logger.Debugf` but change template string and args...

xxx7xxxx: Got it

Issue #407, opened 2021-12-10
Proxy with mirrorPool panics under high QPS


samutamm:
**Describe the bug**

I created HTTPPipeline with Proxy filter that has `mirrorPool` pointing to an [echo server](example/backend-service/echo/echo.go) and `mainPool` pointing to another echo server. I can query HTTPServer with `curl` without any problem, but if I send many queries in parallel with [ab](https://httpd.apache.org/docs/2.4/programs/ab.html) tool, there is an `panic: close of closed channel` error after few seconds.

**To Reproduce**
1. Start Easegress server `bin/easegress-server -f eg-server.yaml`.
2. Add  HTTPServer with `egctl object create -f httpserver.yaml`
3. Add HTTPPipeline with `egctl object create -f mirrorpool-pipeline.yaml`
4. Start [echo server](example/backend-service/echo/echo.go) at HOST2 and HOST3.
5. Send queries in parallel using `ab`:
```bash
ab -k -c 450 -n 1000000 -H "X-Mirror: mirror"  http://{HOST1}:10080/pipeline
```
After few second Easegress server fails with `panic: close of closed channel`.

**Expected behavior**

**Version**
1.4.0

**Configuration**
* Easegress Configuration
```yaml
# eg-server.yaml
name: machine-1
cluster-name: cluster-stress-test
cluster-role: primary
api-addr: localhost:2381
data-dir: ./data
wal-dir: ""
cpu-profile-file: "./cpu-profile.txt"
memory-profile-file: "./memory-profile.txt"
disable-access-log: true
log-dir: ./log
debug: false
cluster:
  listen-peer-urls:
   - http://{HOST1}:2380
  listen-client-urls:
   - http://{HOST1}:2379
  advertise-client-urls:
   - http://{HOST1}:2379
  initial-advertise-peer-urls:
   - http://{HOST1}:2380
  initial-cluster:
   - machine-1: http://{HOST1}:2380
```

* HTTP server configuration
```yaml
# httpserver.yaml
kind: HTTPServer
name: stress-server
port: 10080
keepAlive: true
https: false
maxConnections: 10240
rules:
  - paths:
    - pathPrefix: /pipeline
      backend: mirrorpool-pipeline
```

* Pipeline Configuration
```yaml
name: mirrorpool-pipeline
kind: HTTPPipeline
flow:
  - filter: proxy
filters:
  - kind: Proxy
    name: proxy
    mainPool:
      loadBalance:
        policy: roundRobin
      servers:
[...]

PR #408, opened 2021-12-10
Fix master slave reader panic in concurrent

xxx7xxxx:
Fix https://github.com/megaease/easegress/issues/407

### DIFF
--- a/pkg/filter/proxy/masterslavereader.go
+++ b/pkg/filter/proxy/masterslavereader.go
@@ -20,6 +20,7 @@ package proxy
 import (
 	"bytes"
 	"io"
+	"sync"
 )
 
 type (
@@ -32,8 +33,9 @@ type (
 	}
 
 	masterReader struct {
-		r        io.Reader
-		buffChan chan []byte
+		r             io.Reader
+		buffChan      chan []byte
+		closeChanOnce sync.Once
 	}
 
 	slaveReader struct {
@@ -43,7 +45,7 @@ type (
 )
 
 func newMasterSlaveReader(r io.Reader) (io.ReadCloser, io.Reader) {
-	buffChan := make(chan []byte, 10)
+	buffChan := make(chan []byte, 100)
 	mr := &masterReader{
 		r:        r,
 		buffChan: buffChan,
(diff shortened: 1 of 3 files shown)


samutamm (approved):
This seems to fix https://github.com/megaease/easegress/issues/407. LGTM

suchen-sci (approved):
LGTM!

localvar:
I think this PR could eliminate the issue, but seems the fix is not for the root cause. I'd like to take some time to understand more details about #407 before approval.

suchen-sci:
@localvar Yeah, there are two possible reasons that cause this problem. 1. more than one goroutines call `Read` function. 2. one goroutine call `Read` even `Read` return error. But after carefully check the code, I don't find such places. Another possible place to cause this maybe in stdlib.... Please let me known if you have any idea about this....

megaeasex (changes requested):
[TASK:easegress-PR-test FAILED]megaease/easegress Pull Request 408 Deploy Test failed

pkg/filter/proxy/masterslavereader.go
```diff
@@ -66,7 +68,9 @@ func (mr *masterReader) Read(p []byte) (n int, err error) {
 	}
 
 	if err == io.EOF {
-		close(mr.buffChan)
+		mr.closeChanOnce.Do(func() {
+			close(mr.buffChan)
+		})
```
suchen-sci, line 73: If non-EOF error happens, do we need to take care of `mr.buffChan`?

qdongxu:
f```
unc (t *teeReader) Read(p []byte) (n int, err error) {
   n, err = t.r.Read(p)
   if n > 0 {
      if n, err := t.w.Write(p[:n]); err != nil {
         return n, err
      }
   }
   return
}

this is probably the cause, the teeReader returns an error only when n > 0.  It ignores the err = err.EOF  when n == 0.

```
> @localvar Yeah, there are two possible reasons that cause this problem. 1. more than one goroutines call `Read` function. 2. one goroutine call `Read` even `Read` return error. But after carefully check the code, I don't find such places. Another possible place to cause this maybe in stdlib.... Please let me known if you have any idea about this....

pkg/filter/proxy/masterslavereader.go
xxx7xxxx: No

xxx7xxxx:
Fixed by another PR https://github.com/megaease/easegress/pull/411

PR #410, opened 2021-12-14
fix tiny issue in docs

JackLeeHal:
fix tiny issue in docs I found when following the cook book.


localvar (approved):

suchen-sci (approved):
Thanks for the fix!

PR #411, opened 2021-12-15
Enable calling mirrorPool reader Read after EOF

samutamm:
- based on the discussion, fix https://github.com/megaease/easegress/issues/407 and return EOF at following Read calls after seeing EOF first time
- rename `masterSlaveReader` to `primarySecondaryReader`


pkg/filter/proxy/masterslavereader.go
```diff
@@ -23,40 +23,45 @@ import (
...
 
 type (
-	// masterSlaveReader reads bytes to master,
+	// primarySecondaryReader reads bytes to master,
 	// and synchronize them to slave.
 	// Currently only support one slave.
-	masterSlaveReader struct {
-		masterReader io.Reader
-		slaveReader  io.Reader
+	primarySecondaryReader struct {
```
localvar, line 29: It seems no code is using this struct, if yes, please remove it, if no, please update the comment according to the rename.

pkg/filter/proxy/proxy.go
```diff
@@ -317,7 +317,7 @@ func (b *Proxy) Handle(ctx context.HTTPContext) (result string) {
 
 func (b *Proxy) handle(ctx context.HTTPContext) (result string) {
 	if b.mirrorPool != nil && b.mirrorPool.filter.Filter(ctx) {
-		master, slave := newMasterSlaveReader(ctx.Request().Body())
+		master, slave := newPrimarySecondaryReader(ctx.Request().Body())
```
localvar, line 320: please also rename the local variables.

pkg/filter/proxy/masterslavereader.go
samutamm: It is not used indeed, let's delete it

localvar (approved):

suchen-sci:
Emmmm, based on golang side, "Looks like it's transferWriter probing to determine if it should send a chunked request", which means `transferWriter` may resend the request and call `Read` again. If that is true, will that means two goroutines call `Read` under high QPS? Do we need to use `sync.Once`? So, let me do some experiments about this....

localvar:
@suchen-sci , I don't think two goroutines could call `Read` concurrently, but if it do happens, I'm sure it is a bug of the Go standard library, because `io.Reader.Read` is not required to be go-routine safe.

And as I mentioned in the slack discussion,  if there are concurrent calls, use `sync.Once` cannot fix the issue either.

suchen-sci:
@localvar yeah, you are right. Golang use a writeCh to write these request, although they resend request during high QPS, it still keep thread-safe. So, `sync.Once` is unnecessary. But I still think `Read` after EOF in standard lib is wrong. If there are request body, first time it read the body and reach EOF. Second time, `transferWriter` do resend for chunked request (of course this is a hypothesis from golang issue comments), and read body only get EOF and empty body... And in source code, i don't find any place to re-read after receive EOF... But anyway, it is goroutine-safe.

suchen-sci (approved):

xxx7xxxx (approved):

PR #414, opened 2021-12-17
Improve performance of Proxy filter

samutamm:
## Changes

This PR proposes small performance improvements mainly in `pkg/filter/proxy`, `pkg/context/contexttest/context.go`, `pkg/tracing`, `pkg/logger` and some of the `pkg/utils`.

These changes mainly consist of:
- reducing string concatenations (`LazyTag` for doing string concatenations later)
- reducing dynamic memory allocations using `sync.Pool`
- pre-allocating slices with `make`

## Stress test scenario

Easegress server configuration
```yaml
name: machine-1
cluster-name: cluster-stress-test
cluster-role: primary
api-addr: localhost:2381
data-dir: ./data
wal-dir: ""
cpu-profile-file: "/home/ubuntu/easegress-benchmark/cpu-profile.txt"
memory-profile-file: "/home/ubuntu/easegress-benchmark/memory-profile.txt"
disable-access-log: false
log-dir: ./log
debug: false
cluster:
  listen-peer-urls:
   - http://HOST1:2380
  listen-client-urls:
   - http://HOST1:2379
  advertise-client-urls:
   - http://HOST1:2379
  initial-advertise-peer-urls:
   - http://HOST1:2380
  initial-cluster:
   - machine-1: http://HOST1:2380
```
HTTPServer:
```yaml
kind: HTTPServer
name: server-demo
port: 10080
keepAlive: true
https: false
maxConnections: 10240
rules:
  - paths:
    - pathPrefix: /pipeline
      backend: pipeline-demo
```
HTTPPipeline:
```yaml
name: pipeline-demo
kind: HTTPPipeline
flow:
  - filter: proxy
filters:
  - kind: Proxy
    name: proxy
    mainPool:
      loadBalance:
        policy: roundRobin
      servers:
      - url: http://HOST3:9095
      - url: http://HOST4:9095
```

Hosts 3 and 4 are running echo server on port 9095:
`go run example/backend-service/echo/echo.go > /dev/null`

HOST2 is sending concurrent requests :
`ab -k -c 300 -n 2000000  http://172.20.2.22:10080/pipeline > test_run_commithash.txt `

## Stress test results

Current main branch:
<details><summary><code>ab output</code> Output</summary><br><pre>
Server Software:        
Server Hostname:        172.20.2.22
Server Port:            10080

Document Path:          /pipeline
Document Length:        184 bytes

Concurrency Level:      300
Time taken for tests:   80.757 seconds
Complete requests:      2000000
Failed requests:        0
Keep-Alive requests:    2000000
Total transferred:      652000000 bytes
HTML transferred:       368000000 bytes
Requests per second:    24765.74 [#/sec] (mean)
Time per request:       12.114 [ms] (mean)
Time per request:       0.040 [ms] (mean, across all concurrent requests)
Transfer rate:          7884.41 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       5
Processing:    10   12   2.9     11     100
Waiting:       10   12   2.9     11     100
Total:         10   12   2.9     11     103

Percentage of the requests served within a certain time (ms)
  50%     11
  66%     11
  75%     12
  80%     13
  90%     16
  95%     19
  98%     21
  99%     22
 100%    103 (longest request)
 </pre></details>
mean QPS: **24765.74**

This branch:
<details><summary><code>ab output</code> Output</summary><br><pre>
Server Software:        
Server Hostname:        172.20.2.22
Server Port:            10080

Document Path:          /pipeline
Document Length:        184 bytes

Concurrency Level:      300
Time taken for tests:   79.047 seconds
Complete requests:      2000000
Failed requests:        0
Keep-Alive requests:    2000000
Total transferred:      652000000 bytes
HTML transferred:       368000000 bytes
Requests per second:    25301.52 [#/sec] (mean)
Time per request:       11.857 [ms] (mean)
Time per request:       0.040 [ms] (mean, across all concurrent requests)
Transfer rate:          8054.98 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       5
Processing:    10   12   2.8     11     117
Waiting:       10   12   2.8     11     117
Total:         10   12   2.8     11     120

Percentage of the requests served within a certain time (ms)
  50%     11
  66%     11
  75%     12
  80%     12
  90%     15
  95%     18
  98%     21
  99%     22
 100%    120 (longest request)
 </pre></details>
mean QPS: **25301.52**

Difference: (25301.52 - 24765.74) / 24765.74 = 0.0216 ~ **2.2 % increase in mean QPS**
It's not huge improvement but improvement still.

I tested different number of concurrent requests (-c param for ab tool) and this branch was better in all runs, even though the level of QPS seems to vary between tests and even between days.

I could also observe the impact of these changes in `go tool pprof` cpu profile. For example, introducing LazyTag reduces few seconds of string concatenations and Itoa processing and each `sync.Pool` removed time from runtime.mallocs.


pkg/context/httpcontext.go
```diff
@@ -206,10 +249,40 @@ func (ctx *httpContext) Span() tracing.Span {
...
+func (ctx *httpContext) GetTags() []string {
+	allTags := make([]string, len(ctx.tags)+len(ctx.lazyTags))
+	for i := 0; i < len(ctx.tags); i++ {
+		allTags[i] = ctx.tags[i]
+	}
+	for i := 0; i < len(ctx.lazyTags); i++ {
+		allTags[len(ctx.tags)+i] = ctx.lazyTags[i].String()
+	}
+	return allTags
+}
```
localvar, line 284: Please make sure no one cares about the order of tags, compare to the logic before, the order of the tags is different.

@xxx7xxxx could you please help confirm this?

pkg/context/httpcontext.go
```diff
@@ -206,10 +249,40 @@ func (ctx *httpContext) Span() tracing.Span {
 	return ctx.span
 }
 
+// Fill the empty tag. If all existing Tags are filled, create new one.
 func (ctx *httpContext) AddTag(tag string) {
 	ctx.tags = append(ctx.tags, tag)
 }
 
+// Add new LazyTag.
+func (ctx *httpContext) AddLazyTag(ns string, prefix string, msg string, intMsg int) {
```
localvar, line 258: the meaning and relationship of the parameters are unclear, could we change the parameter to a closure function (like what we do in `logger.LazyDebug`)

pkg/filter/proxy/request.go
```diff
@@ -32,26 +33,41 @@ import (
...
+	requestStatus int16
+
 	request struct {
 		server     *Server
 		std        *http.Request
 		statResult *httpstat.Result
 		createTime time.Time
 		_startTime *time.Time
 		_endTime   *time.Time
+		status     requestStatus
```
localvar, line 45: why add this field? could it improve performance?

propose to change `_startTime` and `_endTime` from `*time.Time` to `time.Time`, this helps avoid some memory allocations, and we can use `time.Time.IsZero()` to check if they are initialized.

samutamm: It does not improve the performance. I added it, because the previous logic did not work when `_startTime` was not nil anymore after being initialized once and recycled via the sync.Pool. Thanks for hint, I will change to `time.Time` and `IsZero`.

pkg/context/httpcontext.go
suchen-sci: the tags is used for debug... I think the order of tags is important, use tags we can see the filter exec order of pipeline. @samutamm Is that a good idea we use same struct for tags and lazy tags, so we can keep the order of tags is the order of filter exec?

pkg/context/httpcontext.go
suchen-sci, line 258: this tag is not very lazy... lazy usually means the program is only executed when the result is needed. some functional programming languages use this schema a lot...

samutamm: Yes, this LazyTag enables to postpone CPU heavy `stringtools.Cat` and `Itoa` operations away from the execution of the current request. Please note that [LazyTag.String](https://github.com/megaease/easegress/blob/f8ffb32564d589c30d3b71b661fe276ae7cd8d01/pkg/context/httpcontext.go#L180-L186) is executed only [here](https://github.com/megaease/easegress/blob/f8ffb32564d589c30d3b71b661fe276ae7cd8d01/pkg/context/httpcontext.go#L364), lazily. 

You are right that `AddLazyTag` does not avoid allocating the memory needed for the LazyTag struct, but this is where the `lazyTagPool` helps out.

pkg/context/httpcontext.go
```diff
@@ -206,10 +249,40 @@ func (ctx *httpContext) Span() tracing.Span {
...
+}
+
+func (ctx *httpContext) ReleaseLazyTags() {
+	for _, lazyTag := range ctx.lazyTags {
+		lazyTagPool.Put(lazyTag)
+	}
+}
+
+// Return all tags in string format.
+func (ctx *httpContext) GetTags() []string {
```
suchen-sci, line 275:
```
type Tag struct {
  tag string // used for add tag
  lazyTag func() string // used for lazy tag
}
```
is this a good idea? we combine tag and lazy tag together, so the order will not change...

pkg/context/httpcontext.go
```diff
@@ -59,6 +60,12 @@ type (
 
 		OnFinish(FinishFunc) // For setting final client statistics, etc.
 		AddTag(tag string)   // For debug, log, etc.
+		// Update next lazy tag, without creating new object if possible.
+		AddLazyTag(ns string, prefix string, msg string, intMsg int)
+		// Converts LazyTags to string and return them with normal tags.
+		GetTags() []string
+		// Recycle LazyTags back to lazyTagPool.
+		ReleaseLazyTags()
```
suchen-sci, line 68: Propose to change `GetTags`, `ReleaseLazyTags` to be `getTags`, `releaseLazyTags`. these methods are used for HTTPContext to debug. They may not allowed to be called outside. I think filters should not call these methods...

pkg/filter/proxy/request.go
samutamm: Well it turned out that `time.Time` is not reinitialized either when added back to pool. So the second `_startTime` does not return true for `time.Time.IsZero()`.
Here's a small Go playground example: https://go.dev/play/p/cvIuNhSe94M
One option would be manually reinitialize times, but there is no function time.Time.Zero() that would give the "_uninitialized_" value.

So I think it's better to keep this `status` field for now.

localvar: `status` is not reinitialized either, we can reset `startTime` and `endTime` in `pool.newRequest`.

Another issue with `status`  is that it introduces some kind of redundancy to existing `startTime` & `endTime`,  this could lead to inconsistency in the code maintenance.

localvar: time could be initialized by assigning a zero value, like:
```go
startTime = time.Time{}
```

pkg/context/httpcontext.go
samutamm: Now the order of the tags should be the same as before. There is only on slice for tags that is appended in order.

pkg/context/httpcontext.go
samutamm: This is a good idea. Even simpler is to use `func() string` type for the slice, so we don't need struct at all:

```go
// Add new Tag.
func (ctx *httpContext) AddTag(tag string) {
	ctx.lazyTags = append(ctx.lazyTags, func() string {
		return tag
	})
}

// Add new LazyTag.
func (ctx *httpContext) AddLazyTag(lazyTag LazyTagFunc) {
	ctx.lazyTags = append(ctx.lazyTags, lazyTag)
}
```

pkg/context/httpcontext.go
samutamm: Good point, thanks!

localvar (approved):

samutamm:
Here is the ab output and QPS comparaison between latest changes:

Main branch: 
<details><summary><code>ab output</code> Output</summary><br><pre>

This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 172.20.2.22 (be patient)


Server Software:        
Server Hostname:        172.20.2.22
Server Port:            10080

Document Path:          /pipeline
Document Length:        184 bytes

Concurrency Level:      300
Time taken for tests:   86.444 seconds
Complete requests:      2000000
Failed requests:        0
Keep-Alive requests:    2000000
Total transferred:      652000000 bytes
HTML transferred:       368000000 bytes
Requests per second:    23136.25 [#/sec] (mean)
Time per request:       12.967 [ms] (mean)
Time per request:       0.043 [ms] (mean, across all concurrent requests)
Transfer rate:          7365.64 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       5
Processing:    10   13   5.6     11     116
Waiting:       10   13   5.6     11     116
Total:         10   13   5.6     11     118

Percentage of the requests served within a certain time (ms)
  50%     11
  66%     11
  75%     12
  80%     12
  90%     20
  95%     28
  98%     33
  99%     35
 100%    118 (longest request)
  </pre></details>
 
mean QPS: **23136.25**

This branch: 
  <details><summary><code>ab output</code> Output</summary><br><pre>
  This is ApacheBench, Version 2.3 <$Revision: 1843412 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking 172.20.2.22 (be patient)


Server Software:        
Server Hostname:        172.20.2.22
Server Port:            10080

Document Path:          /pipeline
Document Length:        184 bytes

Concurrency Level:      300
Time taken for tests:   82.554 seconds
Complete requests:      2000000
Failed requests:        0
Keep-Alive requests:    2000000
Total transferred:      652000000 bytes
HTML transferred:       368000000 bytes
Requests per second:    24226.65 [#/sec] (mean)
Time per request:       12.383 [ms] (mean)
Time per request:       0.041 [ms] (mean, across all concurrent requests)
Transfer rate:          7712.78 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       5
Processing:    10   12   4.9     11     105
Waiting:       10   12   4.9     11     105
Total:         10   12   4.9     11     108

Percentage of the requests served within a certain time (ms)
  50%     11
  66%     11
  75%     11
  80%     12
  90%     15
  95%     26
  98%     32
  99%     34
 100%    108 (longest request)
   </pre></details>
  mean QPS: **24226.65**
  
  So about 4.7 % improvement on this test case, according these ab stress tests.

suchen-sci (approved):
LGTM!

pkg/filter/proxy/pool.go
```diff
@@ -127,11 +129,30 @@ func (p *pool) status() *PoolStatus {
...
+var httpstatResultPool = sync.Pool{
+	New: func() interface{} {
+		return &gohttpstat.Result{}
+	},
+}
+
 func (p *pool) handle(ctx context.HTTPContext, reqBody io.Reader, client *http.Client) string {
-	addTag := func(subPrefix, msg string) {
-		tag := stringtool.Cat(p.tagPrefix, "#", subPrefix, ": ", msg)
+	addLazyTag := func(subPrefix, msg string, intMsg int) {
```
xxx7xxxx, line 145: Why not just convert int into string outside `addLazyTag`. The extra message parameter would be a little confusing.

pkg/context/httpcontext.go
```diff
@@ -303,7 +335,7 @@ func (ctx *httpContext) Finish() {
 			fasttime.Format(ctx.startTime, fasttime.RFC3339Milli),
 			stdr.RemoteAddr, ctx.r.RealIP(), stdr.Method, stdr.RequestURI, stdr.Proto, ctx.w.code,
 			ctx.metric.Duration, ctx.r.Size(), ctx.w.Size(),
-			strings.Join(ctx.tags, " | "))
+			tags)
```
xxx7xxxx, line 338: Can the new tags keep the original format with separator `|`?

pkg/filter/proxy/pool.go
```diff
@@ -216,7 +242,9 @@ func (p *pool) doRequest(ctx context.HTTPContext, req *request, client *http.Cli
 	}
 
 	span := ctx.Span().NewChildWithStart(spanName, req.startTime())
-	span.Tracer().Inject(span.Context(), opentracing.HTTPHeaders, opentracing.HTTPHeadersCarrier(req.std.Header))
+	if !span.IsNoopSpan() {
+		span.Tracer().Inject(span.Context(), opentracing.HTTPHeaders, opentracing.HTTPHeadersCarrier(req.std.Header))
```
xxx7xxxx, line 246: Even this line was executed, it would be a dummy operation if the span is noopspan, right?

pkg/filter/proxy/pool.go
samutamm: Under very heavy load it takes a lot of time to convert integers to string, so idea here is to push the Itoa to logging, HTTPContext.OnFinish. Then when logging is disabled (for example when running stress tests), expensive string concatenations and Itoa are not executed and the performance increases slightly. 
But I agree that it's confusing, I will add an extra comment for that

pkg/context/httpcontext.go
samutamm: Yes, on line 322 they have same separator `|`

pkg/filter/proxy/pool.go
samutamm: Yes, it would be no-ops operations from here https://github.com/opentracing/opentracing-go/blob/master/noop.go

xxx7xxxx: So can this if-clause be omitted?

samutamm: Good point, I omitted it!

xxx7xxxx (approved):

Issue #413, opened 2021-12-16
ResourceExhausted error when creating many HTTPPipelines and HTTPServer paths


samutamm:
**Describe the bug**
Easegress server throws `2021-12-16T03:40:40.8Z	ERROR	statussynccontroller/statussynccontroller.go:217	sync status failed: rpc error: code = ResourceExhausted desc = trying to send message larger than max (2274303 vs. 2097152)`  when creating 1000 dummy HTTPPipelines and one HTTPServer with one backend rule for each pipeline.

**To Reproduce**
1. Generate 1000 identical HTTPPipeline configurations, with unique names `name: pipeline-$i`. Run the second script provided below: `bash generate_pipeline.sh > pipelines.yaml`
2. Generate one HTTPServer configuration with 1000 rules using the first script provided below: `bash generate_server.sh > httpserver.yaml`
3. Start Easegress server: `bin/easegress-server`
4. Create pipelines `bin/easegress-server object create -f pipelines.yaml`
5. Create httpserver `bin/easegress-server object create -f httpserver.yaml`
6. Easegress server will fail in the middle of the object creation.

**Expected behavior**
Easegress-server should not fail when creating (many) objects.

**Version**
1.4.0

**Configuration**
* Easegress Configuration
Default parameters.

* HTTP server configuration
The following bash script generates the server HTTPServer:
```bash
#!/bin/bash
echo "
kind: HTTPServer
name: server-demo
port: 10080
keepAlive: true
https: false
maxConnections: 10240
rules:
  - paths:
"
for i in {0..1000..1}
   do
      echo "    - pathPrefix: /pipeline$i
      backend: pipeline-$i"
done
```

* Pipeline Configuration
The following bash script generates the pipeline:
```bash
#!/bin/bash
for i in {0..1000..1}
   do
      echo "name: pipeline-$i
kind: HTTPPipeline
flow:
  - filter: proxy
filters:
  - kind: Proxy
    name: proxy
    mainPool:
      loadBalance:
        policy: roundRobin
      servers:
      - url: http://172.20.2.14:9095
      - url: http://172.20.2.160:9095
---"
done
```

**Logs**
This is the output of easegress-server when the error happens:
```bash
2021-12-16T04:45:22.857Z	INFO	trafficcontroller/trafficcontroller.go:424	create http pipeline default/pipeline-977
2021-12-16T04:45:22.858Z	INFO	trafficcontroller/trafficcontroller.go:424	create http pipeline default/pipeline-989
[...]

PR #419, opened 2021-12-20
Expose max-sync-message-size in options

samutamm:
Fix https://github.com/megaease/easegress/issues/413

### DIFF
--- a/pkg/object/statussynccontroller/statussynccontroller.go
+++ b/pkg/object/statussynccontroller/statussynccontroller.go
@@ -20,6 +20,7 @@ package statussynccontroller
 import (
 	"runtime/debug"
 	"sync"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 
@@ -214,6 +215,12 @@ func (ssc *StatusSyncController) syncStatusToCluster(statuses map[string]string)
 
 	err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kvs)
 	if err != nil {
+		resourceExhaustedErrMsg := "trying to send message larger than max "
+		if strings.Contains(err.Error(), resourceExhaustedErrMsg) {
+			logger.Errorf("Size of the Easegress objects is too large to be synchronized. " +
+				"Please reduce objects (output of 'egctl object list') or " +
+				"increase the value of cluster.MaxCallSendMsgSize in server configuration.")
+		}
 		logger.Errorf("sync status failed: %v", err)
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/option/option.go
```diff
@@ -46,6 +46,7 @@ type ClusterOptions struct {
 	StateFlag                string            `yaml:"state-flag"`
 	// Secondary members define URLs to connect to cluster formed by primary members.
 	PrimaryListenPeerURLs []string `yaml:"primary-listen-peer-urls"`
+	MaxSyncMessageSize    int      `yaml:"max-sync-message-size"`
```
localvar, line 49: Why use a new name `MaxSyncMessageSize` and not the name `MaxCallSendMsgSize` in Etcd?
I think `MaxCallSendMsgSize` is from GRPC, and Etcd reuse it.

samutamm: Good point, we should probably reuse it also. I'll fix it

suchen-sci:
If user create yaml file larger than `MaxCallSendMsgSize`, I think we better log that to help them find problem. One way to do that is in `statussynccontroller .go` line 217, when err is not nil, except logging the error. We can also check if the length of kvs is exceed `MaxCallSendMsgSize `. Is this a good idea?

suchen-sci (approved):

pkg/object/statussynccontroller/statussynccontroller.go
```diff
@@ -214,6 +215,12 @@ func (ssc *StatusSyncController) syncStatusToCluster(statuses map[string]string)
 
 	err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kvs)
 	if err != nil {
+		resourceExhaustedErrMsg := "trying to send message larger than max "
+		if strings.Contains(err.Error(), resourceExhaustedErrMsg) {
+			logger.Errorf("Size of the Easegress objects is too large to be synchronized. " +
+				"Please reduce objects (output of 'egctl object list') or " +
+				"increase the value of cluster.MaxCallSendMsgSize in server configuration.")
+		}
```
localvar, line 223: It is better not to rely on a specific string for the check, because we cannot get a notification if the string is changed at the Etcd side.

Propose to remove these code and update L224 to something like:

```go
 logger.Errorf("sync status failed, if the message size is too large, please increase the value of cluster.MaxCallSendMsgSize in configuration: %v", err)
```

localvar (approved):

PR #422, opened 2021-12-21
Update helm

suchen-sci:
This pr is mainly used to remove dependencies in our helm charts. So single directory in `helm-charts` can be used to do deployment.


helm-charts/easegress/README.md
```diff
@@ -58,3 +55,18 @@ helm uninstall easegress -n easegress
 kubectl delete pvc easegress-pv-easegress-0 -n easegress
 # same for easegress-pv-easegress-i...n
 ```
+
+## Parameters
+
+The following table lists the configurable parameters of the MegaEase Easegress.
```
samutamm, line 61:
```suggestion
The following table lists the configurable parameters of the MegaEase Easegress Helm installation.
```

helm-charts/easegress/README.md
```diff
@@ -58,3 +55,18 @@ helm uninstall easegress -n easegress
...
+
+| Key | Type | Default | Description |
```
samutamm, line 63: This table duplicates the information in `values.yaml` file, forcing to change two places, when something changes here. Given that [values.yaml](https://helm.sh/docs/chart_best_practices/values/) is common to all Helm installations, I would consider removing this table and only referencing values.yaml here, like `Helm parameters and default values are listed in values.yaml`.

helm-charts/easegress/templates/PersistentVolume.yaml
```diff
@@ -25,6 +25,12 @@ spec:
...
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: easegress-storage
+provisioner: kubernetes.io/no-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
 
-{{- end }}
+{{ end }}
```
samutamm, line 36: Nit: `{{- end}}` will not print extra empty line while `{{ end }}` does. Ref: [[controlling whitespace](https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace)]

helm-charts/ingress-controller/README.md
```diff
@@ -54,3 +51,18 @@ helm uninstall ingress-easegress -n ingress-easegress
 kubectl delete pvc easegress-pv-ingress-easegress-0 -n ingress-easegress
 # same for easegress-pv-ingress-easegress-i...n
 ```
+
+## Parameters
```
samutamm, line 55: ditto

helm-charts/ingress-controller/values.yaml
```diff
@@ -1,6 +1,6 @@
 service:
   # nodePort for easegress service
-  nodePort: 80
+  nodePort: 30080
```
samutamm, line 3: I would keep the Ingress port as 80 as it is the value for Kubernetes Ingress and [HTTP traffic in general](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers).

helm-charts/easegress/README.md
suchen-sci: in kubesphere, https://github.com/kubesphere/helm-charts/tree/master/src/ every readme has parameter part to show what the parameter are...

helm-charts/ingress-controller/values.yaml
suchen-sci: this is nodePort... the real port out k8s to access is defined by `Service`. nodePort is fine to be any number...

helm-charts/easegress/README.md
suchen-sci: So i will keep the parameters...

samutamm (approved):

localvar (approved):

Issue #431, opened 2021-12-23
easegress-server not report error and return with invalid yaml config file


suchen-sci:
**Describe the bug**
`easegress-server` not report error and return with invalid yaml config file.

**To Reproduce**
With invalid yaml `config.yaml`:
```
name: machine-1
cluster-name: cluster-test
cluster-role: primary
api-addr: localhost:2381
data-dir: ./data
wal-dir: ""
cpu-profile-file:
memory-profile-file:
log-dir: ./log
debug: false
cluster:
  listen-peer-urls: 
   - http://127.0.0.1:2380
  listen-client-urls:
   - http://127.0.0.1:2379
  advertise-client-urls:
   - http://127.0.0.1:2379
  initial-advertise-peer-urls:
   - http://127.0.0.1:2380
  initial-cluster: 
   - http://127.0.0.1:2380
```
run `easegress-server --config-file config.yaml` and `easegress-server` will not report any error and keep running. 

right yaml should be:
```
name: machine-1
...
cluster:
  ...
  initial-cluster: 
   - machine-1: http://127.0.0.1:2380
```

**Expected behavior**
With invalid yaml, `easegress-server` should report error and return. So that customers will not get confused.

**Version**
1.4.0

martinou:
can add test options,use valid  yaml

suchen-sci:
@martinou Hi, in pr #432 we fix this problem, now if use invalid yaml with invalid type, `easegress-server` will report error and return. If you ok with this, we will close this issue. By the way, what do you mean `add test options`? do you mean add unit test to `Option` or something else?

martinou:
@suchen-sci no,like nginx -t ./conf/nginx.conf test config file.not add testcase.

suchen-sci:
@martinou Hi, it sounds like `dry-run` in k8s. Thanks for your advice! Since this issue has been solved, we will close it.

PR #432, opened 2021-12-23
fix easegress-server keep running on invalid yaml

suchen-sci:
this pr fix issue #431. if user provide invalid yaml, then `easegress-server` will report error and return. 

ps: here invalid yaml means yaml file provide wrong type for option. For example, `option.cluster.initial-cluster` is a map[string]string, if we provide an array will cause problem.


suchen-sci:
in mqtt, we store session asynchronous. And this cause some mqtt test about session failed occasional (we check before session stored). this pr will fix them too.

localvar (approved):

samutamm (approved):

PR #441, opened 2021-12-24
fix httpserver httppipeline status not show error

### LINKED ISSUE #438: get object status return empty
**Describe the bug**
apis/v1/status/objects/server-demo return {}

**To Reproduce**
1. Execute '...'
2. Send a request to 'apis/v1/status/objects/server-demo'
3. See error
easegress return {}

**Expected behavior**
return object status

**Version**
1.4.0

**Configuration**
* Easegress Configuration
```
```

* HTTP server configuration
```autoCert: false
caCertBase64: ""
cacheSize: 0
certBase64: ""
certs: {}
http3: false
https: false
keepAlive: true
keepAliveTimeout: 60s
keyBase64: ""
keys: {}
kind: HTTPServer
maxConnections: 10240
name: server-demo
port: 6080
rules:
- host: ""
  hostRegexp: ""
  paths:
  - backend: pipeline-demo
    headers: []
    pathPrefix: /
    rewriteTarget: ""
tracing: null
xForwardedFor: false
```

* Pipeline Configuration
```yaml
```

**Logs**
```
Easegress logs, if applicable.
```

**OS and Hardware**
 - OS: centos 7.x
 - CPU:Intel(R) Core(TM) i5-8265U
 - Memory:  32GB

**Additional context**
postman invoke request  apis/v1/status/objects/server-demo

invoke apis/v1/status/objects/ return is ok

suchen-sci:
Fix #438 

return status is map -> map[member id]status... Since our `egctl` does not specific namespace now, we return status from all namespace. this part may need change.


samutamm:
I tested the scenario described in issue #438, using curl command `curl localhost:2381/apis/v1/status/objects/server-demo`. I got more or less the same output with main branch and this.. Would it be possible to list exact steps to reproduce the problem in main branch and then show that these changes fix it?

suchen-sci:
Hi, @samutamm after start up, you need wait some few seconds before request. `easegress` will collect status and store it to etcd. We get status from etcd. And if possible, please use `egctl object status get server-demo` command...

suchen-sci:
```
➜  easegress git:(fix-object-status) ✗ ./bin/egctl object status get server-demo
eg-default-name: |
  {}
➜  easegress git:(fix-object-status) ✗ ./bin/egctl object status get server-demo
eg-default-name: |
  {}
➜  easegress git:(fix-object-status) ✗ ./bin/egctl object status get server-demo
eg-default-name: |
  default: |
    health: ""
    state: running
    status:
      count: 0
      m1: 0
      m5: 0
      m15: 0
      errCount: 0
      m1Err: 0
      m5Err: 0
      m15Err: 0
      m1ErrPercent: 0
      m5ErrPercent: 0
      m15ErrPercent: 0
      min: 0
      max: 0
      mean: 0
      p25: 9.999999e+06
      p50: 9.999999e+06
      p75: 9.999999e+06
      p95: 9.999999e+06
      p98: 9.999999e+06
      p99: 9.999999e+06
      p999: 9.999999e+06
      reqSize: 0
      respSize: 0
      codes: {}
    topN: []
```
here are my results after wait about half minute...

xxx7xxxx:
Actually, the root cause here is that TrafficController took the control of the lifecycle of HTTPServer/HTTPPipeline, so it's intuitive that the statuses of servers/pipelines are reported by the TrafficController, where it will cause the mislocation between config and status of them. 

I'm fine with this quick fix without deep and tidy refactoring. But if there brought more kinds of stuff under TrafficController, we should fix them in a whole policy(TrafficController syncing or move the status to the corresponding location).

xxx7xxxx (approved):

pkg/api/cluster.go
```diff
@@ -164,3 +167,55 @@ func (s *Server) _listStatusObjects() map[string]map[string]interface{} {
...
+}
+
+func (s *Server) _getStatusObjectFromTrafficController(name string, spec *supervisor.Spec) map[string]string {
+	prefix := s.cluster.Layout().StatusObjectPrefix(trafficcontroller.Kind)
+	kvs, err := s.cluster.GetPrefix(prefix)
+	if err != nil {
+		ClusterPanic(err)
+	}
+	status := &trafficcontroller.Status{}
+	// ans := make(map[string]string)
```
samutamm, line 203: nit: commented code

suchen-sci: updated!

suchen-sci:
@xxx7xxxx @samutamm @localvar I updated the code, previous `egctl object status get` will return `map[member id]map[namespace]status`. now it will show `map[member id]status`. we only show the default namespace. Is this ok for you?

xxx7xxxx:
Returning status only from the default namespace is all right.

samutamm (approved):
LGTM!

xxx7xxxx (approved):

pkg/api/cluster.go
```diff
@@ -164,3 +168,53 @@ func (s *Server) _listStatusObjects() map[string]map[string]interface{} {
...
+		if spec.Kind() == httpserver.Kind {
+			if val, ok := ns.HTTPServers[spec.Name()]; ok {
+				b, err := yaml.Marshal(val.Status)
+				if err != nil {
+					ClusterPanic(fmt.Errorf("unmarshal %v to yaml failed: %v", val.Status, err))
+				}
+				return string(b)
+			}
+		}
+		if spec.Kind() == httppipeline.Kind {
```
localvar, line 187:
```suggestion
                     return ""
		} else if spec.Kind() == httppipeline.Kind {
```

pkg/api/cluster.go
```diff
@@ -164,3 +168,53 @@ func (s *Server) _listStatusObjects() map[string]map[string]interface{} {
...
+			if val, ok := ns.HTTPPipelines[spec.Name()]; ok {
+				b, err := yaml.Marshal(val.Status)
+				if err != nil {
+					ClusterPanic(fmt.Errorf("unmarshal %v to yaml failed: %v", val.Status, err))
+				}
+				return string(b)
+			}
+		}
```
localvar, line 195:
```suggestion
			}
			return ""
		}
```

localvar (approved):

PR #439, opened 2021-12-24
Reduce etcd server memory usage

samutamm:
- set snapshot count to lower value to reduce etcd ram usage https://etcd.io/docs/v3.5/tuning/#snapshot-tuning
- fix https://github.com/megaease/easegress/issues/420 ; do not print debug client logs if `--debug` flag is not provided for server

PR #442, opened 2021-12-27
udpate integration test

suchen-sci:
We update example yaml last week (use new cluster version of yaml). As a result, primary-001 can not used alone for integration test since it will wait all clusters to start up. This pr update integration test to use single node primary easegress to do the test.

update github workflow, run test when change yaml file and shell file since these changes may fail integration test.


localvar (approved):

samutamm (approved):
LGTM but just for reference we could also test 3 member cluster like this https://github.com/megaease/easegress/pull/439/commits/85cf29934c37fc8ffda18455e15a582faf3b6ff0

suchen-sci:
Thanks Samu, we may add that later. Test cluster may need some new tests. Objects like `HTTPServer` need open new port, so when run multiple easegress in one machine, some objects may failed to create.

PR #444, opened 2021-12-28
Simplify Ingress controller Helm-chart

samutamm:
## Background

IngressController Helm chart and `doc/reference/ingresscontroller.md` currently sets up IngressController in cluster mode, which is not necessary for Ingress use case. IngressController is stateless so it can be scaled up and down and pods can just start new Ingress controller instance when recovering from crashes.

Here's one IngressController running:

And here's multiple IngressControllers running:

When there are multiple IngressControllers, they run independently of each other. NodePort Service load balances the traffic to different IngressControllers.

## PR Changes
This PR simplifies the Helm charts and tutorial in order to have the previous architecture.  This PR
- changes IngressController from StatefulSet to Deployment
- removes PersistentVolumes and secondary easegress instances
- adapts the tutorial


helm-charts/ingress-controller/README.md
```diff
@@ -29,40 +21,24 @@ helm install ingress-easegress -n ingress-easegress ./helm-charts/ingress-contro
...
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| service.nodePort | int | `30080` | nodePort for easegress service. |
-| service.adminPort | int | `31255` | nodePort for egctl access. |
-| cluster.primaryReplicas | int | `1` | number of easegress service that persists cluster data to disk |
-| cluster.volumeType | string | `emptyDir` | `emptyDir`: use pods internal filesystem that is not persisted when pod crashes. `persistentVolume`, create as many persistenVolumes and persistentVolumeClaims as there are nodeHostnames.
-| cluster.nodeHostnames | list | `[]` | nodeHostnames are hostnames of VMs/Kubernetes nodes. Only used when `volumeType: persistentVolume`. Note that this require nodes to be static. |
-| secondaryReplicas | int | `0` | number of easegress service that not persists cluster data to disk. |
+| service.nodePort | int | `30080` | nodePort for Easegress Ingress Controller. |
+| service.adminPort | int | `31255` | nodePort for egctl admin access. |
```
localvar, line 41: Do we need to expose this port?
The ingress controller instances are all standalone, the admin command sent to this port will be load-balanced to different instances, which results in the pods of the service in an inconsistent state.
But I don't think there's much need for ingress controller to process an admin command.

samutamm: Good point, we don't really need admin access here. I'll remove this.

localvar (approved):

helm-charts/ingress-controller/templates/Deployment.yaml
```diff
@@ -23,10 +22,11 @@ spec:
         - -c
         - |-
           echo name: $EG_NAME > /opt/eg-config/config.yaml &&
```
zhao-kun, line 24: `EG_NAME` are samed with `{{ .Release.Name }}`. You can directly introduce the variable in the helm-charts/ingress-controller/templates/ConfigMap.yaml 
```
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}
data: |
  easegress-server.yaml: |
    name: {{.Release.Name}}
    cluster-name: easegress-ingress-controller
    cluster-role: primary
    api-addr: 0.0.0.0:2381
    data-dir: /opt/easegress/data
    log-dir: /opt/easegress/log
    debug: false
```

zhao-kun, line 24: `EG_NAME` are samed with `{{ .Release.Name }}`. You can directly introduce the variable in the helm-charts/ingress-controller/templates/ConfigMap.yaml 
```
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}
data: |
  easegress-server.yaml: |
    name: {{.Release.Name}}
    cluster-name: easegress-ingress-controller
    cluster-role: primary
    api-addr: 0.0.0.0:2381
    data-dir: /opt/easegress/data
    log-dir: /opt/easegress/log
    debug: false
```

helm-charts/ingress-controller/templates/Deployment.yaml
```diff
@@ -23,10 +22,11 @@ spec:
...
-          cat /opt/eg-config/eg-secondary.yaml >> /opt/eg-config/config.yaml &&
+          cat /opt/eg-config/easegress-server.yaml >> /opt/eg-config/config.yaml &&
```
zhao-kun, line 25: It need not to generate `config.yml`, if you write server configuration with `name` configuration.

xxx7xxxx (approved):

zhao-kun (approved):

suchen-sci (approved):

PR #449, opened 2021-12-31
Fix HTTPContext response header copy issue

suchen-sci:


localvar:
@suchen-sci, have you checked the performance impact of this change?
because `header.VisitAll` is only used here, maybe we can refactor its prototype to `VisitAll(fn func(key, values []string))`, this should have less performance penalty.

suchen-sci:
golang `ResponseWriter` interface only has method `Header()` to return `Header`. `Header` only has methods like `Add(string, string)`... `Header` don't have method like `Set(stirng, []string)` but only `Set(string, string)`... I think they want to avoid data share problem. since []string is a pointer?

localvar:
ok, I'm fine with it.

localvar (approved):

xxx7xxxx (approved):

samutamm (approved):

PR #451, opened 2021-12-31
only set host when server addr is IP (close #447)

### LINKED ISSUE #447: Easegress response with 403 when a domain name is used for server URL
**Describe the bug**
Easegress response with 403 when a domain name is used for server URL.

**To Reproduce**
1. Start Easegress with the default configuration
2. Create HTTP server as the spec below
3. Create a pipeline as the spec below, note to use a domain name for the backend server
4. Send a request with `curl HTTP://127.0.0.1:10080/`
5. Easegress response with 403 Forbidden

**Expected behavior**
Easegress returns the correct response.

**Version**
All version.

**Configuration**
* Easegress Configuration
default

* HTTP server configuration
```yaml
kind: HTTPServer
name: http-server-example
port: 10080
https: false
keepAlive: true
keepAliveTimeout: 75s
maxConnection: 10240
cacheSize: 0
rules:
  - paths:
    - pathPrefix: /
      backend: pipeline
```

* Pipeline Configuration
```yaml
name: pipeline
kind: HTTPPipeline

flow:
  - filter: proxy

filters:
  - name: proxy
    kind: Proxy
    mainPool:
      servers:
      - url: https://www.megaease.com
      loadBalance:
        policy: roundRobin
    compression:
      minLength: 1024
```

**Additional context**
Investigation shows that the request has been sent to the backend server, a possible cause is a wrong `Host` in the request.

localvar:


pkg/filter/proxy/server.go
```diff
@@ -80,10 +84,36 @@ type (
...
+	u, err := url.Parse(s.URL)
+	if err != nil {
+		return
+	}
+	host := u.Host
+
+	square := strings.LastIndexByte(host, ']')
+	colon := strings.LastIndexByte(host, ':')
+
+	// there a port number, remove it
```
xxx7xxxx, line 104:
```suggestion
	// There is a port number, remove it.
```

pkg/filter/proxy/server.go
```diff
@@ -80,10 +84,36 @@ type (
 	}
 )
 
+// String implements the Stringer interface
```
localvar, line 87:
```suggestion
// String implements the Stringer interface.
```

pkg/filter/proxy/server.go
```diff
@@ -80,10 +84,36 @@ type (
...
 func (s *Server) String() string {
 	return fmt.Sprintf("%s,%v,%d", s.URL, s.Tags, s.Weight)
 }
 
+// checkAddrPattern checks whether the server address is host name or ip:port,
+// not all error cases are handled
```
localvar, line 93:
```suggestion
// not all error cases are handled.
```

xxx7xxxx (approved):

doc/reference/filters.md
```diff
@@ -717,7 +717,7 @@ Rules to revise request header. Note that both header name and value can be a te
 
 | Name   | Type     | Description                                                                                                  | Required |
 | ------ | -------- | ------------------------------------------------------------------------------------------------------------ | -------- |
-| url    | string   | Address of the server                                                                                        | Yes      |
+| url    | string   | Address of the server, the address should start with `http://` or `https://`, followed by the hostname or IP address of the server, and then optionally followed by `:{port number}`, for example: `https://www.megaease.com`, `http://10.10.10.10:8080`. When host name is used, the `Host` of a request sent to this server is always the hostname of the server, using a [RequestAdaptor](#requestadaptor) in the pipeline to modify it will not be possible; when IP address is used, the `Host` is the same as the original request, that's can be modified by a [RequestAdaptor](#requestadaptor).        | Yes      |
```
samutamm, line 720:
```suggestion
| url    | string   | Address of the server. The address should start with `http://` or `https://`, followed by the hostname or IP address of the server, and then optionally followed by `:{port number}`, for example: `https://www.megaease.com`, `http://10.10.10.10:8080`. When host name is used, the `Host` of a request sent to this server is always the hostname of the server, and therefore using a [RequestAdaptor](#requestadaptor) in the pipeline to modify it will not be possible; when IP address is used, the `Host` is the same as the original request, that can be modified by a [RequestAdaptor](#requestadaptor).        | Yes      |
```

samutamm (approved):

suchen-sci (approved):

PR #453, opened 2022-01-05
add pipeline route for mqttproxy

suchen-sci:
This pr add `Pipeline` route for `MQTTProxy` and also make backend `Kafka` a MQTT filter. 

New version yaml: 
Now MQTTProxy can route MQTT packets to different pipelines based on the packet type. Currently support: `Connect`, `Disconnect`, `Publish`, `Subscribe`, `Unsubscribe`.
```
kind: MQTTProxy
name: mqttproxy
port: 1883
auth:
  - userName: test
    passBase64: dGVzdA==
  - userName: admin
    passBase64: YWRtaW4=
pipelines:
  - name: publish-pipeline
    packetType: Publish
  - name: connect-pipeline
    packetType: Connect
```

```
name: publish-pipeline
kind: Pipeline 
protocol: MQTT
flow:
  - filter: topicMapper
  - filter: kafka 
filters:
  - name: topicMapper
    kind: TopicMapper
    ...
  - name: kafka
    kind: Kafka
    backend: [":9093"]
```


zhao-kun:
Suggest change YAML formation to:

```yaml
rules:
  - when:
      packetType: Publish
    pipleline: publish-pipeline
```
For now, we just support the `packetType` as the only condition to dispatch the data to the pipeline. In the future, we could add extra conditions in `when`,  such as `topic start with /root/*`, etc...

substitute keyword `pipelines` to `rules` just to align the YAML to HTTPPipeline.

suchen-sci:
updated!

pkg/object/pipeline/mock.go
```diff
@@ -199,3 +193,13 @@ func MockFilterSpec(super *supervisor.Supervisor, rawSpec map[string]interface{}
 		filterSpec: filterSpec,
 	}
 }
+
+// MockGetFilter is used to get running filter from pipeline
+func MockGetFilter(p *Pipeline, name string) Filter {
```
samutamm, line 198: Could this be getter with signature `func (p *Pipeline) Filter(name string)` ?

suchen-sci: this function is used for test only... i think pipeline maybe should not provide these kind of methods to get or change state of filter except for testing purpose...

pkg/filter/authentication/auth.go
```diff
@@ -0,0 +1,155 @@
...
+
+var errAuthFail = errors.New(resultAuthFail)
+
+func init() {
+	pipeline.Register(&Authentication{})
+}
+
+type (
+	// Authentication is used to check authentication for MQTT client
+	Authentication struct {
```
samutamm, line 47: This is Authentication for MQTT only so would `MQTTClientAuth`, `MQTTAuthentication` etc be better name? `pkg/filter/validator` implements currently auth* methods for HTTP and at some point it could support MQTT protocol too.

pkg/filter/authentication/auth.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		a.authMap[auth.UserName] = sha256Sum(passwd)
+	}
+	if len(a.authMap) == 0 {
+		logger.Errorf("empty valid authentication for MQTT filter %v", filterSpec.Name())
+	}
+}
+
+// Inherit init Authentication based on previous generation
+func (k *Authentication) Inherit(filterSpec *pipeline.FilterSpec, previousGeneration pipeline.Filter) {
```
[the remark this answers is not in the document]
suchen-sci: thanks! i will update them now!

pkg/filter/authentication/auth.go
suchen-sci: ok, i will update it!

pkg/object/pipeline/mock.go
samutamm: Alright, good point! If it's used for test only, maybe consider moving it to test file to avoid confusion..

suchen-sci: it used for test not in pipeline directory. but other file directory, like `MQTTProxy` use this method to get publish backend filter and check if packets were published to backend and maybe later we add more tests in HTTPServer and use this method to check if filter is in right status...

pkg/filter/authentication/auth.go
```diff
@@ -0,0 +1,155 @@
...
+	// Authentication is used to check authentication for MQTT client
+	Authentication struct {
+		filterSpec *pipeline.FilterSpec
+		spec       *Spec
+		authMap    map[string]string
+	}
+
+	// Spec is spec for Authentication
+	Spec struct {
+		Auth []Auth `yaml:"auth" jsonschema:"required"`
```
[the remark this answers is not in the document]
suchen-sci: i think add external file is a good idea. but maybe keep this as an option? may use some hash method other than base64 to make it safer? like sha256? in that way, if they put this password into git and won't cause problem...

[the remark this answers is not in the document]
suchen-sci: updated!

zhao-kun (commented):
Overall, look good to me, except for decoupling the Kafka plugin

pkg/filter/kafka/kafka.go
```diff
@@ -0,0 +1,164 @@
...
+// HandleMQTT handle MQTT context
+func (k *Kafka) HandleMQTT(ctx context.MQTTContext) *context.MQTTResult {
+	if ctx.PacketType() != context.MQTTPublish {
+		return &context.MQTTResult{}
+	}
+	p := ctx.PublishPacket()
+	var msg *sarama.ProducerMessage
+	logger.Debugf("produce msg with topic %s", p.TopicName)
+
+	if k.mapFunc != nil {
```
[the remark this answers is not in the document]
suchen-sci: Thanks a lot for your advice!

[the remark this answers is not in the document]
suchen-sci: updated!

pkg/filter/kafka/spec.go
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+package kafka
+
+type (
+	// Spec is spec of Kafka
+	Spec struct {
+		Backend     []string     `yaml:"backend" jsonschema:"required,uniqueItems=true"`
+		TopicMapper *TopicMapper `yaml:"topicMapper" jsonschema:"omitempty"`
```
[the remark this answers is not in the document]
suchen-sci: my mistake! sorry for that!

zhao-kun:
I have some jobs that were suspended for the PR, @localvar please help review and push forward progress of the merging when you are free

pkg/context/mqttcontext.go
```diff
@@ -142,6 +137,10 @@ func NewMQTTContext(ctx stdcontext.Context, backend MQTTBackend, client MQTTClie
 		mqttCtx.packetType = MQTTOther
 	}
 	mqttCtx.packet = packet
+
+	if mqttCtx.packetType == MQTTPublish {
+		mqttCtx.publishPacket = newMQTTPublishPacket(packet.(*packets.PublishPacket))
```
localvar, line 142: please move this line into the `switch` to avoid the check.

pkg/context/mqttpublish.go
```diff
@@ -0,0 +1,93 @@
...
+func (p *mqttPublishPacket) SetTopic(topic string) {
+	p.topic = topic
+}
+
+// GetHeader get header of given key, if key not exist, return empty string
+func (p *mqttPublishPacket) GetHeader(key string) string {
+	if value, ok := p.header[key]; ok {
+		return value
+	}
+	return ""
```
localvar, line 70:
```suggestion
	return p.header[key]
```

pkg/filter/kafka/kafka.go
```diff
@@ -0,0 +1,147 @@
...
+// Inherit init Kafka based on previous generation
+func (k *Kafka) Inherit(filterSpec *pipeline.FilterSpec, previousGeneration pipeline.Filter) {
+	previousGeneration.Close()
+	k.Init(filterSpec)
+}
+
+// Close close Kafka
+func (k *Kafka) Close() {
+	close(k.done)
+	err := k.producer.Close()
```
localvar, line 115: There's a minor race condition here, please move this line to the closure function inside `Init`.

pkg/filter/mqttclientauth/mqttauth.go
```diff
@@ -0,0 +1,185 @@
...
+	// Spec is spec for MQTTClientAuth
+	// authFile format:
+	// - username: test
+	//   passBase64: dGVzdA==
+	Spec struct {
+		AuthFile string `yaml:"authFile" jsonschema:"required"`
+	}
+
+	// Auth describes username and password for MQTTProxy
+	// passSha256 make sure customer's password is safe.
```
localvar, line 65: why passSha256?

pkg/filter/topicmapper/topic.go
```diff
@@ -0,0 +1,130 @@
...
+			} else {
+				m[route.Topic] = append(m[route.Topic], r)
+			}
+		}
+	}
+	return m
+}
+
+func getPolicyMap(ps []*Policy) map[string]*Policy {
+	ans := make(map[string]*Policy)
```
localvar, line 61: what does `ans` mean?

suchen-sci: i usually use `ans`, `res` to represent result or answer of short functions, i will update this.

pkg/filter/mqttclientauth/mqttauth.go
suchen-sci: sorry, my typo...

pkg/filter/topicmapper/topic.go
localvar: Thanks, please use `res` as most developers do to improve readability.

suchen-sci: yeah, `ans` is a habit from LeetCode... i will update that

pkg/filter/kafka/kafka.go
suchen-sci: updated! thanks a lot for your review! learn a lot from it!

localvar (approved):

pkg/filter/topicmapper/topicmapper.go
```diff
@@ -0,0 +1,114 @@
...
+	publish := ctx.PublishPacket()
+	topic, headers, err := k.mapFn(publish.Topic())
+	if err != nil {
+		logger.Errorf("map topic %v failed, %v", publish.Topic(), err)
+		ctx.SetEarlyStop()
+		return &context.MQTTResult{Err: errMQTTTopicMapFailed}
+	}
+	ctx.PublishPacket().SetTopic(topic)
+	ctx.PublishPacket().SetAllHeader(headers)
+	return nil
```
samutamm, line 113: Is this `TopicMapper` filter always the last filter of the Pipeline? If not, consider calling `ctx.CallNextHandler(result)` here..

suchen-sci: emmmm, this is `Pipeline` not `HTTPPipeline`, `HTTPPipeline` use chain of responsibility to call next handler, but `Pipeline` is much simple now....

samutamm (approved):
LGTM

PR #456, opened 2022-01-07
custom data management (for multiple tenant Kafka)

localvar:


pkg/api/customdata.go
```diff
@@ -0,0 +1,151 @@
...
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+	"go.etcd.io/etcd/client/v3/concurrency"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	// CustomDataPrefix is the object prefix.
+	CustomDataPrefix = "/customdata/{kind}"
```
samutamm, line 30: Here `CustomDataPrefix = /customdata/{kind}` but in pkg/cluster/layout.go there is `customDataPrefixFormat = /custom-data/%`. Is it possible to have only one variable for this?

localvar: unfortunately, we are not able to do this, the `{kind}` in `CustomDataPrefix` is for the `chi` framework to extract parameters in the request URL, '%s` in `customDataPrefixFormat` is for `fmt.Sprintf`.

samutamm (approved):

haoel:
@localvar please add some documents for this feature.

localvar:
This feature is ready for the requirement of multiple tenant Kafka now, but the idea of this feature is the same as the CustomResource feature in EaseMesh. So I will do a refactor to avoid code duplication, but this may introduce some changes, so I plan to write the document after the refactor, is this fine?

haoel:
fine. you can include the documents in the refactor.

suchen-sci (approved):

PR #462, opened 2022-01-13
Add KVMap for MQTTContext and update Kafka filter

suchen-sci:


pkg/filter/kafka/spec.go
```diff
@@ -21,5 +21,13 @@ type (
 	// Spec is spec of Kafka
 	Spec struct {
 		Backend []string `yaml:"backend" jsonschema:"required,uniqueItems=true"`
+		KVMap   *KVMap   `yaml:"mqtt" jsonschema:"required"`
```
zhao-kun, line 24: Kafka spec could contain topic (string) and headers (map[string]string) directly. It no need to introduce a sub struct `KVMap`

pkg/filter/kafka/spec.go
```diff
@@ -21,5 +21,13 @@ type (
...
+	}
+
+	// KVMap defines ways to get kafka data from MQTTContext kv map
+	KVMap struct {
+		Topic   string `yaml:"topic" jsonschema:"required"`
+		Headers string `yaml:"headers" jsonschema:"required"`
+		Payload string `yaml:"payload" jsonschema:"required"`
```
zhao-kun, line 31: Default payload has no sense, remove it

pkg/filter/kafka/kafka.go
```diff
@@ -125,22 +141,55 @@ func (k *Kafka) Status() interface{} {
...
-	if ctx.PacketType() != context.MQTTPublish {
-		return &context.MQTTResult{}
+	var topic string
+	var headers map[string]string
+	var payload []byte
+	var ok bool
+
+	// set data from kv map
+	if k.topic != "" {
+		topic, ok = ctx.GetKV(k.topic).(string)
```
zhao-kun, line 151: The topic value in the `Context` has higher priority then default topic.

pkg/filter/kafka/kafka.go
```diff
@@ -125,22 +141,55 @@ func (k *Kafka) Status() interface{} {
...
+		if !ok {
+			return &context.MQTTResult{ErrString: resultGetDataFailed}
+		}
+	}
+	if k.headers != "" {
```
zhao-kun, line 156: The headers value in the `Context` has higher priority then default headers

pkg/filter/kafka/kafka.go
```diff
@@ -125,22 +141,55 @@ func (k *Kafka) Status() interface{} {
...
+	if k.payload != "" {
+		payload, ok = ctx.GetKV(k.payload).([]byte)
+		if !ok {
+			return &context.MQTTResult{ErrString: resultGetDataFailed}
+		}
 	}
 
-	p := ctx.PublishPacket()
-	logger.Debugf("produce msg with topic %s", p.Topic())
+	// set data from PublishPacket if data is missing
```
zhao-kun, line 169: Payload in the Context is higher priority then PublishPacket

pkg/filter/kafka/spec.go
suchen-sci: `KVMap` defines keys to get values from MQTT context KVMap. For example, if set KVMap.Topic = "kafkaTopic", then topic for kafka backend will be MQTTContext.GetKV("kafkaTopic"). maybe i use wrong name.

pkg/filter/kafka/spec.go
suchen-sci: If not use `Payload`, then we only can send data to kafka backend by using publish packet. For example, some mqtt client disconnect, and i want to send some data to kafak, then i can use MQTTContext.SetKV("payload", "some data") and use MQTTContext.GetKV("payload") to get it. `Payload` here is the key to call GetKV...

zhao-kun: Sorry for my misunderstood. The implementation is OK

pkg/filter/kafka/kafka.go
suchen-sci: updated!

localvar (approved):

PR #468, opened 2022-01-17
ci: add go ci checks, fixed some go vet issues.

kevwan:
Changes:
1. fixed some `go vet` issues
2. use `gofumpt` to format go code

Todo:
1. temporarily disabled `go vet` check, enable it after issues get fixed


samutamm (approved):
LGTM

localvar:
Thanks @kevwan for the PR. I have a question about `gofumpt`.

Most developers are writing Go code with vscode or goland (myself is using vscode), and `gofumpt` is not the default formatting tool for vscode (I'm not sure the default formatting tool of goland), this means the introduction of `gofumpt` may force most Easegress contributors to manually install it before creating a PR.

So, is there any strong justification to switch to `gofumpt`?

kevwan:
> Thanks @kevwan for the PR. I have a question about `gofumpt`.
> 
> Most developers are writing Go code with vscode or goland (myself is using vscode), and `gofumpt` is not the default formatting tool for vscode (I'm not sure the default formatting tool of goland), this means the introduction of `gofumpt` may force most Easegress contributors to manually install it before creating a PR.
> 
> So, is there any strong justification to switch to `gofumpt`?

First of all, for the rules from gofumpt, I think it's a better choice. But do you guys think so? It's the most important thing for this PR. :)

When I write code, gofump almost no errors at all, I think it's a good habit to write clean code with clean coding style.

And if the CI detects the issues, the developers will know how to change the code. I think it's good to let the tool to ask all the contributors to write code in the style that looks great.

Thanks for your reply!

localvar:
Thanks for the reply, overall, I think `gofumpt` is better, but my concerns about this change are:
* the default tool is good enough.
* not all contributors know how to check the CI errors, especially for our external contributors.
* change the default tool requires everyone to install the new tool manually.
* not sure if it is configurable, but personally, I prefer using `func foo(a string, b string)` instead of `func foo(a, b string)` because all parameters of the former are in the same style which is more friendly for things like grep.

However, the above is just my personal opinion and I'd like to collect more comments from others.

samutamm:
So is the question which one is better; the current standard go linter [`gofmt`](https://pkg.go.dev/cmd/gofmt) or stricter version of it [`gofumpt`](https://pkg.go.dev/mvdan.cc/gofumpt)? Personally I am fine with both. I do agree with @localvar that the advantage with the standard `gofmt` is that you don't have to install it. Also it's more widely adopted and mature.

kevwan:
Personally, I prefer the stricter one. :)

My thinking is that we need to decide which coding style is preferred. For the tool concern, CI will let contributors know how to do if they don't meet the requirement.

But, I agree with you guys, it's just a choice. Let's pick one.

Cheers.

suchen-sci:
I am fine with both. But anyone knows why `test-win` failed? the error is strange...

mvdan:
> not sure if it is configurable, but personally, I prefer using `func foo(a string, b string)` instead of `func foo(a, b string)` because all parameters of the former are in the same style which is more friendly for things like grep.

FYI, that rule is opt-in via `-extra`.

pkg/supervisor/mock.go
```diff
@@ -18,26 +18,21 @@
...
-		firstHandle:         firstHandle,
-		firstHandleDone:     firstHandleDone,
-		done:                done,
+		options:         options,
+		cls:             cls,
+		objectRegistry:  objectRegistry,
+		watcher:         watcher,
+		firstHandle:     firstHandle,
+		firstHandleDone: firstHandleDone,
+		done:            done,
```
haoel, line 35: I am not sure it is good or not to remove `businessController` and `systemControllers`?

kevwan: Actually, there is a problem that sync.Map contains sync.Mutex, it's pass-by-value.

haoel:
After walking through the code change one by one, I have the following comments,  I think some of them are good, and a few are debatable.

- Remove some of the empty lines of code    🆗
- Add the empty line between the functions   🆗
- Add the beginning space for comments  🆗
- Merge the same type of declaration  ❓ **[not sure]**
- Expand the grouped `var ( )`  declaration which only has one variable   🆗
- Change `var name = value` to `name := value`  🆗
- A new line for right brace '}'  👍
- removed the member variables without code reference.  ❓ **[not sure]**

Actually, I think the above changes have a question mark that is not **strict**, it's a bit **obsessive**.


And one more question - why does this change break the test on Windows?

pkg/supervisor/mock.go
haoel: `passed-by-value` or `passed-by-reference` is a choice in a different scenario.  As this is a New Object function. pass-by-value might make sense.  @localvar please confirm this is a problem to not.

But what I am concerned about, these changes totally modified the function signature, and the reason is we need to align with a lint tool instead of trying to optimize the function interface and make it more usable or readable.

kevwan:
> After walking through the code change one by one, I have the following comments, I think some of them are good, and a few are debatable.
> 
> * Remove some of the empty lines of code    🆗
> * Add the empty line between the functions   🆗
> * Add the beginning space for comments  🆗
> * Merge the same type of declaration  ❓ **[not sure]**
I agree that it's just a coding style choice. I'll remove `-extra` to remove this kind of change.
> * Expand the grouped `var ( )`  declaration which only has one variable   🆗
> * Change `var name = value` to `name := value`  🆗
> * A new line for right brace '}'  👍
> * removed the member variables without code reference.  ❓ **[not sure]**
I'm thinking that if it's not used, why to keep it?
> 
> Actually, I think the above changes have a question mark that is not **strict**, it's a bit **obsessive**.
> 
> And one more question - why does this change break the test on Windows?

pkg/supervisor/mock.go
kevwan: > But what I am concerned about, these changes totally modified the function signature, and the reason is we need to align with a lint tool instead of trying to optimize the function interface and make it more usable or readable.
It's configurable.

localvar: `sync.Map` contains `sync.Mutex` and `atomic.Value`, both "must not be copied after first use", it is not a problem in this case because they are not used yet.

But I think we should avoid this `passed-by-value` because it is hard to track whether they are used or not in all cases, which makes the code error-prone.

haoel: I think this is a problem even they are unused. at least, it is a potential problem, which must be fixed.

The second question is - whether are those two fields are useless? if it is, we should remove it, otherwise, we need to fix it.

localvar: They have usages in production code. Because function `NewMock` is for testing, we can remove the assignment from `NewMock` by now (and add them back if they are required later, in a manner other than `passed-by-value`), as what @kevwan do in this PR.

kevwan:
Because `Merge the same type of declaration` is not sure, I removed `-extra` to ignore it.

pkg/supervisor/mock.go
kevwan: I ignored it by new commits. Because there are a few more issues like this. I'll might work on all of them later.

Let's just keep this PR simple enough.

.github/workflows/test.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         run: |
           go mod verify
           go mod download
-          go test -v ./...
+          go test -race -coverprofile=coverage.txt -covermode=atomic ./...
```
samutamm, line 66:
```suggestion
          go test -v -race ./...
```
Windows does not seem to like filenames without quotes `"`. I think we don't need to compute the code coverage numbers on multiple platforms; the coverage should be the same. But adding `-race` flag is probably a good change here as the behavior of the go routines might differ between platforms..

localvar:
Besides the test failure on Windows, I think another issue that blocks the merge of the PR is: can we improve the installation experience of the new tool? The most ideal solution is we commit some vscode & goland settings to the repo, and these settings will make the editors install `gofumpt` automatically and set it as the default tool.

@mvdan , @kevwan , do you have any suggestions on this?

kevwan:
> Besides the test failure on Windows, I think another issue that blocks the merge of the PR is: can we improve the installation experience of the new tool? The most ideal solution is we commit some vscode & goland settings to the repo, and these settings will make the editors install `gofumpt` automatically and set it as the default tool.
> 
> @mvdan , @kevwan , do you have any suggestions on this?

vscode: https://github.com/mvdan/gofumpt#visual-studio-code
goland: https://github.com/mvdan/gofumpt#goland

PR #472, opened 2022-01-24
Basic Auth etcd mode to use username by default

samutamm:
When configuring Basic Auth using Easegress `custom-data` store, the `key` value is used as unique identifier of the data entry and username. Sometimes two `custom-data` entries might share the same user identity: for example we could have `bob-pearls` and `bob-bananas` data entries but only one identity `bob`.

After this PR, the `custom-data` entries can share same identity by adding `username` in `custom-data`:

```yaml
rebuild: true
list:
 - key: bob-bananas
   username: bob
   password: ...

 - key: bob-pearls
   username: bob
   password: ...
```
Now `key` field is still unique identifier of the entries but username can be shared.

### DIFF
--- a/pkg/filter/validator/basicauth.go
+++ b/pkg/filter/validator/basicauth.go
@@ -94,9 +94,9 @@ type (
 
 	// etcdCredentials defines the format for credentials in etcd
 	etcdCredentials struct {
-		Key      string `yaml:"key" jsonschema:"omitempty"`
-		Username string `yaml:"username" jsonschema:"omitempty"`
-		Password string `yaml:"password" jsonschema:"required"`
+		Key  string `yaml:"key" jsonschema:"omitempty"`
+		User string `yaml:"username" jsonschema:"omitempty"`
+		Pass string `yaml:"password" jsonschema:"required"`
 	}
 )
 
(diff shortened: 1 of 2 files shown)


zouyingjie (approved):

pkg/filter/validator/basicauth.go
```diff
@@ -89,16 +92,26 @@ type (
...
+		Password string `yaml:"password" jsonschema:"required"`
 	}
 )
 
 const (
 	customDataPrefix = "/custom-data/"
 )
 
+// DefaultUsername uses Username if present, otherwise Key
+func (cred *etcdCredentials) DefaultUsername() string {
```
localvar, line 108: propose to rename this function, `DefaultUsername` makes me think there's a default name.

suchen-sci (approved):

pkg/filter/validator/basicauth.go
samutamm: Good point, it's quite confusing name.. https://github.com/megaease/easegress/pull/472/commits/69ad26083b3492c55b9e669ab5738cc43c8831ab renames it to `Username()`

localvar (approved):

Issue #467, opened 2022-01-17
Support mTLS client certificate authenticating


samutamm:
**Is your feature request related to a problem? Please describe.**
In client-server communication, where client is browser or other user terminal, using HTTPS/TLS helps client to trust the server, as server certificate is authenticated by 3rd party. In some scenarios (IOT, microservices, etc) it might be important for server to ensure that client is not compromised. Authenticating client certificates would help server to trust the client.

**Describe the solution you'd like**
Create new Mutual TLS (mTLS) feature. When enabled, Easegress verifies that TLS certificates of the connecting client are signed by same CA certificates as Easegress certificates. Client will equally verify certificates of Easegress.

**Describe alternatives you've considered**
There's no alternative way for Easegress server to trust the client.

**Additional context**

- https://en.wikipedia.org/wiki/Mutual_authentication
- https://en.wikipedia.org/wiki/Certificate_authority
- https://github.com/haoel/mTLS

samutamm:
It turned out **mTLS** is already supported (https://github.com/megaease/easegress/pull/281). 

HTTPServer enable mTLS by setting `https` to true and providing CA certificate in `caCertBase64`;

```yaml
kind: HTTPServer
...
https: true
caCertBase64: "CA_CERT"
certs:
  sub.domain.io: "SERVER_CERT_CRT"
keys:
  sub.domain.io: "SERVER_CERT_KEY"
```

However there's still room for improvement. Currently updating certificates requires deleting and re-creating HTTPServer, causing service downtime. [`AutoCertManager`](https://github.com/megaease/easegress/blob/main/doc/reference/controllers.md#autocertmanager) helps with automatic certificate renewal, but for `mTLS` there are still use cases when certificates should be updated manually. For example, if certificates of one client are revoked and new certificates are created for the rest of the clients and server, there should be a way to update those new certificates to Easegress without restarting HTTPServer.

### Design draft - HTTPServer certificateStorage

Here's an alternative proposal that synchronizes certifications from Easegress [custom-data store](https://github.com/megaease/easegress/pull/456).

```yaml
kind: HTTPServer
...
https: true
# following entry is new
certificateStorage:
  prefix: "certificates/"
# caCertBase64, certs and keys are left empty
```

Certificates can be stored to Easegress using `egctl`:

```bash
echo '                                       
rebuild: true
list:
  - caCertBase64: 'CA_CERT'
     certCrtBase64: 'SERVER_CERT_CRT'
[...]

PR #474, opened 2022-01-24
CertExtractor filter

samutamm:
Resolve https://github.com/megaease/easegress/issues/467

CertExtractor filter adds values from certificates Subject or Issuer to headers.

Here's an example config:

```yaml 
  - kind: "CertExtractor"
    name: "cn-extractor"
    certIndex: -1
    target: "subject"
    field: "CommonName"
    headerKey: "X-AUTH-USER"
```
that extracts the "CommonName" field's value from the subject of the last certificate. The value is set to "X-AUTH-USER" request header.

Here's an example that sets `CommonName` value for new certificate: https://github.com/haoel/mTLS/blob/main/certs.go#L141

### References:
- https://pkg.go.dev/crypto/x509#Certificate
- https://pkg.go.dev/crypto/x509/pkix#Name


doc/reference/filters.md
```diff
@@ -723,8 +725,7 @@ headerMap:
 
 | Name         | Type     | Description                      | Required |
 | ------------ | -------- | -------------------------------- | -------- |
-| headerMap | [][HeaderToJSON.HeaderMap](#headertojsonheadermap) | headerMap defines a map between HTTP header name and corresponding JSON filed name
-  | Yes      |
+| headerMap | [][HeaderToJSON.HeaderMap](#headertojsonheadermap) | headerMap defines a map between HTTP header name and corresponding JSON filed name | Yes      |
```
localvar, line 728:
```suggestion
| headerMap | [][HeaderToJSON.HeaderMap](#headertojsonheadermap) | headerMap defines a map between HTTP header name and corresponding JSON field name | Yes      |
```
```suggestion
| headerMap | [][HeaderToJSON.HeaderMap](#headertojsonheadermap) | headerMap defines a map between HTTP header name and corresponding JSON filed name | Yes      |
```

pkg/context/httpcontext.go
```diff
@@ -105,6 +106,8 @@ type (
 		Std() *http.Request
 
 		Size() uint64 // bytes
+
+		TLS() *tls.ConnectionState
```
localvar, line 110: propose not adding this function to keep the interface as minimal as possible, as the information can be retrieved from `Std().TLS`.
I and @suchen-sci are working on the design of the new pipeline, though not finished yet, we believe a more generic `Context` interface requires fewer code changes to implement the new design.

pkg/filter/certextractor/certextractor.go
```diff
@@ -0,0 +1,140 @@
...
+	result := ce.handle(ctx)
+	return ctx.CallNextHandler(result)
+}
+
+// CertExtractor extracts given field from TLS certificates and sets it to request headers.
+func (ce *CertExtractor) handle(ctx httpcontext.HTTPContext) string {
+	r := ctx.Request()
+	if connectionState := r.TLS(); connectionState != nil {
+		certs := connectionState.PeerCertificates
+		if certs != nil && len(certs) > 0 {
```
localvar, line 95: please refactor to return early to reduce nesting.

pkg/filter/certextractor/spec.go
```diff
@@ -0,0 +1,46 @@
...
+type (
+	// Spec describes the CertExtractor.
+	Spec struct {
+		CertIndex int16  `yaml:"certIndex" jsonschema:"required"`
+		Target    string `yaml:"target" "omitempty,enum=subject,enum=issuer"`
+		// Different field options listed here https://pkg.go.dev/crypto/x509/pkix#Name
+		Field string `yaml:"field" "omitempty,enum=Country,enum=Organization,enum=OrganizationalUnit,
+													enum=Locality,enum=Province,enum=StreetAddress,enum=PostalCode
+													enum=SerialNumber,enum=CommonName"`
+		HeaderKey string `yaml:"headerKey" "omitempty"`
```
localvar, line 33:
```suggestion
		Target    string `yaml:"target" jsonschema:"required,enum=subject,enum=issuer"`
		// Different field options listed here https://pkg.go.dev/crypto/x509/pkix#Name
		Field string `yaml:"field" jsonschema:"required,enum=Country,enum=Organization,enum=OrganizationalUnit,
		HeaderKey string `yaml:"headerKey" jsonschema:"required"`
```

samutamm: It's not very visible here but the removed lines contains enum values and they shouldn't be removed.

localvar: updated.

pkg/context/httpcontext.go
samutamm: https://github.com/megaease/easegress/pull/474/commits/b3d6f608aa880d0fc63a0f3032560b69c7bd6df4 fixes this

pkg/filter/certextractor/spec.go
```diff
@@ -0,0 +1,44 @@
...
+		Field     string `yaml:"field" jsonschema:"required,enum=Country,enum=Organization,enum=OrganizationalUnit,enum=Locality,enum=Province,enum=StreetAddress,enum=PostalCode,enum=SerialNumber,enum=CommonName"`
+		HeaderKey string `yaml:"headerKey" jsonschema:"required"`
+	}
+)
+
+// Validate is dummy as yaml rules already validate Spec.
+func (spec *Spec) Validate() error { return nil }
+
+// DefaultHeaderKey returns HeaderKey or the default value.
+func (spec *Spec) DefaultHeaderKey() string {
```
localvar, line 39: propose to rename this function as it does not always return the default value.
I think one solution is to remove this function because the best name `HeaderKey` is already used in `Spec`.

samutamm: https://github.com/megaease/easegress/pull/474/commits/3148b675d5fa40bccbe1dfcb44cb3dc90903de07 fixes this

localvar (approved):

zouyingjie (approved):

Issue #483, opened 2022-01-27
CORSAdaptor should support CORS requests with "Origin" header


jthann:
The current builtin filter CORSAdaptor implementation just handle the preflight OPTIONS request in CORS not-so-simple request and do nothing for other CORS request.  
The CORS divide request into two category: simple request and not-so-simple request.
For not-so-simple request，after OPTIONS preflight request passed，later request also should be handled through http Origin request Header and add response header Access-Control-Allow-Origin.
For simple request it doest't have preflight request and should also be handled through http Origin request Header and add response header Access-Control-Allow-Origin

So the right way do this as following pseudo-code show:
```go
func handle(req http.Request, resp http.Response) {
         //is it a CORS request
       isCorsRequest:= req.getHeader("Origin")!=""
       if !isCorsRequest {             
            callNextFilter()  //Execute next filter in pipeline and return
            return
       }
       isPreflight:= req.Method()=="OPTIONS" && req.getHeader("Access-Control-Request-Method")!=""
       //is it a CORS preflight request
       if isPreflight {
            // valid CORS config and add all CORS response headers
            corsHandle(req, resp, isPreflight)
            return //here should break the whole request pipeline and just respond to client
       }
       if isCorsRequest {
            //valid CORS config and just add Access-Control-Allow-Origin response header
            corsHandle(req, resp)
            callNextFilter()  /here should execute next filter in pipeline
            return
       }
}
```
By the way we should add MaxAge config in CORSAdaptor spec

We can also see java spring framework source code  **org.springframework.web.cors.DefaultCorsProcessor** and **org.springframework.web.cors.CorsUtils** as references in artifact 'org.springframework:spring-web:5.3.15'

W3C reference: [https://www.w3.org/TR/2020/SPSD-cors-20200602/](https://www.w3.org/TR/2020/SPSD-cors-20200602/)
One Chinese blog that clarify CORS standard: [https://www.ruanyifeng.com/blog/2016/04/cors.html](https://www.ruanyifeng.com/blog/2016/04/cors.html)

samutamm:
Thanks for opening the issue @jthann. The current implementation matches the documentation of [CORSAdaptor](https://github.com/megaease/easegress/blob/main/doc/reference/filters.md#corsadaptor). It aims to handle the _CORS-preflight request_ (the one with OPTIONS method, a.k.a _not-so-simple_ request) and nothing more.
[...]

PR #484, opened 2022-01-27
fix builtin filter CORSAdaptor bug

jthann:
The bug issue describe here: [https://github.com/megaease/easegress/issues/483](https://github.com/megaease/easegress/issues/483)


samutamm:
Changes included in https://github.com/megaease/easegress/pull/498

PR #497, opened 2022-02-07
support decompose request body for RequestAdaptor filter

suchen-sci:
support decompose gzip request body by using RequestAdaptor filter.

### DIFF
--- a/pkg/filter/headertojson/headertojson.go
+++ b/pkg/filter/headertojson/headertojson.go
@@ -19,6 +19,7 @@ package headertojson
 
 import (
 	"bytes"
+	"compress/gzip"
 	"errors"
 	"io"
 	"net/http"
@@ -172,6 +173,24 @@ func getNewBody(reqBody []byte, headerMap map[string]interface{}) (interface{}, 
 	return nil, errJSONEncodeDecode
 }
 
+func getReqBody(ctx context.HTTPContext) ([]byte, error) {
+	encoding := ctx.Request().Header().Get("Content-Encoding")
+	if encoding == "" {
+		return io.ReadAll(ctx.Request().Body())
+
+	} else if encoding == "gzip" {
+		reader, err := gzip.NewReader(ctx.Request().Body())
+		if err != nil {
+			return nil, err
+		}
+		defer reader.Close()
+		return io.ReadAll(reader)
+
+	} else {
+		return nil, errors.New("unknown encoding method")
+	}
+}
+
 func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
 	headerMap := make(map[string]interface{})
 	for header, json := range h.headerMap {
@@ -184,7 +203,7 @@ func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
 		return ""
 	}
 
-	reqBody, err := io.ReadAll(ctx.Request().Body())
+	reqBody, err := getReqBody(ctx)
 	if err != nil {
 		return resultBodyReadErr
 	}
@@ -204,5 +223,6 @@ func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
 		return resultJSONEncodeDecodeErr
 	}
 	ctx.Request().SetBody(bytes.NewReader(bodyBytes))
+	ctx.Request().Header().Del("Content-Encoding")
 	return ""
 }
--- a/pkg/filter/headertojson/headertojson.go
+++ b/pkg/filter/headertojson/headertojson.go
@@ -186,9 +186,8 @@ func getReqBody(ctx context.HTTPContext) ([]byte, error) {
 		defer reader.Close()
 		return io.ReadAll(reader)
 
-	} else {
-		return nil, errors.New("unknown encoding method")
 	}
+	return nil, errors.New("unknown encoding method")
 }
 
 func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
--- a/pkg/filter/headertojson/headertojson.go
+++ b/pkg/filter/headertojson/headertojson.go
@@ -19,7 +19,6 @@ package headertojson
 
 import (
 	"bytes"
-	"compress/gzip"
 	"errors"
 	"io"
 	"net/http"
@@ -173,23 +172,6 @@ func getNewBody(reqBody []byte, headerMap map[string]interface{}) (interface{}, 
 	return nil, errJSONEncodeDecode
 }
 
-func getReqBody(ctx context.HTTPContext) ([]byte, error) {
-	encoding := ctx.Request().Header().Get("Content-Encoding")
-	if encoding == "" {
-		return io.ReadAll(ctx.Request().Body())
-
-	} else if encoding == "gzip" {
-		reader, err := gzip.NewReader(ctx.Request().Body())
-		if err != nil {
-			return nil, err
-		}
-		defer reader.Close()
-		return io.ReadAll(reader)
-
-	}
-	return nil, errors.New("unknown encoding method")
-}
-
 func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
 	headerMap := make(map[string]interface{})
 	for header, json := range h.headerMap {
@@ -202,7 +184,7 @@ func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
 		return ""
 	}
 
-	reqBody, err := getReqBody(ctx)
+	reqBody, err := io.ReadAll(ctx.Request().Body())
 	if err != nil {
 		return resultBodyReadErr
 	}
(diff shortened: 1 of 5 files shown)


samutamm (approved):

pkg/filter/headertojson/headertojson.go
```diff
@@ -204,5 +223,6 @@ func (h *HeaderToJSON) handle(ctx context.HTTPContext) string {
 		return resultJSONEncodeDecodeErr
 	}
 	ctx.Request().SetBody(bytes.NewReader(bodyBytes))
+	ctx.Request().Header().Del("Content-Encoding")
```
localvar, line 226: this should be fine.
but please note keep gzip enabled saves bandwidth, so another option is the compress the body again if gzip is on.

suchen-sci: i think maybe we can make another filter to encoding body, because if we encode body in headertojson filter, then we need decode it again for filters like Kafka...

landyking: In my opinion, this should be a built-in feature for HTTPServer. Decompression should be handled automatically. Compression should require parameters to be specified.

suchen-sci: how about make compress and decompress as a new filter. if we decompress it automatically and then compress it again, it is a little bit waste...

localvar: yes, adding a new filter is better, or we can enhance the existing `RequestAdaptor` to support compress/decompress.
the HTTPServer is to receive requests, it should not care about the content of the request, and most pipelines only process request headers, in which case decompressing the body of all requests is useless.

suchen-sci: sounds great, i will check `RequestAdaptor` and update it now!

suchen-sci: updated!

pkg/filter/requestadaptor/requestadaptor_test.go
```diff
@@ -0,0 +1,190 @@
...
+	assert.NotNil(ra.DefaultSpec())
+	assert.NotEmpty(ra.Results())
+	assert.Nil(ra.Status())
+
+	newRA := &RequestAdaptor{}
+	newRA.Inherit(filterSpec, ra)
+	newRA.Close()
+}
+
+func TestDecompose(t *testing.T) {
```
localvar, line 96:
```suggestion
func TestDecompress(t *testing.T) {
```

pkg/filter/requestadaptor/requestadaptor.go
```diff
@@ -50,11 +54,12 @@ type (
...
-		Method string                `yaml:"method" jsonschema:"omitempty,format=httpmethod"`
-		Path   *pathadaptor.Spec     `yaml:"path,omitempty" jsonschema:"omitempty"`
-		Header *httpheader.AdaptSpec `yaml:"header,omitempty" jsonschema:"omitempty"`
-		Body   string                `yaml:"body" jsonschema:"omitempty"`
+		Host       string                `yaml:"host" jsonschema:"omitempty"`
+		Method     string                `yaml:"method" jsonschema:"omitempty,format=httpmethod"`
+		Path       *pathadaptor.Spec     `yaml:"path,omitempty" jsonschema:"omitempty"`
+		Header     *httpheader.AdaptSpec `yaml:"header,omitempty" jsonschema:"omitempty"`
+		Body       string                `yaml:"body" jsonschema:"omitempty"`
+		Decompress string                `yaml:"decompress" jsonschema:"omitempty"`
```
localvar, line 62: we should also support compress.

pkg/filter/requestadaptor/requestadaptor.go
```diff
@@ -151,6 +159,31 @@ func (ra *RequestAdaptor) handle(ctx context.HTTPContext) string {
 			ctx.Request().SetHost(ra.spec.Host)
 		}
 	}
+
+	if ra.spec.Decompress != "" {
+		res := ra.processDecompress(ctx)
+		if res != "" {
+			return res
+		}
+	}
```
localvar, line 168: there are 3 cases now:
* when `ra.spec.Body` is set, no need to do decompress
* when `ra.spec.Body` is not set, and decompress is set, decompress the body
* when compress (need to add) is set, compress the body
note that decompress and compress cannot be set at the same time.

localvar (changes requested):

pkg/filter/requestadaptor/requestadaptor.go
suchen-sci: i agree with last two cases, but for first case, it may not happen... if set body, the header of "Content-Encoding" is empty, then it will not decompress anyway. if set the body but original body is compressed, then user should also set `Header.Del` to delete "Content-Encoding" because the new body is not compressed. if set body, do we need to delete header of "Content-Encoding" automatically?

localvar: `if set body, the header of "Content-Encoding" is empty`, this is not true when the original request is gzipped.
`if set body, do we need to delete header of "Content-Encoding" automatically?`: yes.

suchen-sci: updated!

pkg/filter/requestadaptor/requestadaptor.go
```diff
@@ -151,6 +168,62 @@ func (ra *RequestAdaptor) handle(ctx context.HTTPContext) string {
...
+		reader, err := gzip.NewReader(ctx.Request().Body())
+		if err != nil {
+			return resultDecompressFail
+		}
+		defer reader.Close()
+		data, err := io.ReadAll(reader)
+		if err != nil {
+			return resultDecompressFail
+		}
+		ctx.Request().SetBody(bytes.NewReader(data))
```
localvar, line 224:
```suggestion
		ctx.Request().SetBody(reader)
```

pkg/filter/requestadaptor/requestadaptor.go
```diff
@@ -81,6 +88,15 @@ func (ra *RequestAdaptor) Results() []string {
...
 	ra.filterSpec, ra.spec = filterSpec, filterSpec.FilterSpec().(*Spec)
+	if ra.spec.Decompress != "" && ra.spec.Decompress != "gzip" {
+		panic("RequestAdaptor only support decompress type of gzip")
+	}
+	if ra.spec.Compress != "" && ra.spec.Compress != "gzip" {
+		panic("RequestAdaptor only support decompress type of gzip")
+	}
+	if ra.spec.Compress != "" && ra.spec.Decompress != "" {
+		panic("RequestAdaptor can only do compress or decompress for given request body, not both")
+	}
```
localvar, line 99:
```suggestion
	if ra.spec.Compress != "" && ra.spec.Decompress != "" {
		panic("RequestAdaptor can only do compress or decompress for given request body, not both")
	}
	if ra.spec.Body != "" && ra.spec.Decompress != "" {
		panic("No need to decompress when body is specified in RequestAdaptor spec")	
	}
```

pkg/filter/requestadaptor/requestadaptor.go
```diff
@@ -151,6 +168,62 @@ func (ra *RequestAdaptor) handle(ctx context.HTTPContext) string {
...
+	data, err := io.ReadAll(ctx.Request().Body())
+	if err != nil {
+		logger.Errorf("read request body failed, %v", err)
+		return resultCompressFail
+	}
+	_, err = gw.Write(data)
+	if err != nil {
+		logger.Errorf("compress request body failed, %v", err)
+		return resultCompressFail
+	}
```
localvar, line 206:
```suggestion
	_, err := io.Copy(gw, ctx.Request().Body())
	if err != nil {
		logger.Errorf("compress request body failed, %v", err)
		return resultCompressFail
	}
	gw.Close()
```

pkg/filter/requestadaptor/requestadaptor.go
suchen-sci: "gzip" reader is a `ReadCloser` but our `ctx.Request().Body` is `Reader`. if set `ReadCloser` directly to `Reader`, Close method will not be called and resource will not be released.....

pkg/filter/requestadaptor/requestadaptor.go
suchen-sci: i didn't panic it because set `Body` and `Decompress` same time will not cause problem, but in this case `Decompress` will not run in this case. But your change is better to reduce confusion.

pkg/filter/requestadaptor/requestadaptor.go
suchen-sci: updated!

localvar (approved):

PR #500, opened 2022-02-08
Custom data

localvar:
refactor custom data, now Easegress & EaseMesh uses the same implementation

### DIFF
--- a/doc/reference/customdata.md
+++ b/doc/reference/customdata.md
@@ -1,0 +1,105 @@
+# Custom Data Management
+
+The `Custom Data` feature implements a storage for 'any' data, which can be used by other components for data persistence.
+
+Because the schema of the data being persisted varies from components, a `CustomDataKind` must be defined to distinguish the data.
+
+## CustomDataKind
+
+The YAML example below defines a CustomDataKind:
+
+```yaml
+name: kind1
+idField: name
+jsonSchema:
+  type: object
+  properties:
+    name:
+	  type: string
+	  required: true
+```
+
+The `name` field is required, it is the kind name of custom data items of this kind.
+The `idField` is optional, and its default value is `name`, the value of this field of a data item is used as its identifier.
+The `jsonSchema` is optional, if provided, data items of this kind will be validated against this [JSON Schema](http://json-schema.org/).
+
+## CustomData
+
+CustomData is a map, the keys of this map must be strings while the values can be any valid JSON values, but when value is also a map, its keys must be strings too.
+
+A CustomData item must contain the `idField` defined by its corresponding CustomDataKind, for example, the data items of the kind defined in the above example must contain the `name` field as their identifiers.
+
+Below is an example of a CustomData:
+
+```yaml
+name: data1
+field1: 12
+field2: abc
+field3: [1, 2, 3, 4]
+```
+
+## API
+
+* **Create a CustomDataKind**
+	* **URL**: http://{ip}:{port}/apis/v1/customdatakinds
+	* **Method**: POST
+	* **Body**: CustomDataKind definition is YAML.
+
+* **Update a CustomDataKind**
+	* **URL**: http://{ip}:{port}/apis/v1/customdatakinds
+	* **Method**: PUT
+	* **Body**: CustomDataKind definition is YAML.
+
+* **Query the definition of a CustomDataKind**
+	* **URL**: http://{ip}:{port}/apis/v1/customdatakinds/{kind name}
+	* **Method**: GET
+  
+* **List the definition of all CustomDataKind**
+	* **URL**: http://{ip}:{port}/apis/v1/customdatakinds
+	* **Method**: GET
+
+* **Delete a CustomDataKind**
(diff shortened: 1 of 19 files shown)


doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+The YAML example below defines a CustomDataKind:
+
+```yaml
+name: kind1
+idField: name
+jsonSchema:
+  type: object
+  properties:
+    name:
+	  type: string
```
samutamm, line 18:
```suggestion
      type: string
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+	  required: true
```
samutamm, line 19:
```suggestion
      required: true
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+The `name` field is required, it is the kind name of custom data items of this kind.
+The `idField` is optional, and its default value is `name`, the value of this field of a data item is used as its identifier.
+The `jsonSchema` is optional, if provided, data items of this kind will be validated against this [JSON Schema](http://json-schema.org/).
+
+## CustomData
+
+CustomData is a map, the keys of this map must be strings while the values can be any valid JSON values, but when value is also a map, its keys must be strings too.
+
```
samutamm, line 29: Probably very rare edge case but what about map containing slice/array of maps? In
`map[string][[]map[string]sometype]` I guess `sometype` can also only be `string` but not `slice` or `map`?

Is it correct to say:
```suggestion
In other words, nested CustomData objects can have at maximum the depth of two levels.
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+* **Delete all CustomData of a kind**
+	* **URL**: http://{ip}:{port}/apis/v1/customdata/{kind name}
+	* **Method**: DELETE
+
+* **Bulk update**
+	* **URL**: http://{ip}:{port}/apis/v1/customdata/{kind name}/items
+	* **Method**: POST
+	* **Body**: A change request in YAML, as defined below.
+
+		```yaml
```
samutamm, line 96:
```suggestion
```yaml
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+		rebuild: false
```
samutamm, line 97:
```suggestion
rebuild: false
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+		delete: [data1, data2]
```
samutamm, line 98:
```suggestion
delete: [data1, data2]
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+		list:
```
samutamm, line 99:
```suggestion
list:
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+		- name: data3
```
samutamm, line 100:
```suggestion
- name: data3
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+			field1: 12
```
samutamm, line 101:
```suggestion
  field1: 12
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+		- name: data4
```
samutamm, line 102:
```suggestion
- name: data4
```

doc/reference/customdata.md
```diff
@@ -0,0 +1,105 @@
...
+			field1: foo
```
samutamm, line 103:
```suggestion
  field1: foo
```

pkg/cluster/clustertest/cluster.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	return 0
+}
+
+// Del implements STM.Del
+func (stm *MockedSTM) Del(key string) {
+	if stm.MockedRev != nil {
+		stm.MockedDel(key)
+	}
+}
```
samutamm, line 264: Just noting here that after this PR this `MockedCluster` could be used in `pkg/filter/headerlookup/headerlookup_test.go` and some other Filter tests too to make the tests run faster.

doc/reference/customdata.md
localvar: No, there's no limitation for the depth.
I mean if there are nested maps (either as a value or as an array item), the keys of the map must be strings.

samutamm: Ah Ok I get it now.

samutamm (approved):

suchen-sci (approved):

pkg/cluster/clustertest/cluster.go
samutamm: https://github.com/megaease/easegress/pull/590

PR #501, opened 2022-02-10
update dependencies

localvar:
Google DNS provider used in AutoCertManager is temporarily disabled due to dependency conflict.


samutamm (approved):
Which packages are conflicting? Can we expect that in the future some of them are updated and conflict will be resolved?

suchen-sci (approved):

localvar:
> Which packages are conflicting? Can we expect that in the future some of them are updated and conflict will be resolved?

it is the google cloud API package, I think it can be fixed after `github.com/libdns/googleclouddns` has been updated to use the latest google cloud API.

PR #509, opened 2022-02-14
sort ingress path

localvar:
make sure longer prefix is always matched before shorter prefix.

### DIFF
--- a/pkg/object/ingresscontroller/translator.go
+++ b/pkg/object/ingresscontroller/translator.go
@@ -20,6 +20,7 @@ package ingresscontroller
 import (
 	"fmt"
 	"net"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/object/ingresscontroller/translator.go
```diff
@@ -379,6 +397,25 @@ func (st *specTranslator) translate() error {
...
 
+	// sort path:
+	// * precise path first
+	// * longer prefix first
+	for _, r := range b.Rules {
+		sort.Slice(r.Paths, func(i, j int) bool {
+			p1, p2 := r.Paths[i], r.Paths[j]
+			switch {
+			case p1.Path == "" && p2.Path == "":
+				return p1.Path < p2.Path
```
samutamm, line 408: Isn't this always false, as `p1.Path` and `p2.Path` are both empty strings `""`, so p2 cannot be larger?

samutamm: I would consider moving this sorting function to `util` package and adding testsuite for it.

localvar: my mistake, it should be `!=`.

samutamm (approved):

suchen-sci (approved):

Issue #505, opened 2022-02-11
Unexpected MQTT client auth filter behaviors


zhao-kun:
**Describe the bug**
The current spec [1] of the MQTT client auth includes a path of the file that contains the credentials of clients. The path is related to the environment, If we'd like to use a file to store credentials, first we need to clarify the path is related to serverside or clientside. Second, in most cases, the file path in the spec should refer to clientside instead of the serverside as servers need high privilege, and the operators also seldom maintain server directly.

**To Reproduce**
1. Write an mqtt client auth spec
```yaml
name: pipeline-mqtt-auth
kind: Pipeline
protocol: MQTT
flow:
- filter: auth
filters:
- name: auth
  kind: MQTTClientAuth
  authFile: /opt/easegress-mqtt/mqtt-spec/passwd.yml
```
2. Send a request to `easegress` server which run on another machine
```bash
$ egclt object create -f mqtt-client-auth.yml --server XXX.XXX.XXX.XXX
```

3. See log of the easegress server, there are errors:

```
2022-02-10T08:38:18.826Z        ERROR   supervisor/object.go:384        pipeline-mqtt-auth: recover from Init, err: invalid file name /opt/easegress-mqtt/mqtt-spec/passwd.yml, stack trace:
goroutine 305 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
github.com/megaease/easegress/pkg/supervisor.(*ObjectEntity).InitWithRecovery.func1()
        github.com/megaease/easegress/pkg/supervisor/object.go:385 +0x79
panic({0x2695c60, 0xc0024d5df0})
        runtime/panic.go:1038 +0x215
github.com/megaease/easegress/pkg/filter/mqttclientauth.(*MQTTClientAuth).loadAuthFromFile(0x412a25, {0xc0005e2d90, 0x29c2620})
        github.com/megaease/easegress/pkg/filter/mqttclientauth/mqttauth.go:97 +0x1ce
github.com/megaease/easegress/pkg/filter/mqttclientauth.(*MQTTClientAuth).Init(0xc002453e90, 0xc002394b40)
        github.com/megaease/easegress/pkg/filter/mqttclientauth/mqttauth.go:126 +0xc5
github.com/megaease/easegress/pkg/object/pipeline.(*Pipeline).reload(0xc0026be720, 0x0)
        github.com/megaease/easegress/pkg/object/pipeline/pipeline.go:182 +0x43d
[...]

PR #510, opened 2022-02-15
udpate mqtt client auth

suchen-sci:
This pr fix issue #505 by remove auth file and put auth usernames and passwords into yaml file. For security, we use salted SHA256 checksum to protect password.

### DIFF
--- a/pkg/filter/mqttclientauth/mqttauth.go
+++ b/pkg/filter/mqttclientauth/mqttauth.go
@@ -19,16 +19,12 @@ package mqttclientauth
 
 import (
 	"crypto/sha256"
-	"encoding/base64"
 	"encoding/hex"
-	"fmt"
-	"os"
 
 	"github.com/eclipse/paho.mqtt.golang/packets"
 	"github.com/megaease/easegress/pkg/context"
 	"github.com/megaease/easegress/pkg/logger"
 	"github.com/megaease/easegress/pkg/object/pipeline"
-	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/filter/mqttclientauth/mqttauth.go
```diff
@@ -48,20 +44,21 @@ type (
...
-	// Spec is spec for MQTTClientAuth
-	// authFile format:
-	// - username: test
-	//   passBase64: dGVzdA==
+	// Spec is spec for MQTTClientAuth.
+	// For security of password, passwords in yaml file should be salted SHA256 checksum.
+	// password = sha256sum(connect.password + salt)
 	Spec struct {
-		AuthFile string `yaml:"authFile" jsonschema:"required"`
+		Salt string  `yaml:"salt" jsonschema:"omitempty"`
```
localvar, line 54: is `Salt` a fixed value for all user/password pairs? then it is helpless for security and can be remvoed.

suchen-sci: it is fixed. I think it mainly used for solving rainbow table attack. making such table may need some time and user can update during that time. One salt for one password is also ok...

zhao-kun: I had a similar question about this, but I think @suchen-sci's explanation makes sense.

zhao-kun (approved):

pkg/filter/mqttclientauth/mqttauth.go
localvar: Yes, got it, it could be some help though I think building a rainbow table won't take too much time.

localvar (approved):

pkg/filter/mqttclientauth/mqttauth.go
suchen-sci: so one salt for one password? and also support salt for all passwords, let user to choose. If they want more secure use one salt for one passwrod?

PR #527, opened 2022-02-24
convert cluster.Mutex to goroutine level

localvar:
This PR resolves the 'session expire' issue when the client sends API requests concurrently.
however, API processing becomes slower after the change.

### DIFF
--- a/pkg/cluster/mutex.go
+++ b/pkg/cluster/mutex.go
@@ -19,6 +19,7 @@ package cluster
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"go.etcd.io/etcd/client/v3/concurrency"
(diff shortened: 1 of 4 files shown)


pkg/cluster/mutex.go
```diff
@@ -31,20 +32,30 @@ type Mutex interface {
...
 func (m *mutex) Lock() error {
+	m.lock.Lock()
+
 	ctx, cancel := context.WithTimeout(context.Background(), m.timeout)
 	defer cancel()
 
-	return m.m.Lock(ctx)
+	err := m.m.Lock(ctx)
+	if err != nil {
+		m.lock.Unlock()
```
xxx7xxxx, line 50: If we make it safest, we should use `defer` statement to unlock if the panic happened perhaps from `m.m.Lock`. Otherwise, it will cause the local lock won't be released.

localvar: when `m.m.Lock` succeeds, we should not unlock `m.lock`, it will be unlocked in `m.Unlock`.

suchen-sci: based on the implementation of etcd mutex, we need to linearize the execution of Lock and Unlock by ourself...

xxx7xxxx: > when m.m.Lock succeeds, we should not unlock m.lock, it will be unlocked in m.Unlock.

Sure, we can know it to panic or not in recovery function. The point here is `m.m.Lock` has a change to panic, then the local mutex will be locked forever. we could use a defer function to avoid it.

localvar: Thanks for the comment, updated.

pkg/cluster/mutex.go
```diff
@@ -31,20 +32,35 @@ type Mutex interface {
...
 	timeout time.Duration
 }
 
-func (m *mutex) Lock() error {
+func (m *mutex) Lock() (err error) {
+	panicked := true
+
+	m.lock.Lock()
+	defer func() {
+		if panicked || err != nil {
```
xxx7xxxx, line 47: (nit) Using `recover` is another option, this way is fine to me.

xxx7xxxx (approved):

pkg/cluster/mutex.go
localvar: This function is unable to handle the panic completely, so using `recover` requires `panic` again to expose the panic to callers, but this will change the original stack trace.

suchen-sci: this looks cool

suchen-sci (approved):

samutamm (approved):

PR #536, opened 2022-03-09
Resolve inconsistent processing logic in mux.ServeHTTP 

### LINKED ISSUE #535: Inconsistent processing logic for `matchMethod` and `matchHeaders` in `mux.ServeHTTP`.
**Describe the bug**

When I was looking at the code of `mux.ServeHTTP`, I found that the logic regarding the processing of `matchMethod` is not consistent with the logic of `matchHeaders`, the specific code range is https://github.com/megaease/easegress/blob/36480f34946b464820360ea1a535880fd7e7751f/pkg/object/httpserver/mux.go#L385-L426,

The execution flow here is as follows:
截屏2022-03-08 下午8 18 22

From the figure, we can see that in the case of `matchMethod=false`, the request will be responded directly, and in the case of `matchHeaders=false`, the next cycle will be performed, and the processing behavior of these two is inconsistent.

**Expected behavior**

I think the logic should be the same when `matchMethod` and `matchHeader` processing result is `false`, either it will go to the `next round of loop` or `respond directly to the failed request`.

I think it's better to go to the next round of the cycle, so that the following configuration can be achieved

```yaml
rules:
  - paths:
    - pathPrefix: /pipeline
      method: GET
      backend: get-demo
    
    - pathPrefix: /pipeline
      method: POST
      backend: post-demo
```

**Version**

master

**Configuration**
* Easegress Configuration
```yaml
```

* HTTP server configuration
```yaml
```

* Pipeline Configuration
```yaml
```

**Logs**
```
Easegress logs, if applicable.
```

**OS and Hardware**
 - OS: [e.g. Ubuntu 20.04]
 - CPU:[e.g. Intel(R) Core(TM) i5-8265U]
 - Memory: [e.g. 16GB]

**Additional context**

---
Thanks for contributing 🎉!

samutamm:
Fix #535.

At high level this PR aims to fix the issue, that when there are two `httpserver.mux` rules, with same path but different Method, only first one is considered.

This involves following changes:

- Extract lines https://github.com/megaease/easegress/blob/36480f34946b464820360ea1a535880fd7e7751f/pkg/object/httpserver/mux.go#L385-L426 to own function `SearchPath`
- make struct `muxPath` public: `MuxPath` so that it can be used as a return type


localvar:
IMHO, packages in the `utils` folder should only depend on the standard packages or 3rd party packages (although there are already exceptions).

The `pathsearch` package of this PR depends on the `context` package, and its interface/implementation is also tightly bound to HTTP, so I propose not to put it in the `utils` folder, and maybe we don't need to define the `MuxPathInterface`.

samutamm:
> IMHO, packages in the `utils` folder should only depend on the standard packages or 3rd party packages (although there are already exceptions).
> 
> The `pathsearch` package of this PR depends on the `context` package, and its interface/implementation is also tightly bound to HTTP, so I propose not to put it in the `utils` folder, and maybe we don't need to define the `MuxPathInterface`.

Good point, I updated the PR and the description. It now involves less refactoring

haoel:
@aniaan please help to review this PR.

pkg/object/httpserver/mux.go
```diff
@@ -382,52 +386,86 @@ func (m *mux) ServeHTTP(stdw http.ResponseWriter, stdr *http.Request) {
...
-				return
+				continue
 			}
+			// at least one path has correct method
+			methodAllowed = true
 
 			if !path.pass(ctx) {
-				m.handleIPNotAllow(ctx)
-				return
+				return IPNotAllowed, path
```
aniaan, line 453: A little suggestion, our `path matching rule` should be `matchPath + matchMethod + matchHeader`, after the above three conditions are matched, only then `request` is considered to match the `path`, so only when the above conditions are met, should we check whether the `IP is passed`.

```golang
if !path.matchMethod(ctx) {
	continue
}
// at least one path has correct method
methodAllowed = true

var searchResult SearchResult

if path.hasHeaders() {
	if !path.matchHeaders(ctx) {
		continue
	}
	searchResult = FoundSkipCache
} else {
	searchResult = Found
}

if !path.pass(ctx) {
	return IPNotAllowed, nil
}

return searchResult, path
````

Comment (aniaan) [reply pkg/object/httpserver/mux.go:453]: In this case, we can implement the following rule configuration

```yaml
rules:
  - paths:
    - pathPrefix: /pipeline
      method: POST
      headers:
       X-version: v1
      ip:
        AllowIPs:
         - "8.8.8.8"
      backend: post-v1-demo
    
    - pathPrefix: /pipeline
      method: POST
      headers:
       X-version: v2
      ip:
        AllowIPs:
         - "9.9.9.9"
      backend: post-v2-demo
```

aniaan (commented):
LGTM, one small suggestion, we should use `path + method + header` as the unique key of path, the ip check should be checked after matching the path, in addition this piece will be better if there is a unit test coverage

pkg/object/httpserver/mux.go
samutamm: Thanks for review @aniaan ! I agree that it's better to have unit tests here so that it would be easier to make these modifications, however currently it would require adding unit test for the whole `httpserver` module.

samutamm: I like your suggestion about routing based on IP and headers.. I think that when the path does not define them, route should match all headers and IPs. If header or IP is defined, it will be used to match and request causes MethodNotAllowed if no correct header or IP is present in the request.

So this will match requests with correct path and method and any IP or any or none header
```yaml
pathPrefix: /pipeline
method: POST
backend: post-v2-demo
```

and in your example it's MethodNotAllowed if request path is correct but header or IP is not present in the request. @aniaan Is this what you mean?

aniaan: Maybe my description is not very accurate, I want to say is that a `path` matching rule is `path + method + headers`, **ip is not part of the path matching**, we have to put the headers check to the front, if the `path + method + headers` are matched to complete, that the user is matching the rule, and then check whether the ip is passed, ip if passed, go backend, not passed, **directly return 403 response**.

image

localvar (approved):

xxx7xxxx:
Good refactoring, but please fix the errors

samutamm:
> Good refactoring, but please fix the errors

Fixed!

xxx7xxxx (approved):

suchen-sci (approved):

PR #537, opened 2022-03-10
fix(proxy): fallback not working

aniaan:
https://github.com/megaease/easegress/blob/36480f34946b464820360ea1a535880fd7e7751f/pkg/filter/proxy/proxy.go#L351-L360

As shown above, handle processing results equal to `""` on behalf of the request is normal, `not ""` on behalf of the occurrence of errors, fallback mechanism is said to occur after the error processing mechanism, so it should be `result ! = ""` in the case of fallback processing


localvar (approved):

samutamm (approved):
LGTM, just please consider the small suggestion from code analysis (visible in "Files changed" tab)

aniaan:
> LGTM, just please consider the small suggestion from code analysis (visible in "Files changed" tab)

Updated

PR #542, opened 2022-03-14
Split Sync statuses to smaller objects in etcd

### LINKED ISSUE #541: Reduce ETCD memory usage when having large number of HTTPPipelines
When creating a HTTPServer with 1000 HTTPPipelines, like described in this issue https://github.com/megaease/easegress/issues/413, the Easegress starts increasing memory usage and in few hours it will run out of memory on 16GB machine. This is due the fact that Easegress uses etcd key-value store and [etcd stores 5000 most recent modifications to memory](https://github.com/etcd-io/etcd/blob/29c3b0f307107fd95a6eb43ddff20db952eeb2fa/server/etcdserver/server.go#L75-L80). In most of the use cases this is totally fine, but when there are 1000 httppipelines, each modification stored to etcd is quite large (> 2MB) and having 5000 of them in memory is too much.

There are multiple ways to reduce the memory consumption:

Let's note the size of a single item stored to etcd by `y`. The total memory consumption (there are probably other factors too) is roughly `5000 * y`

To reduce the memory usage: 
1. Compress data stored to etcd. New compressed data `y' << y` and therefore `5000y' << 5000y`
2. Update some tiny dummy items in etcd for every update of real data. For example, if 10 dummy items is updated for one real data item, the real data items in memory is only `500y` and if dummy items are really small (`~ 0`), the 4500 dummy items stored in memory are `4500 ~ 0 = 0` so total memory usage is about `500y << 5000y`
3. Split the item to smaller individual items. If a item size of `y` is splitted to 1000 (because 1000 pipelines) smaller items of size `y/1000 = y'` and there are still 5000 of these small keys in the memory, the total memory consumption is 1000 times smaller.

What are the advantages and disadvantages of these options?
||Advantage|Disadvantage|branch
|--|--|--|--|
|1|Easy to implement|Needs to decompress when consuming keys and depends whether data is easy to compress |https://github.com/samutamm/easegress/tree/experiment-sync-status-compression|
|2|Easy to implement|Ugly and not correct way to do things. (in quick experience it does not work well) |https://github.com/samutamm/easegress/tree/experiment-add-dummy-etcd-keys|
|3|Clean solution and etcd best practise|Require a bit more refactoring|https://github.com/samutamm/easegress/tree/experiment-smaller-statuses|

easegress_ram_split_transaction

**Describe alternatives you've considered**

**Additional context**

---
Thanks for contributing 🎉!

samutamm:
Close #541 .

Change periodically (every 5s) called `StatusSyncController.handleStatus`.
- save each status to cluster in individual etcd transaction
- For `TrafficController` and `RawconfigSyncController`, split the large status data structure to smaller items so that each `HTTPServer` object and each `HTTPPipeline` object has unique key "${kind}-${specName}" in etcd.

Adapt pkg/api/cluster.go to these changes:
- `egctl object status list` does not require changes
- `egctl object status get ${httpPipelineName|httpServerName}` uses `_getStatusObjectFromTrafficController` --> adapt this function to read new status data from etcd

### DIFF
--- a/pkg/object/statussynccontroller/statussynccontroller.go
+++ b/pkg/object/statussynccontroller/statussynccontroller.go
@@ -165,19 +165,20 @@ func safeMarshal(value *supervisor.Status) (string, bool) {
 	return string(buff), true
 }
 
-func splitRawconfigTrafficControllerStatus(
-	name string,
+func (ssc *StatusSyncController) splitRawconfigTrafficControllerStatus(
+	kind string,
 	status *trafficcontroller.StatusInSameNamespace,
 	statuses map[string]string,
 	statusesRecord *StatusesRecord) bool {
 	for key, value := range status.ToSyncStatus() {
-		statusesRecord.Statuses[name+"-"+key] = value
+		name := ssc.superSpec.Super().Cluster().Layout().StatusObjectName(kind, key)
+		statusesRecord.Statuses[name] = value
 
 		marshalledValue, ok := safeMarshal(value)
 		if !ok {
 			return false
 		}
-		statuses[name+"-"+key] = marshalledValue
+		statuses[name] = marshalledValue
 	}
 	return true
 }
@@ -205,13 +206,13 @@ func (ssc *StatusSyncController) handleStatus(unixTimestamp int64) {
 		if trafficStatus, ok := status.ObjectStatus.(*trafficcontroller.Status); ok {
 			statusInNamespaces := trafficStatus.Specs
 			for _, statInNS := range statusInNamespaces {
-				if !splitRawconfigTrafficControllerStatus(name, statInNS, statuses, statusesRecord) {
+				if !ssc.splitRawconfigTrafficControllerStatus(name, statInNS, statuses, statusesRecord) {
 					return false
 				}
 			}
 			return true
 		} else if rawTrafficStatus, ok := status.ObjectStatus.(*rawconfigtrafficcontroller.Status); ok {
-			return splitRawconfigTrafficControllerStatus(name, rawTrafficStatus, statuses, statusesRecord)
+			return ssc.splitRawconfigTrafficControllerStatus(name, rawTrafficStatus, statuses, statusesRecord)
 		} else {
 			statusesRecord.Statuses[name] = status
 			marshalledValue, ok := safeMarshal(status)
--- a/pkg/object/statussynccontroller/statussynccontroller.go
+++ b/pkg/object/statussynccontroller/statussynccontroller.go
@@ -233,28 +233,25 @@ func (ssc *StatusSyncController) handleStatus(unixTimestamp int64) {
 func (ssc *StatusSyncController) syncStatusToCluster(statuses map[string]string) {
 	// Delete statuses which disappeared in current status.
 	if ssc.lastSyncStatuses != nil {
+		kvs := make(map[string]*string)
 		for k := range ssc.lastSyncStatuses {
 			if _, exists := statuses[k]; !exists {
-				kv := make(map[string]*string)
 				k = ssc.superSpec.Super().Cluster().Layout().StatusObjectKey(k)
-				kv[k] = nil
-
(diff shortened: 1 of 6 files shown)


pkg/object/statussynccontroller/statussynccontroller.go
```diff
@@ -192,29 +231,34 @@ func (ssc *StatusSyncController) handleStatus(unixTimestamp int64) {
...
-				kvs[k] = nil
+				kv[k] = nil
+
+				err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kv)
+				if err != nil {
+					logger.Errorf("sync status failed. If the message size is too large, "+
+						"please increase the value of cluster.MaxCallSendMsgSize in configuration: %v", err)
+				}
 			}
 		}
```
localvar, line 248: propose to do delete in one call to PutAndDeleteUnderLease, that's refactor this piece of code to:
```go
                kv := make(map[string]*string)
		for k := range ssc.lastSyncStatuses {
			if _, exists := statuses[k]; !exists {
				k = ssc.superSpec.Super().Cluster().Layout().StatusObjectKey(k)
				kv[k] = nil
			}
		}
		err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kv)
		if err != nil {
			logger.Errorf("sync status failed. If the message size is too large, "+
				"please increase the value of cluster.MaxCallSendMsgSize in configuration: %v", err)
		}
```

pkg/object/statussynccontroller/statussynccontroller.go
```diff
@@ -192,29 +231,34 @@ func (ssc *StatusSyncController) handleStatus(unixTimestamp int64) {
...
 		k = ssc.superSpec.Super().Cluster().Layout().StatusObjectKey(k)
-		kvs[k] = &v
-	}
-
-	err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kvs)
-	if err != nil {
-		logger.Errorf("sync status failed. If the message size is too large, "+
-			"please increase the value of cluster.MaxCallSendMsgSize in configuration: %v", err)
+		kv[k] = &v
+		err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kv)
```
localvar, line 257: propose to call `PutUnderLease` and avoid creating the kv map.

samutamm:
Here's a memory usage of the same scenario as described in issues https://github.com/megaease/easegress/issues/413 and https://github.com/megaease/easegress/issues/541 : 
easegress_73c72bd_memory_consumption


The memory usage is captured by running `free -s 2` that outputs memory usage every 2 seconds. The two tests have exactly same code, but the difference is that for the orange one has an extra process that runs `egctl object update -f 'path/to/pipelines'` every 5 minutes. The two tests were run in different VMs that  might also differ slightly. However there was no more the issue of growing memory usage during first 1h 40 minutes of execution.

pkg/object/statussynccontroller/statussynccontroller.go
```diff
@@ -192,29 +241,36 @@ func (ssc *StatusSyncController) handleStatus(unixTimestamp int64) {
...
-		k = ssc.superSpec.Super().Cluster().Layout().StatusObjectKey(k)
-		kvs[k] = &v
-	}
-
-	err := ssc.superSpec.Super().Cluster().PutAndDeleteUnderLease(kvs)
-	if err != nil {
-		logger.Errorf("sync status failed. If the message size is too large, "+
-			"please increase the value of cluster.MaxCallSendMsgSize in configuration: %v", err)
+	kvs := make(map[string]*string)
+	for k, value := range statuses {
```
localvar, line 263: if `len(kvs) > 0`, should call `PutAndDeleteUnderLease` after the for loop too.

samutamm: Fixed!

localvar (approved):

suchen-sci (approved):

xxx7xxxx (approved):

PR #543, opened 2022-03-16
Turn profiling on/off runtime

### LINKED ISSUE #532: enable/disable profile at runtime
Easegress could generate CPU & memory profiles for troubleshooting, but this feature can only be enabled from the configuration at startup.

However, some issues only happen after Easegress runs for a long time, if profiling is not enabled at first, it will be very hard to reproduce the issue.

So, we need to add new admin APIs to enable/disable profiles at runtime.

samutamm:
...using admin API.

Close #532

### DIFF
--- a/cmd/client/command/profile.go
+++ b/cmd/client/command/profile.go
@@ -52,7 +52,7 @@ func infoProfileCmd() *cobra.Command {
 func startProfilingCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "start",
-		Short: "Prepare measuring CPU or memory.",
+		Short: "Prepare measuring CPU or memory",
 	}
 	cmd.AddCommand(startCPUCmd())
 	cmd.AddCommand(startMemoryCmd())
@@ -63,7 +63,7 @@ func startProfilingCmd() *cobra.Command {
 func startCPUCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "cpu",
-		Short:   "Prepare measuring CPU.",
+		Short:   "Prepare measuring CPU",
 		Example: "egctl profile start cpu <path/to/cpu-prof-file>",
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
@@ -83,7 +83,7 @@ func startCPUCmd() *cobra.Command {
 func startMemoryCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "memory",
-		Short:   "Prepare measuring memory.",
+		Short:   "Prepare measuring memory",
 		Example: "egctl profile start memory <path/to/memory-prof-file>",
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
--- a/cmd/client/command/profile.go
+++ b/cmd/client/command/profile.go
@@ -40,7 +40,7 @@ func ProfileCmd() *cobra.Command {
 func infoProfileCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "info",
-		Short: "Show memory and CPU profile file paths.",
+		Short: "Show memory and CPU profile file paths",
 		Run: func(cmd *cobra.Command, args []string) {
 			handleRequest(http.MethodGet, makeURL(profileURL), nil, cmd)
 		},
--- a/cmd/client/command/profile.go
+++ b/cmd/client/command/profile.go
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2017, MegaEase
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


pkg/api/profile.go
```diff
@@ -0,0 +1,114 @@
...
+	"sync"
+)
+
+const (
+	// ProfilePrefix is the URL prefix of profile APIs
+	ProfilePrefix = "/profile"
+	// StartAction is the URL for starting profiling
+	StartAction = "/start"
+	// StopAction is the URL for stopping profiling
+	StopAction = "/stop"
```
localvar, line 33:
```suggestion
	StartAction = "start"
	// StopAction is the URL for stopping profiling
	StopAction = "stop"
```

pkg/api/profile.go
```diff
@@ -0,0 +1,114 @@
...
+			Method:  http.MethodPost,
+			Handler: s.startCPUProfile,
+		},
+		{
+			Path:    fmt.Sprintf("%s%s/memory", ProfilePrefix, StartAction),
+			Method:  http.MethodPost,
+			Handler: s.startMemoryProfile,
+		},
+		{
+			Path:    fmt.Sprintf("%s%s", ProfilePrefix, StopAction),
```
localvar, line 67:
```suggestion
			Path:    fmt.Sprintf("%s/%s/cpu", ProfilePrefix, StartAction),
			Method:  http.MethodPost,
			Handler: s.startCPUProfile,
		},
		{
			Path:    fmt.Sprintf("%s/%s/memory", ProfilePrefix, StartAction),
			Method:  http.MethodPost,
			Handler: s.startMemoryProfile,
		},
		{
			Path:    fmt.Sprintf("%s/%s", ProfilePrefix, StopAction),
```

pkg/api/profile.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (s *Server) startCPUProfile(w http.ResponseWriter, r *http.Request) {
+	spr := StartProfilingRequest{}
+	err := yaml.NewDecoder(r.Body).Decode(&spr)
+	if err != nil {
+		HandleAPIError(w, r, http.StatusBadRequest, fmt.Errorf("bad request"))
+		return
+	}
+
+	s.opt.CPUProfileFile = spr.Path
```
localvar, line 94: need to check if spr.Path is valid or not.

cmd/client/command/profile.go
```diff
@@ -0,0 +1,114 @@
...
+	cmd.AddCommand(infoProfileCmd())
+	cmd.AddCommand(startProfilingCmd())
+	cmd.AddCommand(stopProfilingCmd())
+	return cmd
+}
+
+func infoProfileCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "info",
+		Short: "Show memory and CPU profile file paths.",
```
xxx7xxxx, line 43:
```suggestion
		Short: "Show memory and CPU profile file paths",
```

Consistent with other commends, no period in short usage. So are others

pkg/api/profile.go
xxx7xxxx, line 94: I don't think directly changing fields of options is a good idea. The options should truly reflect the static options, which will be put into etcd as member info. So my proposal is to make function be like `profile.StartCPUProfile(path string)` which uses `options.CPUProfilePath` only if path is empty.

localvar:
@samutamm , seems two different changes are in the PR

samutamm:
> @samutamm , seems two different changes are in the PR

Oups! Now it looks ok again

pkg/api/profile.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (s *Server) startCPUProfile(w http.ResponseWriter, r *http.Request) {
+	spr := StartProfilingRequest{}
+	err := yaml.NewDecoder(r.Body).Decode(&spr)
+	if err != nil {
+		HandleAPIError(w, r, http.StatusBadRequest, fmt.Errorf("bad request"))
+		return
+	}
+
+	if spr.Path == "" || !IsPathValid(spr.Path) {
```
localvar, line 95: I comment before to validate `spr.Path`, but I think we only need to check if it is an empty string.
because even `IsPathValid` returns true, it does not mean the path is valid when it is actually used.
and for L99, we can update `UpdateCPUProfile` to return an error and forward the error to the client.

localvar (approved):

suchen-sci (approved):

pkg/profile/profile.go
```diff
@@ -45,20 +57,38 @@ func New(opt *option.Options) (Profile, error) {
...
+		return ""
+	}
+	return p.cpuFile.Name()
+}
+
+func (p *profile) MemoryFileName() string {
+	return p.memFileName
+}
+
+func (p *profile) UpdateMemoryProfile(fp string) {
```
xxx7xxxx, line 79: There could be concurrent calls from the network, we need to handle it, don't we? So do others.

samutamm: I think that we can suppose that there is only one administrator, who wants to turn on/off profiling at once.. However do you think that concurrent calls can cause runtime errors? I'll add the lock like in other admin APIs.

xxx7xxxx: We can't make assumptions from users. Runtime errors would be many kinds of race conditions.

samutamm: Yeah true.. Now it should be better, with lock for modifications.

pkg/api/profile.go
```diff
@@ -94,6 +94,10 @@ func (s *Server) startCPUProfile(w http.ResponseWriter, r *http.Request) {
 		HandleAPIError(w, r, http.StatusBadRequest, fmt.Errorf("missing path"))
 		return
 	}
+
+	s.Lock()
```
xxx7xxxx, line 98: You misused the lock which is the cluster-level config lock. The lock we should use is the localprofile-leveled.

xxx7xxxx (approved):

PR #551, opened 2022-03-28
valid circuit-breaker filter spec by correlation

sodaRyCN:
The circuit breaker filter should valid spec according to the parameter correlation: using the strategy of failure statistics, it should ensure that the relevant required parameters have been set(failureRateThreshold,failureStatusCodes). At this time, the relevant parameters of the slow call are not required, and vice versa.

### DIFF
--- a/pkg/filter/circuitbreaker/circuitbreaker.go
+++ b/pkg/filter/circuitbreaker/circuitbreaker.go
@@ -49,8 +49,8 @@ type (
 	Policy struct {
 		Name                             string `yaml:"name" jsonschema:"required"`
 		SlidingWindowType                string `yaml:"slidingWindowType"  jsonschema:"omitempty,enum=COUNT_BASED,enum=TIME_BASED"`
-		FailureRateThreshold             uint8  `yaml:"failureRateThreshold" jsonschema:"omitempty,minimum=1,maximum=100"`
-		SlowCallRateThreshold            uint8  `yaml:"slowCallRateThreshold" jsonschema:"omitempty,minimum=1,maximum=100"`
+		FailureRateThreshold             uint8  `yaml:"failureRateThreshold" jsonschema:"omitempty,maximum=100"`
+		SlowCallRateThreshold            uint8  `yaml:"slowCallRateThreshold" jsonschema:"omitempty,maximum=100"`
 		CountingNetworkError             bool   `yaml:"countingNetworkError" jsonschema:"omitempty"`
 		SlidingWindowSize                uint32 `yaml:"slidingWindowSize" jsonschema:"omitempty,minimum=1"`
 		PermittedNumberOfCallsInHalfOpen uint32 `yaml:"permittedNumberOfCallsInHalfOpenState" jsonschema:"omitempty"`
@@ -87,8 +87,8 @@ type (
 	}
 )
 
-// Validate implements custom validation for Spec
-func (spec Spec) Validate() error {
+// check policy of url usage whether defined
+func (spec Spec) validateURLPoliciesUsage() error {
 URLLoop:
 	for _, u := range spec.URLs {
 		name := u.PolicyRef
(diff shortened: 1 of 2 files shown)


samutamm (approved):
LGTM!

pkg/filter/circuitbreaker/circuitbreaker.go
```diff
@@ -108,6 +108,40 @@ URLLoop:
...
+		//group of slow
+		if p.SlowCallRateThreshold != 0 && (p.SlowCallDurationThreshold == "" || strings.TrimSpace(p.SlowCallDurationThreshold) == "") {
+			return fmt.Errorf("policy '%s' has setted slow call threshold, but not set slow call duration ", p.Name)
+		}
+		if p.SlowCallRateThreshold == 0 && (p.SlowCallDurationThreshold != "" || strings.TrimSpace(p.SlowCallDurationThreshold) != "") {
+			return fmt.Errorf("policy '%s' has setted slow call duration, but not set slow call threshold ", p.Name)
+		}
+	}
+	return nil
+}
```
localvar, line 130: Please check the `buildPolicy` function below, if `FailureRateThreshold`, `SlowCallRateThreshold`, `SlowCallDurationThreshold` is 0(or emtpy), default values will be used, so it is not an error.

And in the `handle` function, we can see `FailureRateThreshold` could be used with an empty `FailureStatusCodes`.

So I think these validations are not necessary, and we only need to update the `jsonSchema` to remove `minimum`.

sodaRyCN: `And in the handle function, we can see FailureRateThreshold could be used with an empty FailureStatusCodes.`
Does it mean CountingNetworkError? I really missed it here. It seems to be related to the Failure policy. When the user does not specify the FailureStatusCode, the error code defined by eg is used (500/499/...). I think it would be
`if p.FailureRateThreshold != 0 && (len(p.FailureStatusCodes) == 0 || !p.CountingNetworkError)`

My understanding is that when the user configures the circuit breaker, his expectation is to trigger the circuit breaker under certain conditions, so the legality check item should be to ensure that his configuration can correctly enable the circuit breaker, instead of being silent because of the wrong configuration . As for the default value in `handle`, I wonder if the user knows and accepts the default value. It seems that the document does not mention the default value, so when the user mistakenly omits some parameters, the effect of the default value may not be what the user expects, and he may spend part of his time analyzing the positioning.
And the variable initialization in `handle` I prefer to understand it as a solution to the null pointer or to distinguish between the user's explicit setting and the eg's own setting.

ex: When users only set the SlowCallRateThreshold without setting the SlowCallDurationThreshold, eg will set the SlowCallDurationThreshold to 1 minute, and what the user sees is that the requests he tested are assumed to be returned in 5s, then the circuit breaker will not be triggered, and the same is true for Failure.

localvar: > `And in the handle function, we can see FailureRateThreshold could be used with an empty FailureStatusCodes.` Does it mean CountingNetworkError? I really missed it here. It seems to be related to the Failure policy. When the user does not specify the FailureStatusCode, the error code defined by eg is used (500/499/...). I think it would be `if p.FailureRateThreshold != 0 && (len(p.FailureStatusCodes) == 0 || !p.CountingNetworkError)`

Yes, the check should be  `if p.FailureRateThreshold != 0 && (len(p.FailureStatusCodes) == 0 || !p.CountingNetworkError)`

> My understanding is that when the user configures the circuit breaker, his expectation is to trigger the circuit breaker under certain conditions, so the legality check item should be to ensure that his configuration can correctly enable the circuit breaker, instead of being silent because of the wrong configuration . As for the default value in `handle`, I wonder if the user knows and accepts the default value. It seems that the document does not mention the default value, so when the user mistakenly omits some parameters, the effect of the default value may not be what the user expects, and he may spend part of his time analyzing the positioning. And the variable initialization in `handle` I prefer to understand it as a solution to the null pointer or to distinguish between the user's explicit setting and the eg's own setting.
> 
> ex: When users only set the SlowCallRateThreshold without setting the SlowCallDurationThreshold, eg will set the SlowCallDurationThreshold to 1 minute, and what the user sees is that the requests he tested are assumed to be returned in 5s, then the circuit breaker will not be triggered, and the same is true for Failure.

This is a document issue, `SlowCallDurationThreshold` should be listed in [this table](https://github.com/megaease/easegress/blob/main/doc/reference/filters.md#circuitbreakerpolicy)

sodaRyCN: @localvar @samutamm 
modified according to the discussion comments, please review again

pkg/filter/circuitbreaker/circuitbreaker.go
```diff
@@ -108,6 +108,28 @@ URLLoop:
 	return nil
 }
 
+func (spec Spec) validatePoliciesSpec() error {
+	for _, p := range spec.Policies {
+		if p.FailureRateThreshold != 0 && len(p.FailureStatusCodes) == 0 && !p.CountingNetworkError {
+			return fmt.Errorf("policy '%s' has setted failure threshold and countingNetworkError is false, but not set failure status code", p.Name)
```
localvar, line 114:
```suggestion
			return fmt.Errorf("policy '%s' has set failure threshold and countingNetworkError is false, but not set failure status code", p.Name)
```

sodaRyCN: done

localvar (approved):

Issue #552, opened 2022-03-29
`PathPrefix` and `Full Path` support rewriteTarget


aniaan:
**Is your feature request related to a problem? Please describe.**

Currently only "pathRegexp" supports rewriteTarget, in some scenarios, it would be better for path and pathPrefix to support rewriteTarget

https://github.com/megaease/easegress/blob/0cd6b15620601de3021abbca1e310120a566795d/pkg/object/httpserver/mux.go#L508-L513

**Describe the solution you'd like**

Due to some historical reasons, the API design is not standardized. At present, when accessing easegress, it is necessary to develop a new standard API interface, rewrite the old API into a new API, and then forward it through proxy.

During use, it was found that only **pathRegexp** supports rewriteTarget. In most of our scenarios, using **prefix** can basically solve most problems, a small amount can be solved using **full path**, and **pathRegexp** is basically not used, **although prefix can solve the problem, Regexp can also be solved, but in terms of program execution efficiency, prefix performance will be higher**, so I think, we need to support the path rewriting of prefix and full path

eg: 

```golang
// old api: /tianji/exchange/activity/claim/participate
// new api: /api/activity/ex/clain/participate
// old api rewrite target to new api
// use pathRegexp or prefix

package main

import (
	"fmt"
	"regexp"
	"strings"
	"time"
)

func main() {
	n := 100000
	pt := prefix(n)
	rt := regex(n)

	fmt.Println("prefixPath: ", pt)
	fmt.Println("regexPath: ", rt)

}

func prefix(n int) time.Duration {
	repl := "/api/activity/ex/"
	prefix := "/tianji/exchange/activity/"
	path := "/tianji/exchange/activity/claim/participate"
	now := time.Now()

	for i := 0; i < n; i++ {
		path = strings.Replace(path, prefix, repl, 1)
	}

	return time.Since(now)

}

func regex(n int) time.Duration {
	repl := "/api/activity/ex/$1"
	pattern := "^/tianji/exchange/activity/(.*)$"
	re, _ := regexp.Compile(pattern)

	path := "/tianji/exchange/activity/claim/participate"
	now := time.Now()

	for i := 0; i < n; i++ {
		path = re.ReplaceAllString(path, repl)
	}

	return time.Since(now)
}



```

**Additional context**

If you agree with the change, I will follow up with the corresponding PR

---
Thanks for contributing 🎉!

samutamm:
Thanks for opening the issue!

When I run this on my workstation, I get following output
```bash
[...]

PR #553, opened 2022-03-29
perf(MuxPath): prefix and fullpath support rewrite_target

aniaan:
close: #552

### DIFF
--- a/pkg/object/httpserver/mux.go
+++ b/pkg/object/httpserver/mux.go
@@ -78,29 +78,18 @@ type (
 		ipFilter      *ipfilter.IPFilter
 		ipFilterChain *ipfilter.IPFilters
 
-		path            string
-		pathPrefix      string
-		pathRegexp      string
-		pathRE          *regexp.Regexp
-		methods         []string
-		rewriteTarget   string
-		backend         string
-		headers         []*Header
-		matchedPathType PathType
+		path          string
+		pathPrefix    string
+		pathRegexp    string
+		pathRE        *regexp.Regexp
+		methods       []string
+		rewriteTarget string
+		backend       string
+		headers       []*Header
 	}
 
 	// SearchResult is returned by SearchPath
 	SearchResult string
-
-	// PathType is the type of path
-	PathType string
-)
-
-// kinds of path
-const (
-	PATH   PathType = "path"
-	PREFIX PathType = "prefix"
-	REGEXP PathType = "regexp"
 )
 
 // newIPFilterChain returns nil if the number of final filters is zero.
@@ -260,15 +249,14 @@ func (mp *MuxPath) matchPath(ctx context.HTTPContext) bool {
 	}
 
 	if mp.path != "" && mp.path == r.Path() {
-		mp.matchedPathType = PATH
 		return true
 	}
+
 	if mp.pathPrefix != "" && strings.HasPrefix(r.Path(), mp.pathPrefix) {
-		mp.matchedPathType = PREFIX
 		return true
 	}
+
 	if mp.pathRE != nil && mp.pathRE.MatchString(r.Path()) {
-		mp.matchedPathType = REGEXP
 		return true
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/object/httpserver/mux.go
```diff
@@ -261,6 +261,31 @@ func (mp *MuxPath) matchPath(ctx context.HTTPContext) bool {
...
+		path = strings.Replace(path, mp.pathPrefix, mp.rewriteTarget, 1)
+		ctx.Request().SetPath(path)
+		return
+	}
+
+	if mp.pathRE != nil {
+		path = mp.pathRE.ReplaceAllString(path, mp.rewriteTarget)
+		ctx.Request().SetPath(path)
+		return
+	}
```
localvar, line 286: This is incorrect, for example:
if both `mp.path` and `mp.pathPrefix` are not empty and the request matches `mp.pathPrefix`, the path should be rewritten at L277, but was actually rewritten at L272, which is wrong.

samutamm:
This change could be documented here: [controllers.md](https://github.com/megaease/easegress/blob/main/doc/reference/controllers.md?plain=1#L394) . Old doc

`Use pathRegexp.[ReplaceAllString](https://golang.org/pkg/regexp/#Regexp.ReplaceAllString)(path, rewriteTarget) to rewrite request path` by

New doc:

`Use pathRegexp.[ReplaceAllString](https://golang.org/pkg/regexp/#Regexp.ReplaceAllString)(path, rewriteTarget) or pathPrefix [strings.Replace](https://pkg.go.dev/strings#Replace) to rewrite request path`

pkg/object/httpserver/mux.go
aniaan: I missed the point that path and prefix and regexp can co-exist, I previously thought they were a mutually exclusive relationship, I'll fix that

aniaan:
1. If path, prefix, and regexp are empty, but rewriteTarget is not empty, the newly created spec will fail to verify. For the existing spec, rewriteTarget will not take effect, because I didn't expect the three to be empty at present, but The scenario where the path is still to be rewritten, so a check is added
2. There is also the handleRewrite function, you can see if it is appropriate to do so

@samutamm @localvar

pkg/object/httpserver/spec.go
```diff
@@ -198,3 +199,13 @@ func (h *Header) Validate() error {
 
 	return nil
 }
+
+// Validate validates Path.
+
```
samutamm: nit: empty line between comment and function

pkg/object/httpserver/mux.go
```diff
@@ -78,18 +78,28 @@ type (
...
 
 	// SearchResult is returned by SearchPath
 	SearchResult string
+
+	// PathType is the type of path
+	PathType string
+)
+
+const (
+	PATH   PathType = "path"
```
samutamm, line 101: nit: please comment PATH, PREFIX and REGEXP

samutamm (approved):

pkg/object/httpserver/mux.go
localvar: I propose not to introduce the `PathType`, it brings extra complexity and only a few benefits.
The `handleRewrite` function could be changed to below and other code could be changes accordingly:
```go
func (mp *MuxPath) handleRewrite(req context.HTTPRequest) {
	if mp.rewriteTarget == "" {
		return
	}

	path := req.Path()

	if mp.path != "" && mp.path == path {
		req.SetPath(mp.rewriteTarget)
		return
	}
	
	if mp.pathPrefix != "" && strings.HasPrefix(path, mp.pathPrefix) {
		path = mp.pathPrefix + path[len(mp.pathPrefix):]
		req.SetPath(path)
		return
	}
	
	// If we get here, then we are sure the following is true:
	//    mp.pathRE != nil && mp.pathRE.MatchString(path)
	path = mp.pathRE.ReplaceAllString(path, mp.rewriteTarget)
	req.SetPath(path)
}
```

samutamm (approved):

PR #572, opened 2022-04-07
feat(helm): log support custom path and use emptyDir volume 

aniaan:
1. log  support custom path
2. use emptyDir volume, consistent with secondary


samutamm (commented):
Thank you for the PR! Looks good, I would just add few comments to values.yaml. Also `helm-charts/easegress/README.md` and `helm-charts/ingress-controller/README.md`
could contain an extra line in [Parameters](https://github.com/megaease/easegress/blob/main/helm-charts/easegress/README.md#parameters) section, for example
```
| log.path | string | `/opt/easegress/log` | log path inside container |
```

samutamm (approved):

localvar (approved):

PR #579, opened 2022-04-11
Proxy to send zipkin b3 headers

### LINKED ISSUE #578: Proxy filter to propagate Openzipkin b3 headers
Easegress supports sending tracing spans per request to [Zipkin](https://zipkin.io/) tracing system. These spans can help to troubleshoot latency problems in the downstream back-end systems that Easegress pipelines connect to.

Though sometimes the span information comes in incoming requests in [b3-propagation](https://github.com/openzipkin/b3-propagation) headers and it should be passed along to the downstream back-ends. Currently Easegress does not propagate these headers.

When b3 headers are present, propagate them in the Proxy filter to the downstream back-end servers, as desribed in https://github.com/openzipkin/b3-propagation . If b3 headers are not present, generate a new span and inject to it to the header.

**Describe alternatives you've considered**

**Additional context**

---
Thanks for contributing 🎉!

samutamm:
Fix #578 

- [opentracing](https://github.com/opentracing/opentracing-go) is deprecated (in favor of open-telemetry) so remove it and use "native" zipkin for tracing
- call zipkin-go SDK and remove explicit creation of spans and `span.Finish()`
- When tracing object is configured in HTTPServer, share the created `tracer` instance to Proxy filter via HTTPContext.
- Use zipkin to inject B3 headers if not present already

### DIFF
--- a/pkg/filter/proxy/proxy.go
+++ b/pkg/filter/proxy/proxy.go
@@ -353,7 +353,7 @@ func (b *Proxy) updateAndGetClient(tracingInstance *tracing.Tracing) *Client {
 	if client.tracing == tracingInstance {
 		return client
 	}
-
+	// tracingInstance is updated so recreate http.Client
 	newClient := NewClient(b.createHTTPClient(), tracingInstance)
 	b.client.Store(newClient)
 	return newClient
--- a/pkg/filter/proxy/proxy.go
+++ b/pkg/filter/proxy/proxy.go
@@ -82,6 +82,7 @@ type (
 	Client struct {
 		client       *http.Client
 		zipkinClient *zipkinhttp.Client
+		tracing      *tracing.Tracing
 	}
 
 	// Spec describes the Proxy.
@@ -121,21 +122,21 @@ type (
 // NewClient creates a wrapper around http.Client
 func NewClient(cl *http.Client, tr *tracing.Tracing) *Client {
 	var zClient *zipkinhttp.Client
-	if tr != nil {
+	if tr != nil && tr != tracing.NoopTracing {
 		tracer := tr.Tracer
 		zClient, _ = zipkinhttp.NewClient(tracer, zipkinhttp.WithClient(cl))
 	}
 	return &Client{client: cl, zipkinClient: zClient}
 }
 
-// IsTracingInitialized returns true if tracing is initialized
-func (c *Client) IsTracingInitialized() bool {
-	return c.zipkinClient != nil
+// IsTracingUpdated returns true if tracing is updated
+func (c *Client) IsTracingUpdated(newTracing *tracing.Tracing) bool {
+	return c.tracing != newTracing
 }
 
 // Do calls the correct http client
 func (c *Client) Do(r *http.Request) (*http.Response, error) {
-	if c.IsTracingInitialized() {
+	if c.zipkinClient != nil {
 		return c.zipkinClient.DoWithAppSpan(r, r.URL.Path)
 	}
 	return c.client.Do(r)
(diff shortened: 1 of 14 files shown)


pkg/object/httpserver/mux.go
```diff
@@ -349,6 +349,10 @@ func (m *mux) reloadRules(superSpec *supervisor.Spec, muxMapper protocol.MuxMapp
 	} else if oldRules.tracer != nil {
 		tracer = oldRules.tracer
 	}
+	super := superSpec.Super()
+	if super != nil {
+		super.SetTracing(tracer) // share tracing object with proxy filters
+	}
```
localvar, line 355: please check if there's a race condition with the call to `super.SetTracing`.

pkg/filter/proxy/proxy.go
```diff
@@ -348,6 +379,17 @@ func (b *Proxy) handle(ctx context.HTTPContext) (result string) {
...
 
+	super := b.filterSpec.Super()
+	if super != nil {
+		tracing := super.Tracing()
+		if !tracing.IsNoopTracer() && !b.client.IsTracingInitialized() {
+			b.client.UpdateTracing(tracing)
+		} else if tracing.IsNoopTracer() && b.client.IsTracingInitialized() {
+			// if HTTPServer is updated and there is no longer tracing configured
+			b.reload()
+		}
```
localvar, line 390: please check if there are race conditions.

pkg/object/httpserver/mux.go
samutamm: Good catch! Changing supervisor.tracing to `atomic.Value` should fix race conditions

pkg/filter/proxy/proxy.go
samutamm: Now the proxy.client is protected by mutex

pkg/filter/proxy/proxy.go
```diff
@@ -315,6 +351,35 @@ func (b *Proxy) Handle(ctx context.HTTPContext) (result string) {
...
+	}
+	// Write lock to block readers while updating b.client
+	b.clientMutex.Lock()
+	defer b.clientMutex.Unlock()
+
+	if !tracing.IsNoopTracer() && !b.client.IsTracingInitialized() {
+		// tracing not ready yet
+		b.client = NewClient(b.client.client, tracing)
+	} else if tracing.IsNoopTracer() && b.client.IsTracingInitialized() {
+		// if HTTPServer is updated and there is no longer tracing configured
```
localvar, line 377: `super.Tracing` is used as a singleton is EG, so if more than one HTTPServers are configured, and they are configured with different tracing options, what will the `super.Tracing` be?

samutamm: I changed proxy to use the tracing instance of a HTTPServer, so each HTTPServer can have different tracing options. Also HTTPServer passes tracing instance to proxy via Context and supervisor is not used for that.

pkg/filter/proxy/proxy.go
```diff
@@ -315,6 +353,20 @@ func (b *Proxy) Handle(ctx context.HTTPContext) (result string) {
 	return ctx.CallNextHandler(result)
 }
 
+func (b *Proxy) updateAndGetClient(tracingInstance *tracing.Tracing) *Client {
+	b.clientMutex.RLock()
+	needsUpdate := b.client.IsTracingUpdated(tracingInstance)
+	b.clientMutex.RUnlock()
+	if !needsUpdate {
+		return b.client
+	}
```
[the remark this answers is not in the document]
samutamm: Thanks for the idea, updated! However when `tracingInstance` changes, http.Client has to be recreated (`b.createHTTPClient()`). Otherwise if the http.Client is reused, zipkin wrapper hangs for some reason in [DoWithAppSpan](https://github.com/openzipkin/zipkin-go/blob/master/middleware/http/client.go#L111)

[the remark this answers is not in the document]
localvar: Cool, thanks for the information.

localvar (approved):

suchen-sci (approved):

PR #580, opened 2022-04-13
Update Dockerfile

jthann:
In the docker built, the logger time displaying is always 8 hours earlier, by adding ```ln -s /usr/share/zoneinfo/Asia/Shanghai /etc/localtime``` in Dockerfile can fix this


localvar:
hi @jthann , thanks for the effort, but we could not accept this PR.

The fix works for users from China but will confuse users from other parts of the world.
And if local time is desired, commands like  `docker run -d -v /etc/localtime:/etc/localtime:ro -v /etc/timezone:/etc/timezone:ro megaease/easegress` could be used to direct Easegress to use the time of the server.

PR #595, opened 2022-04-27
feat(etcd): support external standalone etcd

### LINKED ISSUE #584: support external standalone etcd

Currently easegress uses the embed etcd, and there are some problems at present: 
1. It does not support primary horizontal expansion, and the number of nodes can only be determined at the beginning. 
2. Some monitoring indicators of etcd have not yet been collected and displayed. #526 

If eaesgress can support external etcd, the operation, maintenance and development work related to etcd can be alleviated, and these things can be done by special etcd tools (expansion + metrics collection visualization)

eg:  https://github.com/tkestack/kstone

---
Thanks for contributing 🎉!

aniaan:
close #584

1. add flag: use-standalone-etcd: bool
2. if use-standalone-etcd is true, all node role is **secondary**

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -375,16 +375,14 @@ func (c *cluster) checkClusterName() error {
 			logger.Errorf("%v", err)
 			panic(err)
 		}
-	} else {
-		if c.opt.UseStandaloneEtcd {
-			err := c.Put(c.Layout().ClusterNameKey(), c.opt.ClusterName)
-			if err != nil {
-				return fmt.Errorf("register cluster name %s failed: %v",
-					c.opt.ClusterName, err)
-			}
-		} else {
-			return fmt.Errorf("key %s not found", c.Layout().ClusterNameKey())
+	} else if c.opt.UseStandaloneEtcd {
+		err := c.Put(c.Layout().ClusterNameKey(), c.opt.ClusterName)
+		if err != nil {
+			return fmt.Errorf("register cluster name %s failed: %v",
+				c.opt.ClusterName, err)
 		}
+	} else {
+		return fmt.Errorf("key %s not found", c.Layout().ClusterNameKey())
 	}
 
 	return nil
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -368,15 +368,23 @@ func (c *cluster) checkClusterName() error {
 		return fmt.Errorf("failed to check cluster name: %v", err)
 	}
 
-	if value == nil {
-		return fmt.Errorf("key %s not found", c.Layout().ClusterNameKey())
-	}
-
-	if c.opt.ClusterName != *value {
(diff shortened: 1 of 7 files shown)


pkg/option/option.go
```diff
@@ -245,6 +276,11 @@ func (opt *Options) Parse() (string, error) {
 	}
 
 	opt.renameLegacyClusterRoles()
+
+	if opt.UseStandaloneEtcd {
+		opt.ClusterRole = "secondary"
```
samutamm, line 281: It could be more visible / explicit that when using "standalone" etcd, the cluster role cannot be "primary". Logging is not yet initialized here but maybe a comment:
```suggestion
		opt.ClusterRole = "secondary" // when using external standalone etcd, the cluster role cannot be "primary"
```

aniaan: updated

pkg/cluster/cluster.go
```diff
@@ -368,15 +368,23 @@ func (c *cluster) checkClusterName() error {
...
+	} else {
+		if c.opt.UseStandaloneEtcd {
+			err := c.Put(c.Layout().ClusterNameKey(), c.opt.ClusterName)
+			if err != nil {
+				return fmt.Errorf("register cluster name %s failed: %v",
+					c.opt.ClusterName, err)
+			}
+		} else {
+			return fmt.Errorf("key %s not found", c.Layout().ClusterNameKey())
+		}
```
localvar, line 387: to reduce nesting.

```suggestion
	} else if c.opt.UseStandaloneEtcd {
			err := c.Put(c.Layout().ClusterNameKey(), c.opt.ClusterName)
			if err != nil {
				return fmt.Errorf("register cluster name %s failed: %v",
					c.opt.ClusterName, err)
			}
	} else {
		return fmt.Errorf("key %s not found", c.Layout().ClusterNameKey())
```

pkg/option/option.go
```diff
@@ -109,30 +110,51 @@ type Options struct {
 
 // addClusterVars introduces cluster arguments.
 func addClusterVars(opt *Options) {
-	opt.flags.StringVar(&opt.ClusterName, "cluster-name", "eg-cluster-default-name", "Human-readable name for the new cluster, ignored while joining an existed cluster.")
-	opt.flags.StringVar(&opt.ClusterRole, "cluster-role", "primary", "Cluster role for this member (primary, secondary).")
-	opt.flags.StringVar(&opt.ClusterRequestTimeout, "cluster-request-timeout", "10s", "Timeout to handle request in the cluster.")
+	opt.flags.StringVar(&opt.ClusterName, "cluster-name", "eg-cluster-default-name",
```
localvar, line 113: per https://github.com/golang/go/wiki/CodeReviewComments#line-length:

> don't add line breaks to keep lines short when they are more readable long--for example, if they are repetitive.

please revert the newly added line breaks.

localvar (approved):

suchen-sci (approved):

PR #614, opened 2022-05-05
sync.Pool should be passed by reference

sodaRyCN:
The sync.Pool used in filter/proxy/pool.go should be passed by reference instead of by value.
use go vet command `go vet pkg/filter/proxy/*.go `, then get the warnings.

note: objects in sync.Pool that are not explicitly cleaned up before call `Put` may cause dirty data to be obtained the next time  call `Get`. I'm guessing `httpStatResultPool` might be the case, but I need experiments to prove it. EG:When the first request succeeds, `httpstat.Result`'s some properties will be set, and when the second request fails (failed to resolve the domain name, failed to establish a tcp link and etc.), some properties (NameLookup\TLSHandshake\serverDone) may use the last one. value，that may be leads to a bias in the statistics

### DIFF
--- a/pkg/filter/proxy/pool.go
+++ b/pkg/filter/proxy/pool.go
@@ -127,13 +127,13 @@ func (p *pool) status() *PoolStatus {
 	return s
 }
 
-var requestPool = sync.Pool{
+var requestPool = &sync.Pool{
 	New: func() interface{} {
 		return &request{}
 	},
 }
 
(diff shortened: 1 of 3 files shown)


localvar:
Thanks, I think this PR fixes an important bug.

localvar (approved):

suchen-sci (approved):

PR #619, opened 2022-05-10
Join rewrite target path

xxx7xxxx:


localvar:
After thinking about the issue twice, I think the current behavior is correct and propose to close this PR.

If the `rewriteTarget` is `/` and prefix is `/abc`, with this PR, we can rewrite `/abc` to `/` and `/abc/12` to `/12` by a single rule, but without this PR, we can still accomplish this by two rules.

And with this PR, if the users want to keep `/abc` as it is, but rewrite `/abc/12` to `/xyz/12`, they will have to define two rules, while they only need one rule at present.

So, I think the current behavior is correct and we don't need to change it.

localvar:
After thinking about the issue twice, I think the current behavior is correct and propose to close this PR.

If the `rewriteTarget` is `/` and prefix is `/abc`, with this PR, we can rewrite `/abc` to `/` and `/abc/12` to `/12` by a single rule, but without this PR, we can still accomplish this by two rules.

And with this PR, if the users want to keep `/abc` as it is, but rewrite `/abc/12` to `/xyz/12`, they will have to define two rules, while they only need one rule at present.

So, I think the current behavior is correct and we don't need to change it.

PR #620, opened 2022-05-11
update mqtt code for new pipeline

suchen-sci:


pkg/object/mqttproxy/broker.go
```diff
@@ -668,3 +670,12 @@ func (b *Broker) close() {
 	}
 	b.clients = nil
 }
+
+func newContext(packet packets.ControlPacket, client mqttprot.Client) *context.Context {
+	ctx := context.New(nil)
```
localvar, line 675: propose to use `tracing.NoopSpan`

pkg/protocols/mqttprot/request.go
```diff
@@ -0,0 +1,164 @@
...
+	// TODO: what header to return?
+	return nil
+}
+
+// SetPayload set the payload of the request to payload.
+func (r *Request) SetPayload(payload []byte) {
+	r.payload = payload
+	if r.packetType == PublishType {
+		r.packet.(*packets.PublishPacket).Payload = payload
+	}
```
localvar, line 143: Is it make sense to set payload if packetType is not PublishType?
If not, we can panic here.

localvar, line 143: Is it make sense to set payload if packetType is not PublishType?
If not, we can panic here.

suchen-sci: For MQTT packets, only publish packet contains payload. But for the convenience of creating customer filters, maybe allow set payload for other packet type is better...

localvar (approved):

pkg/object/mqttproxy/broker.go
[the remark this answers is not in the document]
suchen-sci: this is outdated... i already updated it.

PR #626, opened 2022-05-16
sidecar httpserver keepalive config by spec

sodaRyCN:
In sidecar mode, the httpservers of ingress and egress load keepalive-related configurations according to the mesh-controller spec(it seems that there is no other suitable object to carry this configuration).It seems that there is no other suitable object to carry this configuration. In preliminary consideration, it is not necessary to set different parameters for each worker(if this is done, the ops would be complicated). If there is, use `--label`  would be better.

### DIFF
--- a/pkg/object/meshcontroller/spec/spec.go
+++ b/pkg/object/meshcontroller/spec/spec.go
@@ -93,7 +93,7 @@ const (
 	// ServiceCanaryHeaderKey is the http header key of service canary.
 	ServiceCanaryHeaderKey = "X-Mesh-Service-Canary"
 
-	defaultKeepaliveTimeout = "60s"
+	defaultKeepAliveTimeout = "60s"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/object/meshcontroller/spec/spec.go
```diff
@@ -135,6 +137,25 @@ type (
 		Log4jConfigName           string `yaml:"log4jConfigName" jsonschema:"omitempty"`
 
 		MonitorMTLS *MonitorMTLS `yaml:"monitorMTLS" jsonschema:"omitempty"`
+		WorkerSpec  *WorkerSpec  `yaml:"workerSpec" jsonschema:"omitempty"`
+	}
+
+	// WorkerSpec is the spec of worker
+	WorkerSpec struct {
+		IngressServerSpec IngressServerSpec `yaml:"ingressServerSpec" jsonschema:"omitempty"`
+		EgressServerSpec  EgressServerSpec  `yaml:"egressServerSpec" jsonschema:"omitempty"`
```
localvar, line 146:
```suggestion
		Ingress IngressServerSpec `yaml:"ingressServerSpec" jsonschema:"omitempty"`
		Egress  EgressServerSpec  `yaml:"egressServerSpec" jsonschema:"omitempty"`
```

localvar (approved):

suchen-sci (approved):

PR #634, opened 2022-05-31
adapter scheme header

sodaRyCN:
Some front-end proxies may modify the value of X-Forwarded-Proto, resulting in the obtained schema not meeting expectations


pkg/context/httprequest.go
```diff
@@ -96,8 +102,15 @@ func (r *httpRequest) Scheme() string {
...
-	if scheme := r.std.Header.Get("X-Forwarded-Proto"); scheme != "" {
-		return scheme
+	if scheme := r.std.Header.Get(xForwardedProto); scheme != "" {
+		// note some proxy servers will append data in this format,
+		// and the last one is sent by the last proxy-server and should use for easegress's proxy
+		// header : {
+		//    some-key: "ele1, ele2, ele3"
+		// }
+		// https://github.com/spring-cloud/spring-cloud-gateway/ ProxyExchange.java#appendXForwarded()
+		schemes := strings.Split(scheme, headerValueSeparator)
```
localvar, line 112: I'm fine with the change, but I think it is better to use `,` directly than define a `headerValueSeparator`, the code would be more readable in this case.

localvar (approved):

localvar (approved):

localvar (approved):

suchen-sci (approved):

PR #649, opened 2022-06-07
fix issue 555 about request header context-length missing

### LINKED ISSUE #555: The `content-length` header in the request is always replaced with `transfer-encoding: chunked`
**Describe the bug**
I used Easegress as a reverse proxy server. When I send requests through Easegress to upstream servers, the `content-length` header in the request is always replaced with `transfer-encoding: chunked` header.

This will cause some servers didn't work. For example, the spring cloud gateway will work incorrectly when a `DELETE` HTTP request carries a `transfer-encoding: chunked` header.

**To Reproduce**
1. Echo Server setup. This server will return all the things it received in the request.
```shell
docker run --rm -d -p 23888:80  ealen/echo-server
```
2. Nginx setup
```shell
echo -e "
server {
    listen       80;
    server_name  localhost;

    location / {
        root   /usr/share/nginx/html;
        index  index.html index.htm;
    }

    location /echo {
        proxy_pass   http://host.docker.internal:23888;
    }
}
" > /tmp/default.conf
docker run --rm -d -p 23899:80 -v /tmp/default.conf:/etc/nginx/conf.d/default.conf --name ng-test nginx
```
3. Easegress setup
```shell
docker run --rm -d -p 2381:2381 -p 23898:23898 --name eg-test megaease/easegress:v1.5.0 --api-addr 0.0.0.0:2381

curl -v -X POST http://localhost:2381/apis/v1/objects \
-H 'Content-Type: application/yaml; charset=utf-8' \
--data-raw '
name: pipeline-demo
kind: HTTPPipeline
flow:
  - filter: proxy
filters:
  - name: proxy
    kind: Proxy
    mainPool:
      servers:
      - url: http://host.docker.internal:23888
      loadBalance:
        policy: roundRobin
'

curl -v -X POST http://localhost:2381/apis/v1/objects \
-H 'Content-Type: application/yaml; charset=utf-8' \
--data-raw '
kind: HTTPServer
name: server-demo
port: 23898
keepAlive: true
https: false
rules:
  - paths:
    - pathPrefix: /echo
      backend: pipeline-demo
'
```
4. Sent requests
```shell
# request echo server
curl -v -X POST 'http://localhost:23888/echo/test' -H 'content-type: application/json' -H 'content-length: 7' --data-raw '{"a":1}' |  jq .request.headers
# echo server response
{
  "host": "localhost:23888",
  "user-agent": "curl/7.79.1",
  "accept": "*/*",
  "content-type": "application/json",
  "content-length": "7"
}

# request nginx
curl -v -X POST 'http://localhost:23899/echo/test' -H 'content-type: application/json' -H 'content-length: 7' --data-raw '{"a":1}' |  jq .request.headers
# nginx server response
{
  "host": "host.docker.internal:23888",
  "connection": "close",
  "content-length": "7",
  "user-agent": "curl/7.79.1",
  "accept": "*/*",
  "content-type": "application/json"
}

# request Easegress
curl -v -X POST 'http://localhost:23898/echo/test' -H 'content-type: application/json' -H 'content-length: 7' --data-raw '{"a":1}' |  jq .request.headers
# Easegress server response
{
  "host": "host.docker.internal:23888",
  "user-agent": "curl/7.79.1",
  "transfer-encoding": "chunked",
  "accept": "*/*",
  "content-type": "application/json",
  "accept-encoding": "gzip"
}
```
> `PUT` and `DELETE` methods will return a similar response.

**Expected behavior**

The `content-length: 7` header in the HTTP request should be kept.

**Version**
The version number of Easegress is v1.5.0.

**OS and Hardware**
 - OS: MacOS 12.3
 - CPU: 2.6 GHz 6-Core Intel Core i7
 - Memory: 32GB

suchen-sci:
Golang `http.NewRequestWithContext` will set `http.Request.ContentLength` when body is `bytes.Buffer`, `bytes.Reader` or `strings.Reader`. If `http.Request.ContentLength` is not send, Golang std lib will delete Header of `Context-Length` when sending request. Fix #555

### DIFF
--- a/pkg/filter/proxy/proxy.go
+++ b/pkg/filter/proxy/proxy.go
@@ -352,13 +352,13 @@ func (b *Proxy) fallbackForCodes(ctx context.HTTPContext) bool {
 // redirects can replay the body), and Body is set to NoBody if the
 // ContentLength is 0.
 //
-// So in this way, http.Request.ContextLength will be 0, and when http.Client send this request, it will delete
-// "Context-Length" key in header. We solve this problem by set http.Request.ContextLength equal to
-// http.Request.Header["Context-Length"] (if it is presented).
+// So in this way, http.Request.ContentLength will be 0, and when http.Client send this request, it will delete
+// "Content-Length" key in header. We solve this problem by set http.Request.ContentLength equal to
+// http.Request.Header["Content-Length"] (if it is presented).
 // Reading all context.Request().Body() and create new request with bytes.NewReader is another way, but it may cause performance loss.
 //
-// It is important "Context-Length" in Header is equal to length of Body. In easegress, when a filter change Request.Body,
-// it will delete the header of "Context-Length". So, you should not worry about this when using our filters.
+// It is important that "Content-Length" in the Header is equal to the length of the Body. In easegress, when a filter change Request.Body,
+// it will delete the header of "Content-Length". So, you should not worry about this when using our filters.
 // But for customer filters, developer should make sure to delete or set "Context-Length" value in header when change Request.Body.
 func (b *Proxy) Handle(ctx context.HTTPContext) (result string) {
 	result = b.handle(ctx)
--- a/pkg/filter/proxy/proxy.go
+++ b/pkg/filter/proxy/proxy.go
@@ -343,6 +343,23 @@ func (b *Proxy) fallbackForCodes(ctx context.HTTPContext) bool {
 }
 
 // Handle handles HTTPContext.
+// When we create new request for backend, we call http.NewRequestWithContext method and use context.Request().Body() as body.
+// Based on golang std lib comments:
+// https://github.com/golang/go/blob/95b68e1e02fa713719f02f6c59fb1532bd05e824/src/net/http/request.go#L856-L860
+// If body is of type *bytes.Buffer, *bytes.Reader, or
+// *strings.Reader, the returned request's ContentLength is set to its
+// exact value (instead of -1), GetBody is populated (so 307 and 308
+// redirects can replay the body), and Body is set to NoBody if the
+// ContentLength is 0.
+//
+// So in this way, http.Request.ContextLength will be 0, and when http.Client send this request, it will delete
+// "Context-Length" key in header. We solve this problem by set http.Request.ContextLength equal to
+// http.Request.Header["Context-Length"] (if it is presented).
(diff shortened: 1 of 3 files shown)


pkg/filter/proxy/request.go
```diff
@@ -80,6 +82,17 @@ func (p *pool) newRequest(
 		stdr.Host = r.Host()
 	}
 
+	// golang http.NewRequestWithContext will set ContentLength when body is bytes.Buffer,
```
xxx7xxxx, line 85: Cool, but if you want to comment on its behavior, we had better use the official reference.

zhao-kun: Agreed, I'm also curious about what's said in the Golang office document.

suchen-sci: updated!

suchen-sci:
updated!

zhao-kun (approved):

pkg/filter/proxy/proxy.go
```diff
@@ -343,6 +343,23 @@ func (b *Proxy) fallbackForCodes(ctx context.HTTPContext) bool {
...
+// ContentLength is 0.
+//
+// So in this way, http.Request.ContextLength will be 0, and when http.Client send this request, it will delete
+// "Context-Length" key in header. We solve this problem by set http.Request.ContextLength equal to
+// http.Request.Header["Context-Length"] (if it is presented).
+// Reading all context.Request().Body() and create new request with bytes.NewReader is another way, but it may cause performance loss.
+//
+// It is important "Context-Length" in Header is equal to length of Body. In easegress, when a filter change Request.Body,
+// it will delete the header of "Context-Length". So, you should not worry about this when using our filters.
+// But for customer filters, developer should make sure to delete or set "Context-Length" value in header when change Request.Body.
```
localvar, line 362:
```suggestion
// So in this way, http.Request.ContentLength will be 0, and when http.Client send this request, it will delete
// "Content-Length" key in header. We solve this problem by set http.Request.ContentLength equal to
// http.Request.Header["Content-Length"] (if it is presented).
// Reading all context.Request().Body() and create new request with bytes.NewReader is another way, but it may cause performance loss.
//
// It is important that "Content-Length" in the Header is equal to the length of the Body. In easegress, when a filter change Request.Body,
// it will delete the header of "Content-Length". So, you should not worry about this when using our filters.
// But for customer filters, developer should make sure to delete or set "Context-Length" value in header when change Request.Body.
```

localvar (approved):

PR #651, opened 2022-06-07
udpate k8s ingress controller route policy

### LINKED ISSUE #633: IngressController can not handle Ingress rule priority correctly
**Describe the bug**
IngressController can not handle Ingress rule priority correctly

**To Reproduce**
create two Ingress in k8s like:
```yaml
...
  rules:
  - http:
      paths:
      - backend:
          service:
            name: service1
        path: /v1
        pathType: Prefix
```
and
```yaml
...
  rules:
  - host: www.megaease.com
    http:
      paths:
      - backend:
          service:
            name: service2
        path: /v1
        pathType: Prefix
```
request to URL `http://www.megaease.com/v1/hello` is forwarded to service1.

**Expected behavior**
the request is forwarded to service2

**Version**
v1.5.2

suchen-sci:
fix #633

### DIFF
--- a/pkg/object/ingresscontroller/translator.go
+++ b/pkg/object/ingresscontroller/translator.go
@@ -62,6 +62,30 @@ type (
 	}
 )
 
+func (b *httpServerSpecBuilder) sortRules() {
+	hostRules := []*httpserver.Rule{}
+	hostRegRules := []*httpserver.Rule{}
+	noHostRules := []*httpserver.Rule{}
+	for _, rule := range b.Rules {
+		if rule.Host != "" {
+			hostRules = append(hostRules, rule)
+		} else if rule.HostRegexp != "" {
+			hostRegRules = append(hostRegRules, rule)
+		} else {
+			noHostRules = append(noHostRules, rule)
+		}
+	}
+	sort.Slice(hostRules, func(i, j int) bool {
+		return hostRules[i].Host < hostRules[j].Host
+	})
+	sort.Slice(hostRegRules, func(i, j int) bool {
+		return hostRegRules[i].HostRegexp < hostRegRules[j].HostRegexp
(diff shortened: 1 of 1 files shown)


pkg/object/ingresscontroller/translator.go
```diff
@@ -366,11 +370,23 @@ func (st *specTranslator) translateIngressRules(b *httpServerSpecBuilder, ingres
...
-			b.Rules = append(b.Rules, r)
+			if len(rule.Host) > 0 {
+				if rule.Host[0] == '*' {
+					wildcardHostRules = append(wildcardHostRules, r)
+				} else {
+					hostRules = append(hostRules, r)
+				}
+			} else {
+				noHostRules = append(noHostRules, r)
+			}
```
localvar, line 381: the new rules have not been added to `b.Rules` in this case, that's the `existingRule` may have a wrong value.

pkg/object/ingresscontroller/translator.go
```diff
@@ -317,6 +317,10 @@ func (st *specTranslator) translateTLSConfig(b *httpServerSpecBuilder, ingresses
 }
 
 func (st *specTranslator) translateIngressRules(b *httpServerSpecBuilder, ingress *apinetv1.Ingress) {
+	hostRules := []*httpserver.Rule{}
+	wildcardHostRules := []*httpserver.Rule{}
+	noHostRules := []*httpserver.Rule{}
```
localvar, line 322: there could be more than one ingress, we need to consider this case also.

suchen-sci: updated!

localvar (approved):

zhao-kun (approved):
LGTM

PR #656, opened 2022-06-15
fix: cannot start without specify initial cluster

localvar:


pkg/option/option.go
```diff
@@ -238,6 +234,9 @@ func (opt *Options) Parse() (string, error) {
 	if opt.UseStandaloneEtcd {
 		opt.ClusterRole = "secondary" // when using external standalone etcd, the cluster role cannot be "primary"
 	}
+	if len(opt.Cluster.InitialCluster) == 0 {
+		opt.Cluster.InitialCluster = map[string]string{"eg-default-name": "http://localhost:2380"}
+	}
```
suchen-sci, line 239:
```suggestion
	if opt.ClusterRole == "primary" && len(opt.Cluster.InitialCluster) == 0 {
		opt.Cluster.InitialCluster = map[string]string{opt.Name: opt.InitialAdvertisePeerURLs[0]} 
	}
```

PR #667, opened 2022-06-24
update according to review comments

localvar:
* add version to spec
* change status code to 503 when no response
* add more debug log
* fix typo

### DIFF
--- a/pkg/supervisor/spec.go
+++ b/pkg/supervisor/spec.go
@@ -25,6 +25,7 @@ import (
 	"github.com/megaease/easegress/pkg/v"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/supervisor/spec.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/megaease/easegress/pkg/v"
 )
 
+const DefaultSpecVersion = "easegress.megaease.com/v2"
```
suchen-sci, line 28: `DefaultSpecVersion` needs comments...

localvar: done.

suchen-sci (approved):

PR #686, opened 2022-07-04
update README

localvar:


README.md
```diff
@@ -131,11 +133,17 @@ For full list, see [Cookbook](./doc/README.md#1-cookbook--how-to-guide).
...
+demonstrate the API orchestration feature by including more components in the
+configuration, we will also show the essential concepts and operations of
+Easegress.
 
 ### Setting up Easegress
 
-We can download the latest or history binaries from the [release page](https://github.com/megaease/easegress/releases). The following shell script will do:
+We can download the latest or history binaries from the
+[release page](https://github.com/megaease/easegress/releases). The following
+shell script will:
```
xxx7xxxx, line 146:
```suggestion
shell script will be:
```

localvar: I don't think we need to add the word `be`, as the full sentences are:
```
The following shell script will download and extract the latest binaries to `./easegress` folder  
```
and
```
The following shell script will install the Easegress Systemd service.
```

xxx7xxxx: oops, you're right

xxx7xxxx (approved):

suchen-sci (approved):

PR #699, opened 2022-07-14
fix httpserver header all match

sodaRyCN:


pkg/object/httpserver/mux.go
```diff
@@ -290,8 +290,8 @@ func (mp *MuxPath) matchHeaders(r *httpprot.Request) bool {
 	if mp.matchAllHeader {
 		for _, h := range mp.headers {
 			v := r.HTTPHeader().Get(h.Key)
-			if !stringtool.StrInSlice(v, h.Values) {
-				return false
+			if stringtool.StrInSlice(v, h.Values) {
+				continue
```
localvar, line 294: The change will return `true` if no values are matched and no regexp is configured.

```suggestion
			if len(h.Values) > 0 && !stringtool.StrInSlice(v, h.Values) {
				return false
```

localvar (approved):

suchen-sci (approved):

Issue #713, opened 2022-07-23
The Proxy filter lost all response headers set by other filters(eg: CORSAdaptor) before it


jthann:
**Describe the bug**
In Easegress v2.0.0，the Proxy filter will lost all response headers set by other filters. For example if using CORSAdaptor before Proxy filter，for simple cors request，it will set Vary Access-Control-Allow-Origin Access-Control-Allow-Credentials headers.  All these headers will lost after Proxy filter handling. This will cause simple cors request cannot work properly when api and web page from different domain.

**To Reproduce**
Using following configuration under Easegress v2.0.0：(```echo-backend``` service providing 9095 port is from [Easegress example](https://github.com/megaease/easegress/tree/main/example/backend-service/echo)
```frontend``` just use ResponseBuilder  to build a html page containing javascript that send Cors request to api domain)

Make sure adding following configuration to```/etc/hosts``` file:
```
127.0.0.1 frontend.work api.backend.work 
```

```yaml
kind: HTTPServer
name: http
port: 80
https: false
rules:
  - host: frontend.work
    paths:
      - PathPrefix: /page
        backend: frontend
  - host: api.backend.work
    paths:
      - pathPrefix: /api
        backend: echo-backend
---
name: frontend
kind: Pipeline
flow:
  - filter: respBuilder
  - filter: rsp
filters:
  - name: respBuilder
    kind: ResponseBuilder
    template: |
      statusCode: 200
  - name: rsp
    kind: ResponseAdaptor
    body: >+
      <!DOCTYPE html>
      <html>   
      <body>
      <h1>This is Frontend Page!</h1>
      <script>
          fetch('http://api.backend.work/api', {
              mode: "cors",
              method: "POST",
              credentials: "include",
              cache: 'no-cache',              
          }).then(data => console.log(data))
            .catch((error) => {console.log('Error', error)});
      </script>
      </body>
      </html>
---
name: echo-backend
kind: Pipeline
flow:
  - filter: cors-filter
    jumpIf: { preflighted: END }
  - filter: proxy
filters:
  - name: cors-filter
    kind: CORSAdaptor
    supportCORSRequest: true
    allowedOrigins: ["http://frontend.work"]
    allowCredentials: true
    allowedMethods: ["GET","POST"]
  - name: proxy
    kind: Proxy
    pools:
      - servers:
          - url: http://127.0.0.1:9095
        loadBalance:
          policy: roundRobin
```

[...]

PR #721, opened 2022-07-25
Update pool.go

jthann:
fix [issue 713 bug](https://github.com/megaease/easegress/issues/713)


localvar:
Thanks for the PR, but we think #713 is a little bit complex because v2.0 allows multiple requests/responses, and the response returned by the Proxy filter may not be the one returned to the client, we are looking for a solution, and welcome your contribution.

PR #748, opened 2022-08-05
add v1 APIs back for compatibility

localvar:


xxx7xxxx (approved):

xxx7xxxx:
unit test failure on Windows？

localvar:
> unit test failure on Windows？

it is an error caused by etcd and the low windows performance, we can do nothing to it, disappeared after a re-run.

PR #750, opened 2022-08-09
Release 2.1.0

localvar:


suchen-sci (approved):

pkg/filters/builder/builder.go
```diff
@@ -76,7 +76,7 @@ func (b *Builder) build(data map[string]interface{}, v interface{}) error {
 		return err
 	}
 
-	return codectool.Decode(&result, v)
+	return yaml.Unmarshal(result.Bytes(), v)
```
xxx7xxxx, line 79: It's consistent to use `codecool.UnmarshalYAML` in case of replacing the vendor in one time and one place.

pkg/protocols/httpprot/request.go
```diff
@@ -406,17 +406,17 @@ func (r *builderRequest) YAMLBody() (interface{}, error) {
 
 // requestInfo stores the information of a request.
 type requestInfo struct {
-	Method   string              `json:"method" jsonschema:"omitempty"`
-	URL      string              `json:"url" jsonschema:"omitempty"`
-	Headers  map[string][]string `json:"headers" jsonschema:"omitempty"`
-	Body     string              `json:"body" jsonschema:"omitempty"`
-	FormData map[string]field    `json:"formData" jsonschema:"omitempty"`
+	Method   string              `json:"method" yaml:"method" jsonschema:"omitempty"`
```
xxx7xxxx, line 409: The megaease dedicated yaml vendor will use json tag if there is no yaml tag.

pkg/filters/builder/builder.go
localvar: ok, I will switch to that function.

pkg/protocols/httpprot/request.go
localvar: I think making it explicit would be better because this is a "YAML in YAML" case.

xxx7xxxx: Sure, would you mind adding a line of comment about it

xxx7xxxx (approved):

suchen-sci:
no release note and change log...

suchen-sci (approved):

PR #757, opened 2022-08-12
Set default failure codes 5xx

### LINKED ISSUE #689: Support default failureCodes

EaseMesh will use the circuitbreaker to protect the service, but the `failureCodes` are not the sub-config of it.

When `failureCodes` of proxy pool are empty, it should be 5xx[1] by default, which is the natural range for circuitbreaker. Besides Mesh scenario, the plain usage of Easegress will also benefit from the default behavior since it's annoying to manually configure from 501 to 511.

[1] https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_server_errors

xxx7xxxx:
Fix #689

### DIFF
--- a/doc/reference/filters.md
+++ b/doc/reference/filters.md
@@ -100,7 +100,7 @@ kind: Proxy
 name: proxy-example-1
 pools:
 - servers:
-  - url: http://127.0.0.1:9095 
+  - url: http://127.0.0.1:9095
 ```
 
 Pool without `filter` is considered the main pool, other pools with `filter`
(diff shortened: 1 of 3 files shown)


localvar:
please update the document to mention the default values: https://github.com/megaease/easegress/blob/main/doc/reference/filters.md#proxyserverpoolspec

suchen-sci:
should we keep a way to not set `failureCode`? since `failureCode` is a result of proxy and may influence the execution of pipeline...

xxx7xxxx:
> please update the document to mention the default values

Document updated.

> should we keep a way to not set failureCode? since failureCode is a result of proxy and may influence the execution of pipeline...

Yes, it's very annoying to manually set it one by one e.g [500, 501, 502...]. Because most time 5xx standing for server error would be a failure indication for circuit breaker and retry. [1]

[1] https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_server_errors

localvar (approved):

suchen-sci (approved):

PR #771, opened 2022-08-31
add debug log for matched pipeline

localvar:


pkg/object/httpserver/mux.go
```diff
@@ -494,10 +494,11 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 
 	handler, ok := mi.muxMapper.GetHandler(route.path.backend)
 	if !ok {
-		logger.Debugf("%s: backend %q not found", mi.superSpec.Name(), route.path.backend)
+		logger.Debugf("%s: backend(Pipeline) %q not found", mi.superSpec.Name(), route.path.backend)
 		buildFailureResponse(ctx, http.StatusServiceUnavailable)
 		return
 	}
+	logger.Debugf("%s: the matched backend(Pipeline) is %s", mi.superSpec.Name(), route.path.backend)
```
xxx7xxxx, line 501: More info about the matched request?

jthann: > More info about the matched request?

Yes，we need client request URI，matched configured path，matched path backend

localvar: there could be many fields in the matched path, including headers and methods, which makes it not suitable for logging.
I think the request URI and matched backend is enough to help find the wrong rule, so I prefer not to include the matched path.

any comments?

xxx7xxxx: Logging the method would be a good cost-effective addition

suchen-sci (approved):

pkg/object/httpserver/mux.go
localvar: method added.

xxx7xxxx (approved):

PR #772, opened 2022-08-31
add pipeline name in serializeStats tag

jthann:
Currently，the default console log only prints filter name，It is very hard to troubleshoot URI path route problem，especially for following configure example，all filter names are same，so we need to print pipeline name to distinct which backend the request URI is routed.
```yaml
kind: HTTPServer
name: http
port: 8080
rules:
  - paths:
      - pathRegexp: /a
        backend: a
      - pathRegexp: /b
        backend: b
---
name: a
kind: Pipeline
flow:
  - filter: proxy
filters:
  - name: proxy
    kind: Proxy
    pools:
      - servers:
          - url:
---

name: b
kind: Pipeline
flow:
  - filter: proxy
filters:
  - name: proxy
    kind: Proxy
    pools:
      - servers:
          - url:
```

### DIFF
--- a/pkg/object/pipeline/pipeline.go
+++ b/pkg/object/pipeline/pipeline.go
@@ -174,14 +174,15 @@ func (s *Spec) Validate() (err error) {
 	return nil
 }
 
-func serializeStats(stats []FilterStat) string {
+func (p *Pipeline) serializeStats(stats []FilterStat) string {
 	if len(stats) == 0 {
 		return "pipeline: <empty>"
 	}
 
 	var sb strings.Builder
-	sb.WriteString("pipeline: ")
-
+	sb.WriteString("pipeline(")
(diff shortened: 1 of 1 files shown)


pkg/object/pipeline/pipeline.go
```diff
@@ -174,14 +174,15 @@ func (s *Spec) Validate() (err error) {
...
 	if len(stats) == 0 {
 		return "pipeline: <empty>"
 	}
 
 	var sb strings.Builder
-	sb.WriteString("pipeline: ")
-
+	sb.WriteString("pipeline(")
+	sb.WriteString(p.superSpec.Name())
+	sb.WriteString("): ")
```
localvar, line 185: please also update L179 to include the pipeline name.

localvar (approved):

suchen-sci (approved):

PR #793, opened 2022-09-12
support grpc protocol proxy

sodaRyCN:
企业微信截图_16633259236814
This PR is for the proxy that easegress can support the gRPC protocol.
The simple process is shown in the figure above, which can forward the traffic on the gRPC client side to the downstream gRPC server according to the configured pipeline rules. It should be reminded that gRPC stream is a full-duplex protocol, so the traffic can also be pushed from the server side to the client side, the above picture is more to show the process of proxy, but not to emphasise the characteristics of full-duplex.
The main modification is the addition of the following modules:
1. pkg/object/grpcserver
- grpcserver: implements the `TrafficGate` interface, managed by the Supervisor. It manages the life cycle of the runtime
- runtime: wrapper the `grpc.server` listener and listen for events to manage the server's life cycle, such as restart/shutdown/hot update, etc.
- mux: implements the `grpc.UnknownServiceHandler` interface to process all traffic entering the `runtime`, and implements the same route capability as object/httpserver/mux
2. pkg/filter/grpc
- proxy: implements the filter interface
- pool: the actual traffic forwarding is completed. Like /filter/proxy/pool, it supports docking with the service registry centre and statically configures the downstream instance address. You can configure whether to use the connection pool: some gRPC servers manage physical connections by themselves, and they do not consider gRPC gateway scenarios, such as Nacos: when the actual client goes offline, the server will detect and actively close the server's Connection. because gRPC is based on HTT2, there could be a connection from the same easegress used by multiple real clients to the real server. In this case, the server actively closes the connection, which will cause other normal clients to be affected.
- load balancer: provides load balancers such as round-robin, random, IPHash, etc., and adds a forward proxy balancer. Because `gRPC.Server` does not support tunnelling, it adopts the method of parsing headers
3. pkg/protocols/grpcprot
- request: implements `protocols.Request`, wrapper `grpc.ServerStream`
- response:implements `protocols.Response`, wrapper `status.Status`
- header: wrapper `metadata.md`, used by `grpc.InComingContext/OutgoingContext`
- trailer: sent by the gRPC server side at the end of the data when stream closed. It has the same structure and method as the Header, so it is used as an alias for the `Header`
4. pkg/util/connectionpool
- pool: Defines the connection pool interface
- grpc/codec: implements `grpc.Codec` instead of `encoding.Codec`, because `encoding.Codec` would change the header: use `encoding.Codec`, then header's content-type=application/grpc -> content-type=application/grpc+encoding.Codec.Name()
- grpc/pool: implements pool interface. According to the target address segmented lock, the connection of each address is designed according to the Producer-Consumer pattern


pkg/protocols/grpcprot/request.go
```diff
@@ -0,0 +1,306 @@
...
+	r.ctx = metadata.NewIncomingContext(r.ctx, r.header.md)
+	r.peer = &peer.Peer{Addr: &Addr{}}
+
+	if peerInfo, ok := peer.FromContext(r.ctx); ok {
+		r.peer.Addr.(*Addr).setAddr(peerInfo.Addr.String())
+		r.peer.AuthInfo = peerInfo.AuthInfo
+	}
+	r.ctx = peer.NewContext(r.ctx, r.peer)
+
+	return r
```
localvar, line 134: these lines of code are duplicated with the above function, please do a refactor.

pkg/protocols/grpcprot/response.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+// Response wrapper status.Status
+type Response struct {
+	// instead status.Status
+	// nil or from codes.OK means success, otherwise means fail
+	*status.Status
+	header  *Header
+	trailer *Trailer
+	headMtx sync.Mutex
```
localvar, line 36: it seems this field is never used.

pkg/util/connectionpool/grpc/pool.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+
+	logger.Infof("created grpc connection pool with factory and spec %+v", spec)
+	return &Pool{
+		spec:       spec,
+		connsMutex: &sync.Mutex{},
+		conns:      map[string]*segmentConn{},
+		factory:    factory,
+	}
+}
```
localvar, line 146: please refactor to avoid duplication.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,519 @@
...
+	// req is just request data
+	req *grpcprot.Request
+	// resp is just response data
+	resp *grpcprot.Response
+}
+
+var (
+	desc = &grpc.StreamDesc{
+		// We need to assume that the users of the streamHandler will want to use both.
+		// in test, client and server anyone or both not streamHandler, it's work well too
```
localvar, line 84: please revise the words, it is really confusing.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,519 @@
...
+			if stringtool.StrInSlice(tag, instance.Tags) {
+				servers = append(servers, &Server{
+					URL:    instance.URL(),
+					Tags:   instance.Tags,
+					Weight: instance.Weight,
+				})
+				break
+			}
+		}
+	}
```
localvar, line 232: we have fixed a bug here in the HTTP version, please sync the code.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,519 @@
...
+
+		spCtx.LazyAddTag(func() string {
+			return fmt.Sprintf("status code: %d", err.(serverPoolError).Code())
+		})
+
+		return err
+	}
+
+	// resilience wrappers, note that it is impossible to retry a stream
+	// request as its body can only be read once.
```
localvar, line 292: this comment should be removed.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,519 @@
...
+	svr := lb.ChooseServer(spCtx.req)
+	if f, ok := lb.(ReusableServerLB); ok {
+		defer f.ReturnServer(svr)
+	}
+
+	// if there's no available server.
+	if svr == nil {
+		logger.Debugf("%s: no available server", sp.name)
+		return serverPoolError{status.New(codes.InvalidArgument, "no available server"), resultClientError}
+	}
```
localvar, line 335:
```suggestion
	// if there's no available server.
	if svr == nil {
		logger.Debugf("%s: no available server", sp.name)
		return serverPoolError{status.New(codes.InvalidArgument, "no available server"), resultClientError}
	}

	if f, ok := lb.(ReusableServerLB); ok {
		defer f.ReturnServer(svr)
	}
```

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,519 @@
...
+	SpanName             string              `yaml:"spanName" jsonschema:"omitempty"`
+	Filter               *RequestMatcherSpec `yaml:"filter" jsonschema:"omitempty"`
+	ServerTags           []string            `yaml:"serverTags" jsonschema:"omitempty,uniqueItems=true"`
+	Servers              []*Server           `yaml:"servers" jsonschema:"omitempty"`
+	ServiceRegistry      string              `yaml:"serviceRegistry" jsonschema:"omitempty"`
+	ServiceName          string              `yaml:"serviceName" jsonschema:"omitempty"`
+	LoadBalance          *LoadBalanceSpec    `yaml:"loadBalance" jsonschema:"omitempty"`
+	Timeout              string              `yaml:"timeout" jsonschema:"omitempty,format=duration"`
+	CircuitBreakerPolicy string              `yaml:"circuitBreakerPolicy" jsonschema:"omitempty"`
+	FailureCodes         []int               `yaml:"failureCodes" jsonschema:"omitempty"`
```
localvar, line 116: this field is useless.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,519 @@
...
+)
+
+// ServerPool defines a server pool.
+type ServerPool struct {
+	proxy        *Proxy
+	spec         *ServerPoolSpec
+	done         chan struct{}
+	wg           sync.WaitGroup
+	name         string
+	failureCodes map[int]struct{}
```
localvar, line 97: this field is useless.

pkg/filters/grpcproxy/proxy.go
```diff
@@ -0,0 +1,309 @@
...
+)
+
+const (
+	// Kind is the kind of Proxy.
+	Kind = "GRPCProxy"
+
+	resultInternalError = "internalError"
+	resultClientError   = "clientError"
+	resultServerError   = "serverError"
+	resultFailureCode   = "failureCode"
```
localvar, line 45: useless result.

pkg/filters/grpcproxy/proxy.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// help gc
+	if !p.spec.UseConnectionPool {
+		p.conns = nil
+	} else {
+		p.pool.Close()
+		p.pool = nil
+	}
+	p.closeEvent <- struct{}{}
+	close(p.closeEvent)
```
localvar, line 285: for this purpose, we can simply use an un-buffered channel and close it directly.

pkg/util/connectionpool/grpc/pool.go
```diff
@@ -0,0 +1,311 @@
...
+
+func (m *Pool) Get(addr string) (interface{}, error) {
+	timeout, cancelFunc := context.WithTimeout(context.Background(), m.spec.BorrowTimeout)
+	defer cancelFunc()
+	segment := m.conns[addr]
+	if segment == nil {
+		m.connsMutex.Lock()
+		if m.conns[addr] == nil {
```
localvar, line 194: the pattern of the `connsMutex` usage is wrong, either we don't need it, or there's a race condition.

pkg/util/connectionpool/grpc/pool.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+	}
+}
+
+func (m *Pool) Close() {
+	if m.isDone {
+		return
+	}
+	m.connsMutex.Lock()
+	defer m.connsMutex.Unlock()
```
localvar, line 244: in most cases, we should not use a lock in the cleanup functions, it is often useless or error prone.

jiekun:
Would you mind linking an issue or concluding your change on the PR description rather than leaving it empty?

sodaRyCN:
> Would you mind linking an issue or concluding your change on the PR description rather than leaving it empty?

yeah, I will supply the description of my change as soon as possible.

doc/cookbook/grpc-server.md
```diff
@@ -0,0 +1,233 @@
+# API Aggregation
```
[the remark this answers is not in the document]
sodaRyCN: I forgot to commit the doc and now fixed it. If the content is not good yet, I will exclude the doc from this PR.

pkg/protocols/grpcprot/header.go
```diff
@@ -0,0 +1,109 @@
...
+	h.md.Delete(key)
+}
+
+// MergeHeader all the provided metadata will be merged
+func (h *Header) MergeHeader(header *Header) {
+	if header == nil {
+		return
+	}
+	h.lock.Lock()
+	defer h.lock.Unlock()
```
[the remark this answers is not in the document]
sodaRyCN: I think that type `header` is not thread-safe without the lock, so we cloud discuss it further.

pkg/protocols/grpcprot/trailer.go
```diff
@@ -0,0 +1,92 @@
+/*
```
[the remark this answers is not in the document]
sodaRyCN: I do a reactor: type `trailer` is an alias for `header`.
`trailer` and `header` are two independent concepts in [gRPC protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md), so to avoid confusion, I recommend using two separate types.

pkg/protocols/grpcprot/request.go
```diff
@@ -0,0 +1,306 @@
...
+
+	r.sts = &serverTransportStream{ServerTransportStream: grpc.ServerTransportStreamFromContext(r.ctx)}
+	// grpc's method equals request.path in http standard lib
+	if method, ok := grpc.Method(r.ctx); ok {
+		r.sts.path = method
+	}
+	r.ctx = grpc.NewContextWithServerTransportStream(r.ctx, r.sts)
+	md, ok := metadata.FromIncomingContext(r.ctx)
+	if !ok {
+		logger.Infof("couldn't get headers' md from grpc context")
```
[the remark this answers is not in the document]
sodaRyCN: we cloud discuss it further.

pkg/protocols/grpcprot/trailer.go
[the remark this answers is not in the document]
localvar: `trailer` is a concept of `http`, you may check the standard Go library, it is using `header` as `trailer`, we should follow it:

https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/net/http/request.go;l=277

doc/cookbook/grpc-server.md
```diff
@@ -0,0 +1,52 @@
...
+flow:
+  - filter: grpcforwardproxy
+    jumpIf: {}
+kind: Pipeline
+name: pipeline-grpc-forward
+```
+
+If you don't want to use connection pool , then set `useConnectionPool` to false or not specify. by the way,
+parameters such as `initConnNum`,`maxConnsPerHost`,`connectTimeout`,`borrowTimeout` are valid when `useConnectionPool` is
+true, otherwise, these parameters would be ignored.
```
haoel, line 45:
```suggestion
If you don't want to use the connection pool, then set `useConnectionPool` to false or not specified. by the way,parameters such as `initConnNum`, `maxConnsPerHost`, `connectTimeout`, `borrowTimeout` are valid when `useConnectionPool` is true, otherwise, these parameters would be ignored.
```

doc/cookbook/grpc-server.md
```diff
@@ -0,0 +1,52 @@
...
+
+### When Not to Use Connection Pool
+Some gRPC servers manage physical connections by themselves, and they are not designed with gRPC gateway scenarios, such 
+as Nacos: when the actual client goes offline, the server will detect and actively close the server's Connection. 
+because gRPC is based on HTT2, it has the characteristics of request and response multiplexing, so there may be a 
+connection from the same easegress used by multiple real clients to the real server.In this case, the server actively 
+closes the connection, which will cause other normal clients to be affected.
```
haoel, line 52:
```suggestion
Some gRPC servers manage physical connections by themselves, and they are not designed with gRPC gateway scenarios, such as Nacos: when the actual client goes offline, the server will detect and actively close the server's Connection. because gRPC is based on HTT2, it has the characteristics of request and response multiplexing, so there may be a connection from the same easegress used by multiple real clients to the real server. In this case, the server actively closes the connection, which will cause other normal clients to be affected.
```

doc/cookbook/grpc-server.md
```diff
@@ -0,0 +1,52 @@
...
+    - [Similar to Service Proxy](#similar-to-service-proxy)
+    - [Configure Whether to Use Connection Pool](#configure whether to use connection pool)
+       - [When Not to Use Connection Pool](#When-not-to-use-connection-pool)
+
+Easegress gRPC servers as a forward or reverse proxy base on gRPC protocol. 
+
+
+## Similar to Service Proxy
+Its design and implementation largely refer to the original HTTP Server, so it is strongly recommended that you read 
+the [Service-Proxy documentation](./service-proxy.md) first.
```
haoel, line 13:
```suggestion
Its design and implementation largely refer to the original HTTP Server, so it is strongly recommended that you read the [Service-Proxy documentation](./service-proxy.md) first.
```

pkg/option/option.go
```diff
@@ -80,7 +80,7 @@ type Options struct {
 	ClusterRequestTimeout string         `yaml:"cluster-request-timeout"`
 	Cluster               ClusterOptions `yaml:"cluster"`
 
-	// Path.
+	// FullMethod.
```
localvar, line 83: please be careful when `Replace In All Files`, many code changes in this commit are wrong.

sodaRyCN: yeah, I will be careful.
LOL, I used `rename(shift + F6)` instead of `replace`, but it looks like  `IntelliJ IDEA` has some problems judging references.

pkg/protocols/grpcprot/request.go
```diff
@@ -0,0 +1,298 @@
...
+
+// GetFirstInHeader returns the first value for a given key.
+// k is converted to lowercase before searching in md.
+func (r *Request) GetFirstInHeader(k string) string {
+	v := r.header.md.Get(k)
+	if v == nil {
+		return ""
+	}
+	return v[0]
+}
```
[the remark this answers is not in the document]
sodaRyCN: the value structure in the grpc header is []string, so sometimes it will be necessary to obtain the first one, for reference 
 [scg](https://github.com/spring-cloud/spring-cloud-gateway/blob/fdd348a1045db182661e90eaf6fd351e5f2b21de/spring-cloud-gateway-server/src/main/java/org/springframework/cloud/gateway/filter/WebsocketRoutingFilter.java#L136) . it will be better to use `r.RawHeader().GetFirst(k)`

pkg/filters/grpcproxy/loadbalance.go
```diff
@@ -0,0 +1,251 @@
...
+		vv = ""
+	} else {
+		vv = v[0]
+	}
+	hash := fnv.New32()
+	hash.Write([]byte(vv))
+	return lb.Servers[hash.Sum32()%uint32(len(lb.Servers))]
+}
+
+type forwardLoadBalancer struct {
```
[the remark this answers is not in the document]
sodaRyCN: behavior like a forward proxy

pkg/util/connectionpool/grpc/pool.go
```diff
@@ -0,0 +1,301 @@
...
+				return wrapper, nil
+			}
+			// discard
+			atomic.AddInt32(&smt.count, -1)
+		case <-timeout.Done():
+			return nil, ErrTimeout // it would better returns ctx.Err()
+		default:
+			cur := atomic.LoadInt32(&smt.count)
+			if cur >= int32(smt.Capacity()) || !atomic.CompareAndSwapInt32(&smt.count, cur, cur+1) {
+				continue
```
localvar, line 207: this could create a busy loop and consume a lot of CPU.

pkg/util/connectionpool/grpc/pool.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	p.mu.RUnlock()
+	p.mu.Lock()
+	defer func() {
+		p.isClosed = true
+		p.mu.Unlock()
+	}()
+
+	p.segment.Range(func(key, value any) bool {
+		close(value.(*segment).clients)
```
localvar, line 153: it is possible that close a closed channel, and channel should be closed at the producer side.

pkg/util/connectionpool/grpc/pool.go
```diff
@@ -0,0 +1,301 @@
...
+	c.segment.pool.mu.RLock()
+	poolClosed := c.segment.pool.IsClosed()
+	c.segment.pool.mu.RUnlock()
+	if poolClosed {
+		return ErrClosed
+	}
+
+	if atomic.LoadUint32(&c.returned) == closed || !atomic.CompareAndSwapUint32(&c.returned, normal, closed) {
+		return ErrAlreadyClosed
+	}
```
localvar, line 270: these checks cannot always prevent data races.

doc/cookbook/grpc-server.md
```diff
@@ -0,0 +1,38 @@
...
+        connectTimeout: 300ms
+    name: grpcforwardproxy
+flow:
+  - filter: grpcforwardproxy
+    jumpIf: {}
+kind: Pipeline
+name: pipeline-grpc-forward
+```
+
+tips: Some gRPC servers manage physical connections by themselves, and they do not consider gRPC gateway scenarios, such as Nacos: when the actual client goes offline, the server will detect and actively close the server's Connection. because gRPC is based on HTT2, it has the characteristics of request and response multiplexing, so there may be a connection from the same easegress used by multiple real clients to the real server. In this case, the server actively closes the connection, which will cause other normal clients to be affected.On the other hand, the server in grpc-go is designed to expect that business processing will not be aware of the underlying connection's status, so it does not provide related APIs for obtaining connection status. I have submitted a related [issue](https://github.com/grpc/grpc-go/issues/5835) in grpc-go to discuss that, you can check it out if you are interested.
```
localvar, line 38:
```suggestion
Note: Some gRPC servers manage physical connections by themselves, and they do not consider gRPC gateway scenarios, such as Nacos: when the actual client goes offline, the server will detect and actively close the server's Connection. because gRPC is based on HTTP2, it has the characteristics of request and response multiplexing, so there may be a connection from the same easegress used by multiple real clients to the real server. In this case, the server actively closes the connection, which will cause other normal clients to be affected. On the other hand, the server in grpc-go is designed to expect that business processing will not be aware of the underlying connection's status, so it does not provide related APIs for obtaining connection status. I have submitted a related [issue](https://github.com/grpc/grpc-go/issues/5835) in grpc-go to discuss that, you can check it out if you are interested.
```

pkg/filters/grpcproxy/codec.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// grpcCodec impl grpc.Codec instead of encoding.Codec, because encoding.Codec would changed grpc.header
+// eg: if use encoding.Codec, then header's content-type=application/grpc -> content-type=application/grpc+xxx
+// xxx from encoding.Codec.Name()
+type grpcCodec struct{}
+
+var (
+	once     = sync.Once{}
+	instance *grpcCodec
```
localvar, line 38: as `grpcCodec` is just `struct{}`, we don't need to use the singleton pattern.

doc/cookbook/grpc-server.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+## Similar to Service Proxy
+Its design and implementation largely refer to the original HTTP Server, so I highly recommended that you read the [Service-Proxy documentation](./service-proxy.md) first.
+
+
+### Configure Whether to Use Connection Pool
+
+create a gRPC forward proxy filter to handler traffic
+notice: `connectTimeout` takes effect if and only when `blockUntilConnected` be specified as true
```
localvar, line 17: in this case, please merge `connectTimeout` and `blockUntilConnected` into one option: `connectTimeout`, when it has a valid value, we consider `blockUntilConnected` to be true, and if it is empty or zero (or other invalid value), we consider `blockUntilConnected` to be false, that's, `blockUntilConnected` can be removed.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,485 @@
...
+		logger.Infof("create new stream fail %s for source addr %s, target addr %s, path %s",
+			result.Error(), spCtx.req.SourceHost(), svr.URL, fullMethodName)
+	}
+	return result
+}
+
+func (sp *ServerPool) biTransport(ctx *serverPoolContext, proxyAsClientStream grpc.ClientStream) error {
+	// Explicitly *do not Close* c2sErrChan and c2sErrChan, otherwise the select below will not terminate.
+	// Channels do not have to be closed, it is just a control flow mechanism, see
+	// https://groups.google.com/forum/#!msg/golang-nuts/pZwdYRGxCIk/qpbHxRRPJdUJ
```
localvar, line 399: it is better to close the channels in the goroutine created by `sp.forwardE2E`, just before it exit.

pkg/filters/grpcproxy/pool.go
```diff
@@ -0,0 +1,485 @@
...
+	c2sErrChan := sp.forwardE2E(ctx.stdr, proxyAsClientStream, nil)
+	s2cErrChan := sp.forwardE2E(proxyAsClientStream, ctx.stdw, ctx.resp.RawHeader())
+	// We don't know which side is going to stop sending first, so we need a select between the two.
+	state := stateNormal
+	for state != stateException {
+		select {
+		case c2sErr := <-c2sErrChan:
+			if c2sErr == io.EOF {
+				// this is the happy case where the sender has encountered io.EOF, and won't be sending anymore./
+				// the clientStream>serverStream may continue pumping though.
```
localvar, line 409:
```suggestion
				// this is the happy case where the sender has encountered io.EOF, and won't be sending anymore.
				// the clientStream --> serverStream may continue pumping though.
```

localvar: for "the clientStream --> serverStream may continue pumping though", is it true? or should it be "the serverStream --> clientStream may continue pumping though"?

sodaRyCN: it is confusing, so I replaced it with a clearer statement.
by the way, easegress use grpc.ClientStream to recv msg from server and send message to server, easegress use grpc.ServerStream to recv msg from client and send message to client.

pkg/filters/grpcproxy/pool.go
sodaRyCN: closing the channel is not always a good option in this scenario. because the client indicates that it will not continue to send datato server on the ClientStream after called `ClientStream.CloseSend`, the server can take a long time to handle business before return IO.EOF or continue to send data in one direction, and the client can also continue processing , so C2S and S2C are not always sync. In this scenario, `select in for loop` will always select the closed channel and take a lot of CPU

localvar (approved):

pkg/object/grpcserver/runtime.go
```diff
@@ -0,0 +1,323 @@
...
+		state         atomic.Value // stateType
+		err           atomic.Value
+		limitListener *limitlistener.LimitListener
+		roundNum      uint64
+
+		eventChan chan interface{}
+	}
+	// Status contains all status generated by runtime, for displaying to users.
+	Status struct {
+		Health bool      `yaml:"health"`
```
xxx7xxxx, line 79: Ditto. We also use json tags in Status, which could be shown in yaml on the client side.

suchen-sci (approved):

xxx7xxxx (approved):

PR #796, opened 2022-09-19
support sticky sessions for load balancer

### LINKED ISSUE #795: [feature] Support sticky sessions for load balancer
### Background
Sticky sesssions is useful in genaral situations, to support it the X problem is:
#### How to identify session
- Session identifier type: cookie, header, parameter, properties(source IP, etc.)
- Session identifier source:  defined by user or created by server
#### Where to persist
- Client-side: cookie is suitable for browser to be added to request, not sure about others
- Server-side: identifiers of all types can be used, but need extra memory and more complicated

### Preferred Solution
1. Use cookie to sticky session, add custom cookie to first response
2. Choose server based on consistent hash of session identifier to reduce remapping when scale
3. For identifier defined by user, use IP hash, header hash and maybe cookie hash

#### Istio Ingress
Istio Ingress uses consistentHash based on HTTP headers, cookies, or other properties (source IP, etc.)
#### Nginx Open Source
Nginx Open Source uses ip_hash and generic hash on variables (source IP, port, URI, etc.)
- IP hash uses first three octets of the IPv4 address or the whole IPv6 address
- IP hash uses down parameter to preserve the current hashing of client IP addresses
- generic hash uses optional consistent parameter to the hash directive enables [ketama](https://www.last.fm/user/RJ/journal/2007/04/10/rz_libketama_-_a_consistent_hashing_algo_for_memcache_clients) consistent to reduce remapping
#### Nginx Plus
Nginx Plus use sticky directive
- sticky cookie adds a session cookie to the first response
`sticky cookie srv_id expires=1h domain=.example.com path=/`
- sticky route assigns a “route” to the client, either a cookie or the request URI
`sticky route $route_cookie $route_uri`
- sticky learn uses cookie created by upstream or else and keeps all info server-side in the shared memory zone
  ```
  sticky learn
     create=$upstream_cookie_examplecookie
     lookup=$cookie_examplecookie
     zone=client_sessions:1m
     timeout=1h
     sync
  ```

### References
[https://istio.tetratelabs.io/istio-in-practice/sticky-sessions/](https://istio.tetratelabs.io/istio-in-practice/sticky-sessions/)
[https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/](https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/)
[https://kubernetes.github.io/ingress-nginx/examples/affinity/cookie/](https://kubernetes.github.io/ingress-nginx/examples/affinity/cookie/)
[https://cloud.google.com/load-balancing/docs/https#session_affinity](https://cloud.google.com/load-balancing/docs/https#session_affinity)
[https://cloud.google.com/load-balancing/docs/backend-service#session_affinity](https://cloud.google.com/load-balancing/docs/backend-service#session_affinity)

samanhappy:
resolves #795

Code changes:
- add sticky configs in BaseLoadBalancer
- choose server using BaseLoadBalancer first, if nil use specific load balancer
- introduce consistent hash


pkg/filters/proxy/loadbalance.go
```diff
@@ -194,11 +202,61 @@ func newHeaderHashLoadBalancer(servers []*Server, key string) *headerHashLoadBal
...
+	lb.SessionCookieName = cookieName
+	if cookieExpires != "" {
+		if d, err := time.ParseDuration(cookieExpires); err == nil {
+			lb.SessionCookieExpires = d
+		}
+	}
+	return lb
+}
+
+func (lb *stickySessionsLoadBalancer) ChooseServer(spCtx *serverPoolContext) *Server {
```
localvar, line 240: This could fail in the case of service discovery, because the order of the available servers could change from time to time.
I think the cookie value should be something related to the url of the servers.

pkg/filters/proxy/loadbalance.go
```diff
@@ -38,17 +40,21 @@ const (
...
 	// LoadBalancePolicyHeaderHash is the load balance policy of HTTP header hash.
 	LoadBalancePolicyHeaderHash = "headerHash"
+	// LoadBalancePolicyStickySessions is the load balance policy of sticky session.
+	LoadBalancePolicyStickySessions = "stickySessions"
 )
 
 // LoadBalancer is the interface of an HTTP load balancer.
 type LoadBalancer interface {
-	ChooseServer(req *httpprot.Request) *Server
+	ChooseServer(spCtx *serverPoolContext) *Server
```
localvar, line 49: As we are discussing in Discord, there's a plan to make the LoadBalancer more generic, that's independent to protocols, change the argument type to `serverPoolContext` makes things harder.

pkg/filters/proxy/pool.go
```diff
@@ -682,6 +685,7 @@ func (sp *ServerPool) buildResponseFromCache(spCtx *serverPoolContext) bool {
 
 	resp.SetStatusCode(ce.StatusCode)
 	resp.SetPayload(ce.Body)
+	setCookies(resp, spCtx.respCookies)
```
localvar, line 688: no need to do this, we don't have cookies in this case.

pkg/filters/proxy/pool.go
```diff
@@ -635,6 +636,8 @@ func (sp *ServerPool) buildResponse(spCtx *serverPoolContext) (err error) {
 		resp = r
 	}
 
+	setCookies(resp, spCtx.respCookies)
```
localvar, line 639: I would prefer to add a new method to the `LoadBalancer` interface for this task.
maybe the method name could be `BindToServer`, the `StickySessionsLoadBalancer` set the cookies, while other load balancers do nothing.

README.zh-CN.md
```diff
@@ -87,7 +87,7 @@
 - **高性能和可用性**
   - **改编**：使用过滤器改编请求和应答。
   - **验证**：标头验证、OAuth2、JWT 和 HMAC 验证。
-  - **负载平衡**：轮询、随机、加权随机、IP哈希、标头哈希。
+  - **负载平衡**：轮询、随机、加权随机、IP哈希、标头哈希、会话保持。
```
localvar, line 90:
```suggestion
  - **负载平衡**：轮询、随机、加权随机、IP哈希、标头哈希、会话锁定。
```

pkg/filters/proxy/pool.go
```diff
@@ -559,6 +559,10 @@ func (sp *ServerPool) doHandle(stdctx stdcontext.Context, spCtx *serverPoolConte
 		return serverPoolError{http.StatusInternalServerError, resultInternalError}
 	}
 
+	if clb, ok := sp.LoadBalancer().(CustomizableLoadBalancer); ok {
+		clb.Customize(svr, spCtx.resp)
+	}
+
```
localvar, line 565: please add a new function to the `LoadBalancer` interface instead of using a new interface, otherwise, developers may adding more and more customized interfaces, and we will have to type assert every interface here.

pkg/filters/proxy/loadbalance.go
```diff
@@ -203,3 +219,62 @@ func (lb *headerHashLoadBalancer) ChooseServer(req *httpprot.Request) *Server {
...
+	if c, err := req.Cookie(lb.SessionName); err == nil {
+		for _, server := range lb.Servers {
+			if server.URL == c.Value {
+				return server
+			}
+		}
+	}
+	// default roundRobin policy
+	counter := atomic.AddUint64(&lb.counter, 1) - 1
+	return lb.Servers[int(counter)%len(lb.Servers)]
```
localvar, line 263:
```suggestion
	return lb.roundRobinLoadBalancer.ChooseServer(req)
```

pkg/filters/proxy/loadbalance.go
```diff
@@ -203,3 +219,62 @@ func (lb *headerHashLoadBalancer) ChooseServer(req *httpprot.Request) *Server {
 	hash.Write([]byte(v))
 	return lb.Servers[hash.Sum32()%uint32(len(lb.Servers))]
 }
+
+// stickySessionsLoadBalancer does load balancing based on sticky sessions.
+type stickySessionsLoadBalancer struct {
+	BaseLoadBalancer
+	counter        uint64
```
localvar, line 226:
```suggestion
	roundRobinLoadBalancer
```

pkg/filters/proxy/loadbalance.go
```diff
@@ -38,53 +41,126 @@ const (
...
 // NewLoadBalancer creates a load balancer for servers according to spec.
-func NewLoadBalancer(spec *LoadBalanceSpec, servers []*Server) LoadBalancer {
+func NewLoadBalancer(spec *LoadBalanceSpec, servers []*Server, slots []*Server) LoadBalancer {
+	d, err := time.ParseDuration(spec.StickyExpire)
+	if err != nil {
+		logger.Errorf("failed to parse duration: %s", spec.StickyExpire)
+	}
+	if spec.StickyCookie == LoadBalanceStickyCookie {
+		logger.Errorf("can't use system cookie name: %s", LoadBalanceStickyCookie)
+	}
```
haoel, line 72: question: why do we keep processing when we encounter the error? if we want to process with the errors, it's better to use the default value.

pkg/filters/proxy/loadbalance.go
```diff
@@ -38,53 +41,144 @@ const (
...
+
+// confSticky configures sticky settings
+func (base *BaseLoadBalancer) confSticky(spec *LoadBalanceSpec) {
+	if spec.Sticky {
+		base.Sticky = spec.Sticky
+
+		if spec.StickyCookie == LoadBalanceStickyCookie {
+			logger.Errorf("can't use system cookie name: %s, will be ignored", LoadBalanceStickyCookie)
+			spec.StickyCookie = ""
+		}
```
haoel, line 101: Question: does here mean if the user uses the cookie name as the same as the system name, we just use the system default cookie name? and what's the difference behavior between the system cookie and the user-defined cookie? if they have the same behavior, I think we should not log any error here.

pkg/filters/proxy/loadbalance.go
```diff
@@ -38,53 +41,144 @@ const (
...
+
+		dur := 2 * time.Hour
+		if spec.StickyExpire != "" {
+			if dur, err := time.ParseDuration(spec.StickyExpire); err == nil {
+				base.StickyExpire = dur
+			} else {
+				logger.Errorf("failed to parse duration: %s, fallback to default 2h", spec.StickyExpire)
+			}
+		}
+		base.StickyExpire = dur
```
haoel, line 112: I think  we can simplify this snippet  to 

```suggestion
		dur, err := time.ParseDuration(spec.StickyExpire)
		if err != nil {
		        dur = 2 * time.Hour
			logger.Warnf("failed to parse duration: %s, fallback to default 2h", spec.StickyExpire)
		}
		base.StickyExpire = dur
```

localvar (approved):

suchen-sci (approved):

PR #802, opened 2022-09-23
add WebSocketProxy and remove WebSocketServer

localvar:


pkg/filters/proxy/wspool.go
```diff
@@ -0,0 +1,233 @@
...
+
+	stdw, _ := ctx.GetData("HTTP_RESPONSE_WRITER").(http.ResponseWriter)
+	if stdw == nil {
+		logger.Debugf("%s: cannot get response writer from context", sp.name)
+		sp.buildFailureResponse(ctx, http.StatusInternalServerError)
+		metric.StatusCode = http.StatusInternalServerError
+		return resultInternalError
+	}
+
+	wssvr := websocket.Server{}
```
suchen-sci, line 165: propose to use `websocket.Handler` here. `websocket.Server` may looks heavy here. 
```
handler := websocket.Handler(func(clntConn *websocket.Conn) {
		// dial to the server
		svrConn, err := sp.dialServer(svr, req)
                ...
})

handler.ServeHTTP(stdw, req.Request)
```

localvar: actually, `websocket.Handler` creates a `websocket.Server`, so it is more heavy.
and we have to use `websocket.Server` because we need to use a nil handshaker.

suchen-sci (approved):

pkg/filters/proxy/wsproxy.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/megaease/easegress/pkg/context"
+	"github.com/megaease/easegress/pkg/filters"
+	"github.com/megaease/easegress/pkg/protocols/httpprot"
+	"github.com/megaease/easegress/pkg/supervisor"
+)
+
+// WebSocketProxyKind is the kind of WebSocketProxy.
+const WebSocketProxyKind = "WebSocketProxy"
+
+var kindWebSocketProxy = &filters.Kind{
```
xxx7xxxx, line 32: IMO, it's better to split every filter into different directories even they have the same infrastructure, which makes the layout clean and has no particular case.

localvar: because the two filters share a lot of code, and the limit time for implement this feature, we have to balance between:
* put them in the same folder
* duplicate the code

and I think the 2nd is worse.

we can do a refactor to put them into different folders later, this only require code level changes, and won't impact the external interface, so we can do it at any time.

As I mentioned before, because we are now using Go 1.18, it is possible for us to extract the load balance and request matching stuff into seperate packages, and I think it will be easier for us to do the refactor after that.

xxx7xxxx: Your argument makes sense to me. The only thing I was curious about is why different folders mean duplicating code with exporting infrastructure.

localvar: the shared code including:
* load balance
* request matching
* service discovery

if I put the filters into different folders, and avoid code duplication at the same time, I need to extract them into seperate packages, but, I don't have enough time to make a good design for the interfaces of the new packages (in fact, I already did some refactor towards this, e.g. extract service discovery to the `BaseServerPool` struct).

zhao-kun: If we compromise this for time cost, I think we should add some TODO comments in the implementation to remind us to refactor it in the future.

localvar: sure, added.

zhao-kun (approved):

pkg/filters/proxy/wsproxy.go
xxx7xxxx: I see, the real problem is no time to design interfaces instead of inevitable duplicate code.

xxx7xxxx (approved):

PR #823, opened 2022-10-10
support routing by query string

nullsimon:
This PR is support routing by query string.

You can config as follows:

- using key and values 
- using regexp

Config examples:

```yaml
  - path: /queryParamsMultiKey
    methods: [GET]
    queries:
    - key: "q"
      values: ["v1", "v2"]
    - key: "q2"
      values: ["v3", "v4"]
  - path: /queryParamsRegexpAndValues2
    methods: [GET]
    queries:
    - key: "id"
      values: ["011"]
      regexp: "[0-9]+"
    backend: 123-pipeline
```

Rules behind

- If your request query key have more than one values, then we just use the first value to match.

Test Examples
```go
// query string values and regexp
stdr, _ = http.NewRequest(http.MethodGet, "http://www.megaease.com/queryParamsRegexpAndValues2", http.NoBody)
stdr.URL.RawQuery = "id=011&&id=baz"
req, _ = httpprot.NewRequest(stdr)
assert.Equal(0, mi.search(req).code)

// query string values and regexp
stdr, _ = http.NewRequest(http.MethodGet, "http://www.megaease.com/queryParamsRegexpAndValues2", http.NoBody)
stdr.URL.RawQuery = "id=baz&&id=011"
req, _ = httpprot.NewRequest(stdr)
assert.Equal(400, mi.search(req).code)

// query string values and regexp
stdr, _ = http.NewRequest(http.MethodGet, "http://www.megaease.com/queryParamsRegexpAndValues2", http.NoBody)
stdr.URL.RawQuery = "id=baz"
req, _ = httpprot.NewRequest(stdr)
assert.Equal(400, mi.search(req).code)
```

### DIFF
--- a/pkg/object/httpserver/mux.go
+++ b/pkg/object/httpserver/mux.go
@@ -331,7 +331,7 @@ func (mp *MuxPath) matchQueries(r *httpprot.Request) bool {
 			return false
 		}
 
-		if q.Regexp != "" && !q.queryRE.MatchString(v) {
+		if q.Regexp != "" && !q.re.MatchString(v) {
 			return false
 		}
 	}
@@ -575,7 +575,7 @@ func (mi *muxInstance) search(req *httpprot.Request) *route {
 
 	// The key of the cache is req.Host + req.Method + req.URL.Path,
 	// and if a path is cached, we are sure it does not contain any
-	// headers.
+	// headers or any queries.
 	r := mi.getRouteFromCache(req)
 	if r != nil {
 		if r.code != 0 {
@@ -628,7 +628,7 @@ func (mi *muxInstance) search(req *httpprot.Request) *route {
 				queryMismatch = true
 				continue
 			}
-			
+
 			if !allowIP(path.ipFilter, ip) {
 				return forbidden
 			}
--- a/pkg/object/httpserver/mux.go
+++ b/pkg/object/httpserver/mux.go
@@ -93,6 +93,7 @@ type (
 		headers           []*Header
 		clientMaxBodySize int64
 		matchAllHeader    bool
+		queries           []*Query
 	}
 
 	route struct {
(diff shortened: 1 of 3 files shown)


pkg/object/httpserver/mux.go
```diff
@@ -544,7 +569,7 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 }
 
 func (mi *muxInstance) search(req *httpprot.Request) *route {
-	headerMismatch, methodMismatch := false, false
+	headerMismatch, methodMismatch, queryMismatch := false, false, false
```
localvar, line 572: the comment at L576 - L578 should also be updated to mention that we are sure the cached path does not contain any queries.

pkg/object/httpserver/spec.go
```diff
@@ -211,3 +221,15 @@ func (p *Path) Validate() error {
 
 	return nil
 }
+
+func (q *Query) initQueryRoute() {
+	q.queryRE = regexp.MustCompile(q.Regexp)
```
localvar, line 226: when the regexp is an empty string, we don't need to compile it, this could save some memory.

pkg/object/httpserver/spec.go
```diff
@@ -101,6 +102,15 @@ type (
...
 		headerRE *regexp.Regexp
 	}
+
+	// Query is the third level entry
+	Query struct {
+		Key    string   `json:"key" jsonschema:"required"`
+		Values []string `json:"values,omitempty" jsonschema:"omitempty,uniqueItems=true"`
+		Regexp string   `json:"regexp,omitempty" jsonschema:"omitempty,format=regexp"`
+
+		queryRE *regexp.Regexp
```
localvar, line 112: per Golang naming convention, we should not repeat type names in field names.

```suggestion
		re *regexp.Regexp
```

localvar (approved):

suchen-sci (approved):

PR #825, opened 2022-10-12
fix random test failure

localvar:


pkg/cluster/member.go
```diff
@@ -139,10 +139,10 @@ func (m *members) store() {
 func (m *members) self() *member {
 	m.RLock()
 	defer m.RUnlock()
-	return m._self()
+	return m.unsafeSelf()
```
xxx7xxxx, line 142: Actually, the member feature was deprecated and should be deleted.

localvar: Yes, but it requires some efforts. And the code changes in this PR are not as much as it looks, most code changes are a result of moving a function from one file to another.

xxx7xxxx (approved):

PR #834, opened 2022-10-20
fix unexpected EOF when compression is enabled

localvar:
when compression is enabled in a `Proxy`, it may fail with `io.UnexpectedEOF` and return HTTP 500 to the client.

### DIFF
--- a/pkg/object/httpserver/mux.go
+++ b/pkg/object/httpserver/mux.go
@@ -549,7 +549,7 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 	}
 	err := req.FetchPayload(maxBodySize)
 	if err == httpprot.ErrRequestEntityTooLarge {
-		logger.Debugf("%s: %s", mi.superSpec.Name(), err.Error())
+		logger.Debugf("%s: %s, you may need to increase 'clientMaxBodySize' or set it to -1", mi.superSpec.Name(), err.Error())
 		buildFailureResponse(ctx, http.StatusRequestEntityTooLarge)
 		return
 	}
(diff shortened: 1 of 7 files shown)


xxx7xxxx (approved):

pkg/object/httpserver/mux.go
```diff
@@ -549,7 +549,7 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 	}
 	err := req.FetchPayload(maxBodySize)
 	if err == httpprot.ErrRequestEntityTooLarge {
-		logger.Debugf("%s: %s", mi.superSpec.Name(), err.Error())
+		logger.Debugf("%s: %s, you may need to increase 'clientMaxBodySize' or set it to -1", mi.superSpec.Name(), err.Error())
```
zhao-kun, line 552: should be logged at the level of the ERROR?

localvar: I think these messages can be triggered by the client, which means an attacker can generate tons of them.
So I set them all to debug.

suchen-sci: use tools like sync.Once to print it once?

localvar: using `sync.Once` means the user cannot reproduce the error. 
I think a better solution is to support changing the lowest log level at runtime.

zhao-kun: Just a little cost of the I/O, in debug level, the user hard to diagnose the problem. If users switch to output debug, they will pay more cost of IO and disk.

localvar: If we change this one to `Error`, we should also change some `Debug`s to `Error`s, so the I/O cost may not be minimal.

and IMHO, I think 'error' is something unexpected and we cannot handle correctly, while `debug` is something we know could happen.

zhao-kun: In my understanding, from the user's perspective, it's unexpected, the error was produced by the easegress instead of backend or frontend.

zhao-kun: the users couldn't be familiar with the codes like you.

localvar: Ok, updated.

zhao-kun (approved):

PR #835, opened 2022-10-22
add support routing by clientIPs

nullsimon:
# add support routing by clientIPs

## What

This PR adds support routing by clientIPs.

## Why

provide rich routing rules

## How

- support muxRule and path for clientIPs routing with allow and exclude ip lists based on ipFilter
- update ipfilter allow function strategy.

## How was this change tested?

- added some unit test cases

## Need to Check?

is this the right way?

pkg/util/ipfilter/ipfilter.go
```go
// if AllowIPs is not empty, only allow IPs in AllowIPs
if len(f.spec.AllowIPs) > 0 && !allowed {
    return false
}
```


pkg/object/httpserver/mux.go
```diff
@@ -569,25 +569,16 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
...
 func (mi *muxInstance) search(req *httpprot.Request) *route {
-	headerMismatch, methodMismatch, queryMismatch := false, false, false
+	headerMismatch, methodMismatch, queryMismatch, clientIPMismatch := false, false, false, false
 
 	ip := req.RealIP()
 
 	// The key of the cache is req.Host + req.Method + req.URL.Path,
 	// and if a path is cached, we are sure it does not contain any
-	// headers or any queries.
+	// headers,any queries and any ipFilters.
```
localvar, line 578:
```suggestion
	// headers, any queries, and any ipFilters.
```

pkg/object/httpserver/mux.go
```diff
@@ -613,8 +605,8 @@ func (mi *muxInstance) search(req *httpprot.Request) *route {
 				continue
 			}
 
-			// only if headers and query are empty, we can cache the result.
-			if len(path.headers) == 0 && len(path.queries) == 0 {
+			// only if headers,queries and ipFilter are empty, we can cache the result.
```
localvar, line 608:
```suggestion
			// we can cache the result only when headers, queries, and ipFilter are empty,
```

pkg/object/httpserver/mux.go
```diff
@@ -569,25 +569,16 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
...
-			return r
-		}
-		if r.path.ipFilterChain == nil {
-			return r
-		}
-		if r.path.ipFilterChain.Allow(ip) {
-			return r
-		}
-		return forbidden
+		return r
```
localvar, line 581: this change hasn't covered the case at L585, which result in a blocked IP being allowed.

localvar (changes requested):

pkg/object/httpserver/mux.go
nullsimon: Actually, if a route can be put into cache, there are no ipFilterChain at all, so this case won't happen.

So it is a problem here, if there is a ipFilter at muxInstance, there will be no cache.  :(

I'm trying to get a better way to do that. WIP.

localvar:
@nullsimon , `path.ipFilterChain` seems useless after this PR, could help confirm this and remove it if yes?

nullsimon:
> @nullsimon , `path.ipFilterChain` seems useless after this PR, could help confirm this and remove it if yes?

definitely yes!

further more, I think as for now in mi.search,ipFilterChain is useless now. So I removed ipFilterChain for path, host and muxInstance. Please check it out :)

localvar (approved):

suchen-sci (approved):

PR #837, opened 2022-10-27
Update keyfunc for compatibility and bug fixes

MicahParks:
The purpose of this PR is to update the `github.com/MicahParks/keyfunc` package. The primary reason I'm making this PR is to introduce the changes from these releases:
* [Support non-RFC compliant base64url padding](https://github.com/MicahParks/keyfunc/releases/tag/v1.1.0) (Compatibility update.)
* [Check HTTP status code](https://github.com/MicahParks/keyfunc/releases/tag/v1.4.0) (Confirms the HTTP status code is 200. Bug fix.)
* [JSON Web Keys restricted by "use" parameter](https://github.com/MicahParks/keyfunc/releases/tag/v1.5.1) (Better RFC compliance. Could be viewed as a bug fix.)

If there is any interest in the other features added since `v1.0.3`, please see the release page:
https://github.com/MicahParks/keyfunc/releases

For a quick summary of the new `keyfunc.Options`, see the documentation for its fields:
https://pkg.go.dev/github.com/MicahParks/keyfunc#Options

### DIFF
--- a/pkg/filters/oidcadaptor/oidcadaptor.go
+++ b/pkg/filters/oidcadaptor/oidcadaptor.go
@@ -223,10 +223,8 @@ func (o *OIDCAdaptor) initDiscoveryOIDCConf() {
 		}
 	}
 	jwks, err := keyfunc.Get(oidcConf.JwksUri, keyfunc.Options{
-		Client:            httpCli,
-		JWKUseNoWhitelist: true,
-		RefreshInterval:   interval,
-		ResponseExtractor: keyfunc.ResponseExtractorStatusAny,
+		Client:          httpCli,
+		RefreshInterval: interval,
 	})
 	if err != nil {
 		logger.Errorf("failed to get the JWKS from the given URL error: %s", err)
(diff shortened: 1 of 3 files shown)


pkg/filters/oidcadaptor/oidcadaptor.go
```diff
@@ -223,8 +223,10 @@ func (o *OIDCAdaptor) initDiscoveryOIDCConf() {
 		}
 	}
 	jwks, err := keyfunc.Get(oidcConf.JwksUri, keyfunc.Options{
-		Client:          httpCli,
-		RefreshInterval: interval,
+		Client:            httpCli,
+		JWKUseNoWhitelist: true,
+		RefreshInterval:   interval,
+		ResponseExtractor: keyfunc.ResponseExtractorStatusAny,
```
MicahParks, line 229: The new options here, `JWKUseNoWhitelist` and `ResponseExtractor` are 100% backwards compatible fields. I will remove these fields in an upcoming commit to this PR.

`JWKUseNoWhitelist` is set to `true` which means that any `use` JWK parameter is valid. The project has since changed it's default to only allow those with a `use` of `sig`. See [more discussion here](https://github.com/MicahParks/keyfunc/releases/tag/v1.5.1).

`ResponseExtractor` is set to `keyfunc.ResponseExtractorStatusAny` which means that any HTTP response status code will not produce an error. The project has since changed it's default to only allow `http.StatusOK` to not produce an error. See [more discussion here](https://github.com/MicahParks/keyfunc/releases/tag/v1.4.0).

localvar (approved):

suchen-sci (approved):

PR #838, opened 2022-10-28
introduce server match model for sticky session

samanhappy:
This PR is trying to enhance sticky session by introducing a server match model.

The existing model `CookieConsistentHash` will influence a bit stickiness when servers scale out because of re-hashing, but the new model will keep the max stickiness in the same situation by matching server exactly.


pkg/filters/proxy/loadbalance.go
```diff
@@ -147,6 +200,73 @@ func (blb *BaseLoadBalancer) ChooseServer(req *httpprot.Request) *Server {
...
+// match matches plain text byte array for encoded string
+func match(encoded string, plain []byte) bool {
+	signed, err := hex.DecodeString(encoded)
+	if err != nil {
+		return false
+	}
+
+	if len(signed) != KeyLen+sha256.Size {
+		return false
+	}
```
localvar, line 260: we can do this in `chooseServerByServerMatch` to improve performance.
and we can also move the rest of this function into `chooseServerByServerMatch` as it is very simple and has no other usages.

pkg/filters/proxy/loadbalance.go
```diff
@@ -40,18 +47,33 @@ const (
 	LoadBalancePolicyIPHash = "ipHash"
 	// LoadBalancePolicyHeaderHash is the load balance policy of HTTP header hash.
 	LoadBalancePolicyHeaderHash = "headerHash"
+	// StickySessionModeCookieConsistentHash is the sticky session mode of consistent hash on app cookie.
+	StickySessionModeCookieConsistentHash = "CookieConsistentHash"
+	// StickySessionModeCookieServerMatch is the sticky session mode of matching server on cookie generated by load balancer.
+	StickySessionModeCookieServerMatch = "CookieServerMatch"
```
localvar, line 53: we need to find a better name for this mode.

pkg/filters/proxy/loadbalance.go
samanhappy: good suggestion, applied

pkg/filters/proxy/loadbalance.go
localvar:
```suggestion
	// StickySessionModeDurationBased uses a load balancer-generated cookie for stickiness.
	StickySessionModeDurationBased = "DurationBased"
```

localvar (approved):

suchen-sci (approved):

PR #871, opened 2022-12-01
Add LDAP mode for basic authentication

samanhappy:


pkg/filters/validator/ldap.go
```diff
@@ -0,0 +1,106 @@
...
+		UseSSL:             spec.UseSSL,
+		SkipTLS:            spec.SkipTLS,
+		InsecureSkipVerify: spec.Insecure,
+		ServerName:         spec.ServerName,
+		ClientCertificates: spec.certificates,
+	}
+	return v
+}
+
+var fnAuthLDAP = func(v *LDAPValidator, username, password string) error {
```
localvar, line 79: why define this function as a variable?

pkg/filters/validator/basicauth.go
```diff
@@ -313,6 +337,62 @@ func (euc *etcdUserCache) Match(username string, password string) bool {
...
+		return false
+	}
+
+	userdn := fmt.Sprintf("%s=%s,%s", luc.spec.UID, username, luc.spec.BaseDN)
+	if err := luc.client.Conn.Bind(userdn, password); err != nil {
+		logger.Warnf("failed to bind LDAP user %v", err)
+		return false
+	}
+
+	return true
```
localvar, line 379: could this function be called concurrently?

samanhappy: this quesion is very considerate, the `client.Connect()` will create a connection if not exists and then store it which can be used by other function like `Conn.Bind`, but the code `defer luc.client.Close()` will break it and could case concurrency problem, so I remove it.

localvar: do we sure that `luc.client.Conn.Bind` is goroutine safe.

samanhappy: it should be safe according to [https://github.com/go-ldap/ldap/blob/master/ldap_test.go#L198](https://github.com/go-ldap/ldap/blob/master/ldap_test.go#L198), and I also did a test which shows safe.

localvar (approved):

suchen-sci (approved):

PR #875, opened 2022-12-14
Support https for both client and server

### LINKED ISSUE #60: [issue] Cli supports http only

samanhappy:
Resolve #60

### DIFF
--- a/cmd/client/command/common.go
+++ b/cmd/client/command/common.go
@@ -124,11 +124,7 @@ const (
 )
 
 func makeURL(urlTemplate string, a ...interface{}) string {
-	p := HTTPProtocal
-	if CommandlineGlobalFlags.ForceTLS {
-		p = HTTPSProtocal
-	}
-	return p + CommandlineGlobalFlags.Server + fmt.Sprintf(urlTemplate, a...)
+	return CommandlineGlobalFlags.Server + fmt.Sprintf(urlTemplate, a...)
 }
 
 func successfulStatusCode(code int) bool {
@@ -145,20 +141,22 @@ func handleRequest(httpMethod string, url string, yamlBody []byte, cmd *cobra.Co
 		}
 	}
 
+	p := HTTPProtocal
+	if CommandlineGlobalFlags.ForceTLS {
+		p = HTTPSProtocal
+	}
 	tr := http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: CommandlineGlobalFlags.InsecureSkipVerify},
 	}
 	client := &http.Client{Transport: &tr}
-	resp, body := doRequest(httpMethod, url, jsonBody, client, cmd)
+	resp, body := doRequest(httpMethod, p+url, jsonBody, client, cmd)
 
-	if resp.StatusCode == 400 && strings.Contains(string(body), "Client sent an HTTP request to an HTTPS server") {
-		fmt.Println("Warning: upgraded to HTTPS, it's better to turn on option --force-tls")
-		url = strings.ReplaceAll(url, HTTPProtocal, HTTPSProtocal)
-		resp, body = doRequest(httpMethod, url, jsonBody, client, cmd)
+	msg := string(body)
+	if resp.StatusCode == 400 && strings.Contains(msg, "HTTP") && strings.Contains(msg, "HTTPS") {
+		resp, body = doRequest(httpMethod, HTTPSProtocal+url, jsonBody, client, cmd)
 	}
 
 	if !successfulStatusCode(resp.StatusCode) {
-		msg := string(body)
 		apiErr := &APIErr{}
 		err := codectool.Unmarshal(body, apiErr)
 		if err == nil {
--- a/cmd/client/command/common.go
+++ b/cmd/client/command/common.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/megaease/easegress/pkg/util/codectool"
 	"github.com/spf13/cobra"
@@ -32,7 +33,7 @@ type (
 	// GlobalFlags is the global flags for the whole client.
 	GlobalFlags struct {
 		Server             string
-		SSL                bool
+		ForceTLS           bool
 		InsecureSkipVerify bool
 		OutputFormat       string
 	}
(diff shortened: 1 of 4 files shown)


pkg/api/server.go
```diff
@@ -84,7 +84,16 @@ func MustNewServer(opt *option.Options, cls cluster.Cluster, super *supervisor.S
 
 	go func() {
 		logger.Infof("api server running in %s", opt.APIAddr)
-		s.server.ListenAndServe()
+		var err error
+		if s.opt.Secure {
+			logger.Infof("api server running in secure model(https)")
+			err = s.server.ListenAndServeTLS(s.opt.CertFile, s.opt.KeyFile)
+		} else {
+			err = s.server.ListenAndServe()
```
localvar, line 92: remove L86 and:
```suggestion
			logger.Infof("api server (https) running in %s", opt.APIAddr)
			err = s.server.ListenAndServeTLS(s.opt.CertFile, s.opt.KeyFile)
		} else {
		        logger.Infof("api server running in %s", opt.APIAddr)
			err = s.server.ListenAndServe()
```

pkg/api/server.go
```diff
@@ -83,8 +83,21 @@ func MustNewServer(opt *option.Options, cls cluster.Cluster, super *supervisor.S
...
 
 	go func() {
-		logger.Infof("api server running in %s", opt.APIAddr)
-		s.server.ListenAndServe()
+		var err error
+		if s.opt.SSL {
+			if s.opt.CertFile == "" || s.opt.KeyFile == "" {
+				logger.Errorf("cert file and key file must be provided for SSL")
+				return
+			}
```
localvar, line 91: please do the validation in `Options.validate`

localvar:
hi @samanhappy , I was thinking of improving the user experience of `egctl`, I think we can make it auto-detect if HTTPS is enabled in the server or not.

to achieve this, we need the following:
* if `tls` is false, use `http`
* if `tls` is true, use `https`
* if `tls` is not configured, then default to `auto`, which means try `http` then `https` (or vice versa)

the only issue is, under `auto` mode, how much time `egctl` will spend to know that the server does not support HTTP or https.

samanhappy:
@localvar`auto` is always good for user, I found specific error messages for nonconsistent http/https requests:

- use http visit https will get a 400 status and body "Client sent an HTTP request to an HTTPS server.\n"
- use https visit http will get a error with message "http: server gave HTTP response to HTTPS client"

maybe we can do a retry work for these error message, how about it?

localvar:
@samanhappy how much time does it cost? I think this is fine if it is under a few hundred milliseconds.
but please rely on the status code instead of the error message, because, it is possible that there's an nginx between easegress and egctl.

I propose to try HTTP first, and if we got 400, then try HTTPS.

samanhappy:
@localvar it should be quick because there's only a network overhead without business logic, how about this flow:

- `tls` is `false` by default, so use HTTP first, if got a 400 and  body `Client sent an HTTP request to an HTTPS server.\n`, then use HTTPS
- `tls` is `true`, use HTTPS

localvar:
this is fine, but please change the option name `tls` to something like `force-tls`.

cmd/client/command/common.go
```diff
@@ -132,26 +145,22 @@ func handleRequest(httpMethod string, url string, yamlBody []byte, cmd *cobra.Co
...
+	client := &http.Client{Transport: &tr}
+	resp, body := doRequest(httpMethod, url, jsonBody, client, cmd)
 
-	body, err := io.ReadAll(resp.Body)
-	if err != nil {
-		ExitWithErrorf("%s failed: %v", cmd.Short, err)
+	if resp.StatusCode == 400 && strings.Contains(string(body), "Client sent an HTTP request to an HTTPS server") {
+		fmt.Println("Warning: upgraded to HTTPS, it's better to turn on option --force-tls")
+		url = strings.ReplaceAll(url, HTTPProtocal, HTTPSProtocal)
+		resp, body = doRequest(httpMethod, url, jsonBody, client, cmd)
```
localvar, line 157: 1. L154: the error string may change, and if a nginx is deployed between egctl and easegress, the error string may also be different. Propose to only check the if the string contains `HTTP` and `HTTPS`
2. L155: I prefer a transparent upgrade and not log this message
3. L156: if the URL contains another `http`, `ReplaceAll` result in a bug (consider the code will change from time to time). and I think replace `http` with `https` is not a good pattern either. I propose to refactor `makeURL` to return an URL without the scheme, and let `handleRequest` to prepend `http` and/or `https`.

cmd/client/command/common.go
```diff
@@ -145,20 +141,22 @@ func handleRequest(httpMethod string, url string, yamlBody []byte, cmd *cobra.Co
...
 	client := &http.Client{Transport: &tr}
-	resp, body := doRequest(httpMethod, url, jsonBody, client, cmd)
+	resp, body := doRequest(httpMethod, p+url, jsonBody, client, cmd)
 
-	if resp.StatusCode == 400 && strings.Contains(string(body), "Client sent an HTTP request to an HTTPS server") {
-		fmt.Println("Warning: upgraded to HTTPS, it's better to turn on option --force-tls")
-		url = strings.ReplaceAll(url, HTTPProtocal, HTTPSProtocal)
-		resp, body = doRequest(httpMethod, url, jsonBody, client, cmd)
+	msg := string(body)
+	if resp.StatusCode == 400 && strings.Contains(msg, "HTTP") && strings.Contains(msg, "HTTPS") {
```
localvar, line 155: `HTTPS` contains `HTTP`

samanhappy: you are right, we can use " HTTP " with spaces or just check HTTPS, which do you prefer?

localvar: if resp.StatusCode == StatusBadRequest  && !CommandlineGlobalFlags.ForceTLS && strings.Contains(strings.ToUpper(msg), "HTTPS") {

localvar (approved):

suchen-sci (approved):

PR #876, opened 2022-12-14
Update mqtt proxy to broker

suchen-sci:
this pr is to make `MQTTProxy` support broker mode by set `brokerMode` true in spec. Original `MQTTProxy` only publish messages to backend `Kafka`. Now, `MQTTProxy` not only publish message to Kafka but also send messages to the subscribed clients directly. 

Main changes:
1. `cachedTopicManager`: it contains topic and mqtt client mapping information from all easegress nodes. 
2. `SessionCacheManager`: it contains all session information for all mqtt clients from all easegress nodes and update `cachedTopicManager`. 
3. `broker`: watch etcd changes and update `SessionCacheManager`.
4. other updates include optim resend and store, solve conflict of client connect to new easegress instance without disconnect previous one (more details in comment).

### DIFF
--- a/pkg/object/mqttproxy/broker.go
+++ b/pkg/object/mqttproxy/broker.go
@@ -553,13 +553,13 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload 
 	}
 }
 
-func (b *Broker) processBrokerModePublish(clientID string, publish *packets.PublishPacket) {
-	if !b.spec.BrokerMode {
-		return
-	}
-	span := generateNewSpanContext(clientID, publish.TopicName)
-	subscribers, _ := b.topicMgr.findSubscribers(publish.TopicName)
+// splitSubscribers split subscribers to local and remote on broker mode and return
+// client ids of local subscribers and eg names of remote subscribers
+func (b *Broker) splitSubscribers(span *model.SpanContext, publish *packets.PublishPacket) ([]string, map[string]struct{}) {
 	egNames := make(map[string]struct{})
+	clients := []string{}
+
+	subscribers, _ := b.topicMgr.findSubscribers(publish.TopicName)
 	for clientID, subQos := range subscribers {
 		if subQos < publish.Qos {
 			continue
@@ -569,6 +569,13 @@ func (b *Broker) processBrokerModePublish(clientID string, publish *packets.Publ
 			egNames[egName] = struct{}{}
 			continue
 		}
+		clients = append(clients, clientID)
+	}
+	return clients, egNames
+}
+
+func (b *Broker) sendMsgToLocalClient(span *model.SpanContext, publish *packets.PublishPacket, clients []string) {
+	for _, clientID := range clients {
 		client := b.getClient(clientID)
 		if client == nil {
 			logger.SpanDebugf(span, "client %v not on broker %v in eg %v", clientID, b.name, b.egName)
@@ -576,12 +583,14 @@ func (b *Broker) processBrokerModePublish(clientID string, publish *packets.Publ
 			client.session.publish(span, publish.TopicName, publish.Payload, publish.Qos)
 		}
 	}
+}
 
+func (b *Broker) requestTransferToCertainInstances(span *model.SpanContext, remoteEgs map[string]struct{}, publish *packets.PublishPacket) {
 	data := &HTTPJsonData{}
 	data.init(publish)
 	urls, err := b.memberURL(b.egName, b.name)
 	if err != nil {
-		logger.SpanErrorf(span, "eg %v find urls for other egs failed:%v", b.egName, err)
+		logger.SpanErrorf(span, "eg %v find urls for other egs failed: %v", b.egName, err)
 		return
 	}
 	jsonData, err := codectool.MarshalJSON(data)
@@ -589,10 +598,10 @@ func (b *Broker) processBrokerModePublish(clientID string, publish *packets.Publ
 		logger.SpanErrorf(span, "json data marshal failed: %v", err)
 		return
 	}
-	for egName := range egNames {
+	for egName := range remoteEgs {
 		url, ok := urls[egName]
 		if !ok {
-			logger.SpanErrorf(span, "eg %v not find url for eg %v", b.egName, egName)
+			logger.SpanErrorf(span, "eg %s not find url for eg %s", b.egName, egName)
 			continue
 		}
 		req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(jsonData))
@@ -612,6 +621,17 @@ func (b *Broker) processBrokerModePublish(clientID string, publish *packets.Publ
 			resp.Body.Close()
 		}
 	}
+
+}
+
+func (b *Broker) processBrokerModePublish(clientID string, publish *packets.PublishPacket) {
+	if !b.spec.BrokerMode {
+		return
+	}
+	span := generateNewSpanContext(clientID, publish.TopicName)
+	localClients, remoteEgs := b.splitSubscribers(span, publish)
+	b.sendMsgToLocalClient(span, publish, localClients)
+	b.requestTransferToCertainInstances(span, remoteEgs, publish)
 }
 
 func (b *Broker) getClient(clientID string) *Client {
--- a/pkg/object/mqttproxy/broker.go
+++ b/pkg/object/mqttproxy/broker.go
@@ -150,12 +150,7 @@ func newBroker(spec *Spec, store storage, muxMapper context.MuxMapper, memberURL
 
 	} else {
 		broker.sessionCacheMgr = newSessionCacheManager(spec, broker.topicMgr)
-		ch, closeFunc, err := broker.sessMgr.store.watch(sessionStoreKey(""))
-		if (err != nil) || (ch == nil) {
-			logger.SpanErrorf(nil, "get watcher for session failed, %v", err)
-		} else {
-			go broker.watch(ch, closeFunc)
-		}
+		broker.connectWatcher()
 	}
 	return broker
 }
@@ -242,7 +237,7 @@ func (b *Broker) watchDelete(ch <-chan map[string]*string, closeFunc func()) {
 	}
 }
 
-func (b *Broker) reconnectWatcher() {
+func (b *Broker) connectWatcher() {
 	if b.closed() {
 		return
 	}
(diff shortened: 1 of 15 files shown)


suchen-sci:
Previous comment is in #874 . Sorry for inconvince...

pkg/object/mqttproxy/broker.go
```diff
@@ -173,19 +182,24 @@ func (b *Broker) setListener() error {
...
 		return
 	}
 
-	ch, cancelFunc, err := b.sessMgr.store.watchDelete(sessionStoreKey(""))
-	if err != nil {
+	var ch <-chan map[string]*string
+	var cancelFunc func()
+	var err error
+
+	for {
```
localvar, line 194: this could now be a dead loop

pkg/object/mqttproxy/broker.go
```diff
@@ -173,19 +182,24 @@ func (b *Broker) setListener() error {
...
+		if err == nil {
+			break
+		}
 		logger.SpanErrorf(nil, "get watcher for session failed, %v", err)
 		time.Sleep(10 * time.Second)
-		go b.reconnectWatcher()
-		return
 	}
-	go b.watchDelete(ch, cancelFunc)
+	go b.onlyWatchDelete(ch, cancelFunc)
```
localvar, line 202: the code below this line could also be called concurrently, please make sure this is not an issue.

pkg/object/mqttproxy/broker.go
```diff
@@ -230,6 +244,93 @@ func (b *Broker) watchDelete(ch <-chan map[string]*string, closeFunc func()) {
...
 
+func (b *Broker) connectWatcher() {
+	if b.closed() {
+		return
+	}
+
+	var ch <-chan map[string]*string
+	var cancelFunc func()
+	var err error
+	for {
```
localvar, line 255: Ditto

pkg/object/mqttproxy/broker.go
```diff
@@ -450,6 +557,65 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload
...
+			continue
+		}
+		req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(jsonData))
+		if err != nil {
+			logger.SpanErrorf(span, "make new request failed: %v", err)
+		}
+		err = b3.InjectHTTP(req, b3.WithSingleHeaderOnly())(*span)
+		if err != nil {
+			logger.SpanErrorf(span, "inject span failed: %v", err)
+		}
```
localvar, line 609:
```suggestion
		if err != nil {
			logger.SpanErrorf(span, "make new request failed: %v", err)
			continue
		}
		err = b3.InjectHTTP(req, b3.WithSingleHeaderOnly())(*span)
		if err != nil {
			logger.SpanErrorf(span, "inject span failed: %v", err)
			continue
		}
```

pkg/object/mqttproxy/topicmgr.go
```diff
@@ -0,0 +1,241 @@
...
+
+// findSubscribers is used to find all clients that subscribe a certain topic directly or use wildcard.
+// for example, topic "loc/device/event" will find clients that subscribe topic "+/+/+" or "loc/+/event" or "loc/device/event"
+// so, clients subscribe topics that contain or not contain wildcard, and this function will find all subscribed topics that match
+// the given topic.
+func (mgr *levelTopicManager) findSubscribers(levels []string) map[string]byte {
+	mgr.RLock()
+	defer mgr.RUnlock()
+
+	ans := make(map[string]byte)
```
localvar, line 81: naming

suchen-sci: updated!

pkg/object/mqttproxy/broker.go
```diff
@@ -450,6 +553,67 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload
...
+		client := b.getClient(clientID)
+		if client == nil {
+			logger.SpanDebugf(span, "client %v not on broker %v in eg %v", clientID, b.name, b.egName)
+		} else {
+			client.session.publish(span, publish.TopicName, publish.Payload, publish.Qos)
+		}
+	}
+
+	data := &HTTPJsonData{}
```
zhao-kun, line 580: Why broadcast? We have all the subscribers' information, we should accurately forward the packet to the subscribers' node.

pkg/object/mqttproxy/broker.go
```diff
@@ -450,6 +553,67 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload
...
+	}
+	jsonData, err := codectool.MarshalJSON(data)
+	if err != nil {
+		logger.SpanErrorf(span, "json data marshal failed: %v", err)
+		return
+	}
+	for egName := range egNames {
+		url, ok := urls[egName]
+		if !ok {
+			logger.SpanErrorf(span, "eg %v not find url for eg %v", b.egName, egName)
```
zhao-kun, line 595: nit: format with %s instead of %v for string.

pkg/object/mqttproxy/broker.go
suchen-sci: in line 563 to 578, we find subscribers eg name and put it in a map. And only send data to these egs.... It's not broadcast...

pkg/object/mqttproxy/broker.go
```diff
@@ -450,6 +553,67 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload
...
+		} else {
+			client.session.publish(span, publish.TopicName, publish.Payload, publish.Qos)
+		}
+	}
+
+	data := &HTTPJsonData{}
+	data.init(publish)
+	urls, err := b.memberURL(b.egName, b.name)
+	if err != nil {
+		logger.SpanErrorf(span, "eg %v find urls for other egs failed:%v", b.egName, err)
```
zhao-kun, line 584: nit: `failed:%v` -> `failed: %v`

pkg/object/mqttproxy/broker.go
```diff
@@ -450,6 +553,67 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload
 	}
 }
 
+func (b *Broker) processBrokerModePublish(clientID string, publish *packets.PublishPacket) {
```
zhao-kun, line 556: This function accomplishes two things: one is to publish a package to the subscribed devices on the current node, and the other is to publish a package to the subscribed devices on other nodes. I strongly recommend implementing this function as two separate functions and assembling them within this larger function.

pkg/object/mqttproxy/broker.go
zhao-kun: Got. I misunderstood.

pkg/object/mqttproxy/broker.go
suchen-sci: good advice, i will update it now

pkg/object/mqttproxy/client.go
```diff
@@ -329,6 +333,7 @@ func pipelineWrapper(fn processFn, packetType PacketType) processFnWithErr {
 
 func processPublish(c *Client, packet packets.ControlPacket) {
 	publish := packet.(*packets.PublishPacket)
+	c.broker.processBrokerModePublish(c.info.cid, publish)
```
zhao-kun, line 336: `processBrokerModelPublish` could have a significant impact on the throughput of the device's publishing operation.  Can we call it in the goroutine? 
I have a question, I think that it might be more efficient to call the function after sending an acknowledgment to the client."

suchen-sci: yeah, u r right... i will update that...

pkg/object/mqttproxy/broker.go
```diff
@@ -450,6 +553,87 @@ func (b *Broker) sendMsgToClient(span *model.SpanContext, topic string, payload
...
+	}
+
+}
+
+func (b *Broker) processBrokerModePublish(clientID string, publish *packets.PublishPacket) {
+	if !b.spec.BrokerMode {
+		return
+	}
+	span := generateNewSpanContext(clientID, publish.TopicName)
+	localClients, remoteEgs := b.splitSubscribers(span, publish)
```
zhao-kun, line 632: 0. In the function splitSubscribers, the argument span is redundant.
1. It is possible to call the function `generateNewSpanContext` after the call to function `splitSubscribers`.
2. After calling of `splitSubscribers`, we need to check whether the variables of the`localClients` and the `remoteEgs` are both empty. If they are empty, we needn't to call the function `sendMsgToLocalClient` and the function `requestTransferToCertainInstances`
3.  The arguments of `publish` to the functions `sendMsgToLocalClient` and `requestTransferToCertainInstances` should have the same position. The current position of the argument is counter-intuitive.

suchen-sci: thanks a lot for your advice, I will update it now.

suchen-sci: updated!

pkg/object/mqttproxy/spec.go
```diff
@@ -71,6 +71,9 @@ type (
 		ConnectionLimit      *RateLimit    `json:"connectionLimit" jsonschema:"omitempty"`
 		ClientPublishLimit   *RateLimit    `json:"clientPublishLimit" jsonschema:"omitempty"`
 		Rules                []*Rule       `json:"rules" jsonschema:"omitempty"`
+		BrokerMode           bool          `json:"brokerMode" jsonschema:"omitempty"`
```
zhao-kun, line 74: I prefer to enable the broker model by default  because:

1. backward compatible, the previous service can upgrade to the broker model by default instead of updating their spec.
2. Broker model has more extensive use cases.

suchen-sci: I am more prefer not to enable it because previous users can upgrade it and the behavior of MQTTProxy stays unchanged unless they update their spec... But I am open to update it.

pkg/object/mqttproxy/broker.go
```diff
@@ -205,29 +220,111 @@ func (b *Broker) reconnectWatcher() {
...
+			// we should check the `egName` in new session (variable v indicated)
+			// is a different value with current broker name. If it is a different
+			// value, and the current broker contains the same session, it means that
+			// we should disconnect session in the current broker, but we might attention
+			// that we just disconnect session and don't clear global store session
+			// information
+			go b.handleNewSessionInCluster(clientID, v, info)
+
+		} else {
+			logger.SpanDebugf(nil, "client %v recv delete watch %v", clientID, v)
```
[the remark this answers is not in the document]
zhao-kun: @suchen-sci what do you think about it?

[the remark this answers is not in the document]
suchen-sci: I agree, I will update it now.

pkg/object/mqttproxy/broker.go
suchen-sci: updated

zhao-kun (approved):

localvar (approved):

PR #880, opened 2022-12-22
Fix/enhancing delete 486

coderabbit214:
enhance object delete 
add -f,-a flag to delete object from a yaml file or delete all object


cmd/client/command/object.go
```diff
@@ -96,23 +96,33 @@ func updateObjectCmd() *cobra.Command {
...
+				visitor.Visit(func(s *spec) error {
+					handleRequest(http.MethodDelete, makeURL(objectURL, s.Name), nil, cmd)
+					return nil
+				})
+				visitor.Close()
+				return
+			}
 
-		Run: func(cmd *cobra.Command, args []string) {
 			handleRequest(http.MethodDelete, makeURL(objectURL, args[0]), nil, cmd)
```
localvar, line 121: args[0] could be invalid after this change.
and `delete` is different from `create` or `update` as it only requires a name instead of full YAML spec, so it is much more easier for the user to make a fault operation. and we should do more checks for the arguments, like what to do if there're both `--all` and `spec file` in the command line?

pkg/api/object.go
```diff
@@ -72,6 +72,11 @@ func (s *Server) objectAPIEntries() []*Entry {
 			Method:  "DELETE",
 			Handler: s.deleteObject,
 		},
+		{
+			Path:    ObjectPrefix,
```
localvar, line 76: this could implement the feature. however, deleting all objects could be a disastrous operation, and some people use `curl` to send the command, this `Path` increases the possibility of their faulty operation.

cmd/client/command/object.go
```diff
@@ -96,23 +96,33 @@ func updateObjectCmd() *cobra.Command {
...
 		},
 	}
-
+	cmd.Flags().StringVarP(&specFile, "file", "f", "", "A yaml file specifying the object.")
+	cmd.Flags().BoolVarP(&allFlag, "all", "a", false, "Delete all object.")
```
localvar, line 125: don't use the short version to help users avoid faulty operation.

```suggestion
	cmd.Flags().BoolVarP(&allFlag, "all", "", false, "Delete all object.")
```

cmd/client/command/object.go
```diff
@@ -96,23 +97,50 @@ func updateObjectCmd() *cobra.Command {
...
 	cmd := &cobra.Command{
-		Use:     "delete",
-		Short:   "Delete an object",
-		Example: "egctl object delete <object_name>",
+		Use:   "delete",
+		Short: "Delete an object from a yaml file or name",
 		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) != 1 {
-				return errors.New("requires one object name to be deleted")
+
```
localvar, line 106:
```suggestion
```

cmd/client/command/object.go
```diff
@@ -96,23 +97,50 @@ func updateObjectCmd() *cobra.Command {
...
+			if len(args) != 0 && len(specFile) != 0 {
+				return errors.New("--file and <object_name> cannot be used together")
 			}
 
 			return nil
 		},
-
 		Run: func(cmd *cobra.Command, args []string) {
+
```
localvar, line 123:
```suggestion
```

pkg/api/object.go
```diff
@@ -151,6 +156,22 @@ func (s *Server) deleteObject(w http.ResponseWriter, r *http.Request) {
...
+		defer s.Unlock()
+
+		specs := s._listObjects()
+		for _, spec := range specs {
+			s._deleteObject(spec.Name())
+		}
+
+		s.upgradeConfigVersion(w, r)
+	}
+
```
localvar, line 172:
```suggestion
```

localvar (approved):

suchen-sci (approved):

PR #883, opened 2022-12-27
add new filter of redirector to do redirect of 3xx

suchen-sci:


pkg/filters/redirector/redirector_test.go
```diff
@@ -0,0 +1,170 @@
...
+		getSpec("(.*)", "uri", "$1", 200), // invalid status code
+		getSpec("(+)", "uri", "$1", 301),  // invalid regex
+	} {
+		msg := fmt.Sprintf("case %d failed.", i)
+		r := &Redirector{spec: t}
+		assert.Panics(func() { r.Init() }, msg)
+	}
+
+	// test complicated regex
+	for i, t := range []testCase{
```
haoel, line 126: I'd like to add the following classical test cases:

- **Case 1: URI Prefix Redirect**
   -  Spec: matchPart: `uri`, match `^(.*)$`, replacement `/prefix$1`
   -  input: `https://example.com/path/to/api/?key1=123&key2=456`
   -  output: `https://example.com/prefix/path/to/api/?key1=123&key2=456`

- **Case 2: Domain Redirect**
   -  Spec: matchPart: `url`, match `(^.*\/\/)([^\/]*)(.*$)`, replacement `$1my.com/$3`
   -  input: `https://example.com/path/to/api/?key1=123&key2=456`
   -  output: `https://my.com/path/to/api/?key1=123&key2=456`

- **Case 3: Path Redirect**
   -  Spec: matchPart: `path`, match `(user)\.php\?id=(\d*)`, replacement `/api/$1/$2`
   -  input: `https://example.com/path/to/user.php?id=123`
   -  output: `https://example.com/api/user/123`

haoel (changes requested):

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,153 @@
...
+func (r *Redirector) Inherit(previousGeneration filters.Filter) {
+	r.Init()
+}
+
+func (r *Redirector) reload() {
+	if r.spec.StatusCode == 0 {
+		r.spec.StatusCode = 301
+	}
+	if r.spec.StatusCode != 301 && r.spec.StatusCode != 302 {
+		panic("invalid status code of Redirector, only 301 and 302 are supported")
```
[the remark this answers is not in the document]
suchen-sci: No... Easegress will init them with `recover()`. I think logging a warning message and set to default value is better in this case.... I will update it now.

pkg/filters/redirector/redirector_test.go
suchen-sci: All updated! Please check

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Inherit inherits previous generation of Redirector.
+func (r *Redirector) Inherit(previousGeneration filters.Filter) {
+	r.Init()
+}
+
+func (r *Redirector) reload() {
+	if r.spec.StatusCode == 0 {
+		r.spec.StatusCode = 301
+	} else {
```
haoel, line 115: I think we can remove this checking

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,175 @@
...
+		if _, ok := statusCodeMap[r.spec.StatusCode]; !ok {
+			logger.Warnf("invalid status code of Redirector, support 300, 301, 302, 303, 304, 307, 308, use 301 instead")
+			r.spec.StatusCode = 301
+		}
+	}
+
+	if r.spec.MatchPart == "" {
+		r.spec.MatchPart = matchPartURI
+	}
```
haoel, line 124: I think we can remove this checking as well

haoel (changes requested):
just have two additional comments.

pkg/filters/redirector/redirector.go
haoel: I just understand we need the warning log. OK.

haoel (approved):

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,174 @@
...
+func (r *Redirector) Inherit(previousGeneration filters.Filter) {
+	r.Init()
+}
+
+func (r *Redirector) reload() {
+	if r.spec.StatusCode == 0 {
+		r.spec.StatusCode = 301
+	}
+	if _, ok := statusCodeMap[r.spec.StatusCode]; !ok {
+		logger.Warnf("invalid status code of Redirector, support 300, 301, 302, 303, 304, 307, 308, use 301 instead")
```
xxx7xxxx, line 117: Actually, you can log `logger.Errorf("BUG: ...")`, because it can't get the wrong code here after validating spec.

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,174 @@
...
+		r.spec.StatusCode = 301
+	}
+
+	if r.spec.MatchPart == "" {
+		r.spec.MatchPart = matchPartURI
+	}
+	r.spec.MatchPart = strings.ToLower(r.spec.MatchPart)
+	if !stringtool.StrInSlice(r.spec.MatchPart, []string{matchPartURI, matchPartFull, matchPartPath}) {
+		logger.Warnf("invalid match part of Redirector, only uri, full and path are supported, use uri instead")
```
xxx7xxxx, line 126: Ditto.

suchen-sci: updated!

xxx7xxxx (approved):

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,174 @@
...
+		spec *Spec
+		re   *regexp.Regexp
+	}
+
+	// Spec describes the Redirector.
+	Spec struct {
+		filters.BaseSpec `json:",inline"`
+
+		Match       string `json:"match" jsonschema:"required"`
+		MatchPart   string `json:"matchPart" jsonschema:"required,enum=uri|full|path"` // default uri
```
localvar, line 81: propose to improve the name `matchPart`, as `match` is a regular expression, it seems we don't need to distinguish `uri` and `path`.

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (r *Redirector) reload() {
+	if r.spec.StatusCode == 0 {
+		r.spec.StatusCode = 301
+	}
+	if _, ok := statusCodeMap[r.spec.StatusCode]; !ok {
+		logger.Errorf("invalid status code of Redirector, support 300, 301, 302, 303, 304, 307, 308, use 301 instead")
+		r.spec.StatusCode = 301
+	}
```
localvar, line 119: these checks should be done in 'Spec.Validate`.

pkg/filters/redirector/redirector.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if r.spec.MatchPart == "" {
+		r.spec.MatchPart = matchPartURI
```
localvar, line 122: this should be done in `kind.DefaultSpec`

PR #901, opened 2023-01-11
Enhance access log

### LINKED ISSUE #301: API access log need to be extended to include the header info at least
Need the header of each api request to get some useful information with the header flags
Exetend the log format with the header  of each request

**Additional context**
please refer to the aliyun api gateway log setting to provide the solution.i.e. we can customized the log format.
https://www.alibabacloud.com/help/zh/doc-detail/64818.htm

samanhappy:
Resolve #301, this PR is trying to enhance access log by

- include request headers and response headers
- make access log format configurable

### DIFF
--- a/pkg/object/httpserver/mux.go
+++ b/pkg/object/httpserver/mux.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"text/template"
@@ -310,8 +311,8 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 				Proto:       stdr.Proto,
 				StatusCode:  metric.StatusCode,
 				Duration:    metric.Duration,
-				ReqSize:     fmt.Sprintf("%vB", metric.ReqSize),
-				RespSize:    fmt.Sprintf("%vB", metric.RespSize),
+				ReqSize:     strconv.FormatUint(metric.ReqSize, 10) + "B",
+				RespSize:    strconv.FormatUint(metric.RespSize, 10) + "B",
 				Tags:        ctx.Tags(),
 				ReqHeaders:  printHeader(stdr.Header),
 				RespHeaders: printHeader(respHeader),
--- a/pkg/object/httpserver/mux.go
+++ b/pkg/object/httpserver/mux.go
@@ -24,7 +24,6 @@ import (
 	"net/http"
 	"reflect"
 	"regexp"
-	"strconv"
 	"strings"
 	"sync/atomic"
 	"text/template"
@@ -50,7 +49,7 @@ import (
 )
 
 const (
-	defaultAccessLogFormat = "[$Time] [$RemoteAddr $RealIP $Method $URI $Proto $StatusCode] [$Duration rx:$ReqSize tx:$RespSize] [$Tags]"
+	defaultAccessLogFormat = "[{{Time}}] [{{RemoteAddr}} {{RealIP}} {{Method}} {{URI}} {{Proto}} {{StatusCode}}] [{{Duration}} rx:{{ReqSize}}B tx:{{RespSize}}B] [{{Tags}}]"
 )
 
 type (
@@ -97,8 +96,8 @@ type (
 		Proto       string
 		StatusCode  int
 		Duration    time.Duration
-		ReqSize     string
-		RespSize    string
+		ReqSize     uint64
+		RespSize    uint64
 		ReqHeaders  string
 		RespHeaders string
 		Tags        string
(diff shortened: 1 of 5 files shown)


pkg/object/httpserver/mux.go
```diff
@@ -431,3 +467,34 @@ func (mi *muxInstance) exportPrometheusMetrics(stat *httpstat.Metric, backend st
...
+
+func (formatter *accessLogFormatter) format(log *accessLog) string {
+	var buf bytes.Buffer
+	if err := formatter.template.Execute(&buf, log); err != nil {
+		logger.Errorf("format access log failed: %v", err)
+	}
+	return buf.String()
+}
+
+func printHeader(header protocols.Header) string {
```
localvar, line 491: we know the header is an HTTP one, so please use `http.Header` as the argument type.

pkg/object/httpserver/mux.go
```diff
@@ -253,6 +284,7 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 				StatusCode: statusCode,
 				ReqSize:    uint64(reqMetaSize) + uint64(body.BytesRead()),
 				RespSize:   respSize,
+				RespHeader: respHeader,
```
localvar, line 287: it seems `RespHeader` is added to `httpstat.Metric` only for carrying the response header to the logger, then it should not be part of `httpstat.Metric`, we can use a local variable to hold the header.

pkg/object/httpserver/mux.go
```diff
@@ -44,6 +49,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const (
+	defaultAccessLogFormat = "$Time $RemoteAddr $RealIP $Method $URI $Proto $StatusCode $Duration $ReqSize $RespSize ReqHeaders:{$ReqHeaders} RespHeaders:{$RespHeaders} Tags:{$Tags}"
```
localvar, line 53: better keep this default format to the same format before this PR, some external tools may rely on the format for analysis, and we will break them if the format changes.

pkg/object/httpserver/mux.go
```diff
@@ -53,11 +62,12 @@ type (
...
-		spec      *Spec
-		httpStat  *httpstat.HTTPStat
-		topN      *httpstat.TopN
-		metrics   *metrics
+		superSpec     *supervisor.Spec
+		spec          *Spec
+		httpStat      *httpstat.HTTPStat
+		topN          *httpstat.TopN
+		metrics       *metrics
+		alogFormatter *accessLogFormatter
```
localvar, line 70:
```suggestion
		accessLogFormatter *accessLogFormatter
```

pkg/object/httpserver/mux.go
```diff
@@ -431,3 +467,40 @@ func (mi *muxInstance) exportPrometheusMetrics(stat *httpstat.Metric, backend st
...
+			buf.WriteString(", ")
+		}
+		buf.WriteString(fmt.Sprintf("%v: %v", key, values))
+		i++
+	}
+	return buf.String()
+}
+
+func printSize(size uint64) string {
+	return fmt.Sprintf("%vB", size)
```
localvar, line 505: we can add `B` to the format string and avoid using `fmt.Sprintf` (and `printSize`) because `fmt.Sprintf` is slow.

pkg/object/httpserver/mux.go
```diff
@@ -269,18 +301,22 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
...
+				Time:        fasttime.Format(startAt, fasttime.RFC3339Milli),
+				RemoteAddr:  stdr.RemoteAddr,
+				RealIP:      req.RealIP(),
+				Method:      stdr.Method,
+				URI:         stdr.RequestURI,
+				Proto:       stdr.Proto,
+				StatusCode:  metric.StatusCode,
+				Duration:    metric.Duration,
+				ReqSize:     printSize(metric.ReqSize),
+				RespSize:    printSize(metric.RespSize),
```
localvar, line 314: using `fmt.Sprintf` directly is more straightforward and makes the code easier to read.

pkg/object/httpserver/mux.go
localvar: Ok, I see the problem. Seems we need to find a better way for the format string...

samanhappy: yeah, `fmt. Sprintf` is a bit slow, so I replace it to `strconv.FormatUint() + "B"` which is faster.

pkg/object/httpserver/mux.go
```diff
@@ -44,6 +49,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const (
+	defaultAccessLogFormat = "[$Time] [$RemoteAddr $RealIP $Method $URI $Proto $StatusCode] [$Duration rx:$ReqSize tx:$RespSize] [$Tags]"
```
localvar, line 53: for the log format, the current design, only support adding a prefix to a variable but does not support adding a suffix to a variable.
I propose we use `{{` and `}}` to mark a variable, that's, the default format becomes:

```suggestion
	defaultAccessLogFormat = "[{{Time}}] [{{RemoteAddr}} {{RealIP}} {{Method}} {{URI}} {{Proto}} {{StatusCode}}] [{{Duration}} rx:{{ReqSize}}B tx:{{RespSize}}] [{{Tags}}]"
```
note the `B` for `ReqSize` and `RespSize` are now part of the format.

we will need to update `newAccessLogFormatter` accordingly.

localvar (approved):

suchen-sci (approved):

Issue #903, opened 2023-01-12
When body is large, RequestBuilder is costly.  Can RequestAdaptor support template ? So that no deep copy from RequestBuilder will be needed.


hhhcj:
**Is your feature request related to a problem? Please describe.**
I use RemoteFilter to authenticate requeset with no DEFAULT body. If authenticated, RemoteFilter  will add a new header "user-id" to the DEFAULT request and proxy to the backend with DEFAULT body.  IF I want dynamically add a new header to the DEFAULT, I should use RequestBuilder with deep copy from DEFAULT. Some of the request body is large, so this operation is costly.  With deep copy, sometimes I should handle specail charaters from headers and body.  Some header has quotes, some header start with non-alphabet and non-number.

**Describe the solution you'd like**
Can RequestAdaptor support template ? So that no deep copy from RequestBuilder will be needed.

**Describe alternatives you've considered**
- RequestAdaptor support template
- RemoteFilter support noRequestBody option . When enabled noRequestBody option, no request body will be shipped to the remote. And DEFAULT request body will not be changed, but headers changed should be keeped to the namespace.

I prefer the first solution, because the second  couples with business detail.

**Additional context**

---
Thanks for contributing 🎉!

localvar:
hi @hhhcj , I haven't clearly got your point.

Based on my understanding, I think what you want to do is build a new request that has an empty body with RequestBuilder.
This could be done if you don't use `body` in the template of the RequestBuilder.

For RequestAdaptor, we currently just want to keep it simple and don't tend to add template support to it.

hhhcj:
The RequestBuilder is send to RemoteFilter, and the RemoteFilter backend  adds a header to the request. So how can I add this header to the DEFAULT request. Maybe I should make another RequestBuilder with deep copy from the large body.

localvar:
if possible, please attach your pipeline spec here, so that we can understand your requirements.

hhhcj:
pipeline
`
name: csp-be
kind: Pipeline
flow:
  - filter: requestBuilder1
    namespace: verifyRequest
  - filter: remoteFilter
    namespace: verifyRequest
  - filter: resultBuilder
    jumpIf:
      result1: requestBuilder2
      result2: failResponseBuilder
  - filter: requestBuilder2
  - filter: proxy
  - filter: END
  - filter: failResponseBuilder
filters:
  - name: requestBuilder1
    kind: RequestBuilder
    template: |
[...]

PR #905, opened 2023-01-17
sourceNamespace and template parameters can be placed togather in httpprot

hhhcj:
As descirbe in the #903 , the commit let sourceNamespace and template parameters can be placed togather. Template option in requestBuilder takes higher priority than sourceNamespace. As member in template is non-default value, it will overlap the same member in sourceNamespace.


pkg/filters/builder/requestbuilder.go
```diff
@@ -106,16 +102,20 @@ func (rb *RequestBuilder) Inherit(previousGeneration filters.Filter) {
...
 // Handle builds request.
 func (rb *RequestBuilder) Handle(ctx *context.Context) (result string) {
+	p := protocols.Get(rb.spec.Protocol)
 	if rb.spec.SourceNamespace != "" {
 		ctx.CopyRequest(rb.spec.SourceNamespace)
-		return ""
+		if rb.spec.Template == "" {
+			return ""
+		}
+		p.SetRef(ctx.GetOutputRequest())
```
localvar, line 118: this function will be called concurrently, so this will not work.

localvar:
@hhhcj , Thanks for the PR.
The code it changes is one of our most complex parts, so we will need time to check it carefully.

samanhappy:
@hhhcj, thanks great for your contribution, I hava some doubts about this PR

- it seems that the logic in `BuildRequestWithRef` is modifying the current active request but not to build a new request from it, so it may in conflict with the design intent of `RequestBuilder` 
- I remember what you need is to copy some headers from other responses, but you are copying from a request, am I missing something?
- we are considering to enhance the `RequestAdaptor` to modify the original request from other responses, what do you think of it?

hhhcj:
> we are considering to enhance the RequestAdaptor to modify the original request from other responses, what do you think of it?

I agree with it. I prefer this firstly, and I think we can place template option to all of the filters. When [localvar](https://github.com/localvar) said 
> For RequestAdaptor, we currently just want to keep it simple and don't tend to add template support to it.

 at #903, I turn to modify the RequestBuilder.

I will make a pr with template support for RequestAdaptor later. @samanhappy

localvar:
hi @hhhcj , I'm closing the PR as it cannot be accepted:
* `RequestBuilder` is used to build a new request, not modify an existing one.
* the body (payload) of a request could be a [stream](https://github.com/megaease/easegress/blob/main/doc/reference/stream.md), in this case, it is cannot be copied from one request to another.
* a request instance could be referenced in more than one namespace, that's, even `sourceNamespace` and target namespace is identical, copy the existing body to another request may break other references in other namespaces.

PR #907, opened 2023-01-18
RequestAdaptor support HeaderTemplate

hhhcj:
Another way to solve #903 , besides #905

### DIFF
--- a/pkg/filters/builder/builder.go
+++ b/pkg/filters/builder/builder.go
@@ -51,7 +51,7 @@ func (spec *Spec) Validate() error {
 
 func (b *Builder) reload(spec *Spec) {
 	t := template.New("").Delims(spec.LeftDelim, spec.RightDelim)
-	t.Funcs(sprig.TxtFuncMap()).Funcs(extraFuncs)
+	t.Funcs(sprig.TxtFuncMap()).Funcs(ExtraFuncs)
 	b.template = template.Must(t.Parse(spec.Template))
 }
 
@@ -74,7 +74,7 @@ func (b *Builder) Status() interface{} {
 func (b *Builder) Close() {
 }
 
-func prepareBuilderData(ctx *context.Context) (map[string]interface{}, error) {
+func PrepareBuilderData(ctx *context.Context) (map[string]interface{}, error) {
 	requests := make(map[string]interface{})
 	responses := make(map[string]interface{})
 
--- a/pkg/filters/builder/databuilder.go
+++ b/pkg/filters/builder/databuilder.go
@@ -105,10 +105,10 @@ func (db *DataBuilder) reload() {
 
 // Handle builds request.
 func (db *DataBuilder) Handle(ctx *context.Context) (result string) {
-	data, err := prepareBuilderData(ctx)
+	data, err := PrepareBuilderData(ctx)
 
 	if err != nil {
-		logger.Warnf("prepareBuilderData failed: %v", err)
+		logger.Warnf("PrepareBuilderData failed: %v", err)
 		return resultBuildErr
 	}
 
--- a/pkg/filters/builder/extrafuncs.go
+++ b/pkg/filters/builder/extrafuncs.go
@@ -90,7 +90,7 @@ func mergeObject(objs ...map[string]interface{}) interface{} {
 	return out
 }
 
-var extraFuncs = template.FuncMap{
+var ExtraFuncs = template.FuncMap{
 	"addf": func(a, b interface{}) float64 {
 		x, y := toFloat64(a), toFloat64(b)
 		return x + y
--- a/pkg/filters/builder/extrafuncs_test.go
+++ b/pkg/filters/builder/extrafuncs_test.go
@@ -46,20 +46,20 @@ func TestToFloat64(t *testing.T) {
 
 func TestExtraFuncs(t *testing.T) {
 	assert := assert.New(t)
-	assert.Equal(float64(123), extraFuncs["addf"].(func(a, b interface{}) float64)(120, 3))
-	assert.Equal(float64(117), extraFuncs["subf"].(func(a, b interface{}) float64)(120, 3))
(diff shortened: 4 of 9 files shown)


suchen-sci:
Hi, thanks a lot for your contribution! 

To support templates in `RequestAdaptor`, I prefer to support template in all specs of `RequestAdaptor`, not headers only. Any idea here?

hhhcj:
> To support templates in RequestAdaptor, I prefer to support template in all specs of RequestAdaptor, not headers only. 

Glad to do that. I'll resubmit this pr days later.

samanhappy:
Hi guys, I agree that we need to support template in a uniform way, but there are different ways to do it according to me

- Solution A(Template Building): use one template to build a complete `RequestAdaptor` like what we are doing in `RequestBuilder`
  - Pros: easy, same solution
  - Cons: less readable and less convenient when user only want to use one or less values

- Solution B(Templates/Variables Injecting): use syntax `{{}}` to inject template values or `${}` to inject variables
  - Pros: more readable and convenient
  - Cons: more work to do, different solutions
  
How do you think about it @localvar @suchen-sci @hhhcj?

localvar:
hi @samanhappy , for solution B, could you please add an example?

and no matter which solution we choose, we need also update the `ResponseAdaptor` together.

for this PR, files in the `builder` folder have been modified and become a dependency of `RequestAdaptor`, this is not a good practice, if we want to reuse the code, a better way is to refactor the code and move the shared code to a separate package.



> Hi guys, I agree that we need to support template in a uniform way, but there are different ways to do it according to me
> 
> * Solution A(Template Building): use one template to build a complete `RequestAdaptor` like what we are doing in `RequestBuilder`
>   
>   * Pros: easy, same solution
>   * Cons: less readable and less convenient when user only want to use one or less values
> * Solution B(Templates/Variables Injecting): use syntax `{{}}` to inject template values or `${}` to inject variables
>   
>   * Pros: more readable and convenient
>   * Cons: more work to do, different solutions
> 
> How do you think about it @localvar @suchen-sci @hhhcj?

samanhappy:
In my view, solution B can inject values directly like
```
kind: RequestAdaptor
name: request-adaptor-example
header:
  add: 
    token_from_template: {{.requests.DEFAULT.Header.token}}
    token_from_variable: ${token}
```
by the way, the variable references can be combined with `ResultBuilder` if possible.

hhhcj:
After I made a pr #910 , I saw this conversaion.
How about we move both `RequestAdpator` and `ResponseAdaptor` to the builder folder, as DataBuilder , RequestBuilder, ResponseBuilder, ResultBuilder do.  @localvar 
> and no matter which solution we choose, we need also update the `ResponseAdaptor` together.
> 
> for this PR, files in the `builder` folder have been modified and become a dependency of `RequestAdaptor`, this is not a good practice, if we want to reuse the code, a better way is to refactor the code and move the shared code to a separate package.

localvar:
> After I made a pr #910 , I saw this conversaion. How about we move both `RequestAdpator` and `ResponseAdaptor` to the builder folder, as DataBuilder , RequestBuilder, ResponseBuilder, ResultBuilder do. @localvar
> 
> > and no matter which solution we choose, we need also update the `ResponseAdaptor` together.
> > for this PR, files in the `builder` folder have been modified and become a dependency of `RequestAdaptor`, this is not a good practice, if we want to reuse the code, a better way is to refactor the code and move the shared code to a separate package.

this is an option, but we need to take all things into account before we make the final decision.

hhhcj:
I didn't find much difference between solution A and B. Injecting template values is just what uniform template do. No matter you inject one value or inject whole values, the template parsing is needed.

> Hi guys, I agree that we need to support template in a uniform way, but there are different ways to do it according to me
> 
> * Solution A(Template Building): use one template to build a complete `RequestAdaptor` like what we are doing in `RequestBuilder`
>   
>   * Pros: easy, same solution
>   * Cons: less readable and less convenient when user only want to use one or less values
> * Solution B(Templates/Variables Injecting): use syntax `{{}}` to inject template values or `${}` to inject variables
>   
>   * Pros: more readable and convenient
>   * Cons: more work to do, different solutions
> 
> How do you think about it @localvar @suchen-sci @hhhcj?

hhhcj:
We can seperate the solving issuer #903 and making a refactor as two different things. Supporting a uniform template will make a heavy test requirement  and may be blocked by some unexpected problems when adapting to one or more filter. It makes a long-term conversation and testing to some extent. 
We can find an easy and moderate way to fix issuer #903 , and then go on thinking about how to make a uniform template and refactor. 
> this is an option, but we need to take all things into account before we make the final decision.

localvar:
> We can seperate the solving issuer #903 and making a refactor as two different things. Supporting a uniform template will make a heavy test requirement and may be blocked by some unexpected problems when adapting to one or more filter. It makes a long-term conversation and testing to some extent. We can find an easy and moderate way to fix issuer #903 , and then go on thinking about how to make a uniform template and refactor. @localvar
> 
> > this is an option, but we need to take all things into account before we make the final decision.

This is another option: backward compatibility is a requirement of Easegress, if the solution takes two steps, then we must make sure they are compatible.

hhhcj:
> 
> This is another option: backward compatibility is a requirement of Easegress, if the solution takes two steps, then we must make sure they are compatible.

Pr #910 supplies template support for RequestAdaptor as RequestBuilder does. For the backward compatibility, we need update ResponseAdaptor together. Considering the builder folder have been modified and become a dependency of RequestAdaptor, #910  move the RequestAdaptor to the builder folder. 

In addition, sourceNamespace and template are mutually exclusive in the  RequestBuilder, but template will override the static params in pr #910 for the flexible. For consistency, we could change the exclusive clause in RequestBuilder and ResponseBuilder, or update #910 for the compliance with the exclusive clause. @localvar

localvar:
close as duplicated with #910

PR #910, opened 2023-01-29
RequestAdaptor support all fields template

hhhcj:
Another way to solve https://github.com/megaease/easegress/issues/903 , besides https://github.com/megaease/easegress/pull/905, https://github.com/megaease/easegress/pull/907


localvar:
hi @hhhcj , thanks for the PRs.
however, as I mentioned in another thread, this is one of the most complex parts of Easegress though it looks simple.

the best practice would be analysis the problem and make a careful design before coding.

hhhcj:
> hi @hhhcj , thanks for the PRs. however, as I mentioned in another thread, this is one of the most complex parts of Easegress though it looks simple.
> 
> the best practice would be analysis the problem and make a careful design before coding.

A new commit with template support for `ResponseAdaptor`. @localvar

pkg/filters/builder/requestadaptor/requestadaptor.go
```diff
@@ -248,6 +254,22 @@ func (ra *RequestAdaptor) Handle(ctx *context.Context) string {
...
+			return builder.ResultBuildErr
+		}
+
+		tempSpec := &Spec{}
+		if err = ra.Builder.Build(data, tempSpec); err != nil {
+			msgFmt := "RequestAdaptor(%s): failed to build adaptor info: %v"
+			logger.Warnf(msgFmt, ra.Name(), err)
+			return builder.ResultBuildErr
+		}
+		ra.adaptSpecWithTemplate(tempSpec)
```
localvar, line 270: `Handle` will be called concurrently, modifying the spec result in race conditions.
Same issue for the `ResponseAdaptor`.

hhhcj: There are 3 ways to fix the race condition.
- Deep copy the static spec. As the length of static spec is usually small, it's not that costly.
- Give a lock to the static spec, and modify the spec with reference. No deep copy, but the route with template become sequence, and it's costly.
- keep static spec mutually exclusive with template as RequestBuilder does. Lose flexible.

What's your opinion?

localvar: none of the three is good enough, please just use simple if/else, if template has the field then use template, otherwise, use the spec.

hhhcj: I choose first method. If not, it may looks like
```golang
	if tempSpec.Method != "" && ra.spec.Method != method {
		ctx.AddTag(stringtool.Cat("requestAdaptor: method ", method, " adapted to ", tempSpec.Method))
		req.SetMethod(tempSpec.Method)
	} else if tempSpec.Method == "" && ra.spec.Method != "" && ra.spec.Method != method {
		ctx.AddTag(stringtool.Cat("requestAdaptor: method ", method, " adapted to ", ra.spec.Method))
		req.SetMethod(ra.spec.Method)
	}
```
compared to original
```golang
	if ra.spec.Method != "" && ra.spec.Method != method {
		ctx.AddTag(stringtool.Cat("requestAdaptor: method ", method, " adapted to ", ra.spec.Method))
		req.SetMethod(ra.spec.Method)
	}
```
It become so ugly that I can't bare it. To fix the race condition with support to the all field template, I think it's a good way to place the spec into namespace either. Then it's possible to change the spec as wish as we can. 

I simply removed the template support with Path and Sign field for the `RequestAdaptor` , because it needs some consideration for the race condition.

pkg/filters/builder/builder.go
```diff
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	resultBuildErr = "buildErr"
+	// ResultBuildErr build error
+	ResultBuildErr = "buildErr"
```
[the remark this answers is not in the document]
hhhcj: I'll try it.

pkg/filters/builder/requestadaptor/requestadaptor.go
samanhappy: maybe we can merge static spec into template before we call `ra.Builder.Reload(&ra.spec.Spec)` in `reload`, so we can use template to build spec which can be used to modify request/response always in `Handle`

pkg/filters/builder/requestadaptor/requestadaptor.go
```diff
@@ -298,6 +319,28 @@ func (ra *RequestAdaptor) Handle(ctx *context.Context) string {
...
+	}
+	if tempSpec.Host == "" {
+		tempSpec.Host = ra.spec.Host
+	}
+	if tempSpec.Decompress == "" {
+		tempSpec.Decompress = ra.spec.Decompress
+	}
+	if tempSpec.Compress == "" {
+		tempSpec.Compress = ra.spec.Compress
+	}
```
localvar, line 341: `Compress` and `Decompress` should not be a part of the template.

pkg/filters/builder/requestadaptor/requestadaptor.go
```diff
@@ -248,9 +254,24 @@ func (ra *RequestAdaptor) Handle(ctx *context.Context) string {
 	req := ctx.GetInputRequest().(*httpprot.Request)
 	method, path := req.Method(), req.Path()
 
-	if ra.spec.Method != "" && ra.spec.Method != method {
-		ctx.AddTag(stringtool.Cat("requestAdaptor: method ", method, " adapted to ", ra.spec.Method))
-		req.SetMethod(ra.spec.Method)
+	tempSpec := &Spec{}
```
localvar, line 257: `tempSepc` is not a good name, does `temp` mean `template` or `temporary`?

pkg/filters/builder/requestadaptor/requestadaptor.go
localvar: the code is ugly in your example above, but it can be not so ugly if we do a little refactoring, like:

```go
newMethod := tempSpec.Method
if newMethod == "" {
    newMethod = ra.spec.Method
}
if newMethod != "" && newMethod != method {
    ctx.AddTag(stringtool.Cat("requestAdaptor: method ", method, " adapted to ", newMethod))
    req.SetMethod(newMethod)
}
```

localvar: the `sign` field should not be a part of the template, so it is correct to exclude it.
but we **must** include the `path` field, however, we don't need to (and should not) use `pathadaptor.Spec` for it, because in the template, we can easily do more than `pathadaptor.Spec` in a more straightforward way.

pkg/filters/builder/requestadaptor/requestadaptor.go
hhhcj: Maybe someone want to determine Compress or Decompress dynamically. And exposing it to the user is not bad, isn't it?

pkg/filters/builder/requestadaptor/requestadaptor.go
hhhcj: When you use `newMethod := tempSpec.Method`, the deep-copy exist. So it has the same cost as [fix race conditions](https://github.com/megaease/easegress/pull/910/commits/70070741f70c03298d64afaab12a76b38a2c44aa) does, which using adaptSpecWithTemplate to override the empty value of templateSpec with static spec and has less space cost.


> the code is ugly in your example above, but it can be not so ugly if we do a little refactoring, like:
> 
> ```go
> newMethod := tempSpec.Method
> if newMethod == "" {
>     newMethod = ra.spec.Method
> }
> if newMethod != "" && newMethod != method {
>     ctx.AddTag(stringtool.Cat("requestAdaptor: method ", method, " adapted to ", newMethod))
>     req.SetMethod(newMethod)
> }
> ```

pkg/filters/builder/requestadaptor/requestadaptor.go
hhhcj: will rename it to templateSpec

pkg/filters/builder/requestadaptor/requestadaptor.go
localvar: No.
the template should only include fields of the request, `Compress`, `Decompress`, `sign` are configurations of the filter, they are different things.

pkg/filters/builder/requestadaptor/requestadaptor.go
hhhcj: It's a good idea. Not `reload` but `build`. I'll try it.

> maybe we can merge static spec into template before we call `ra.Builder.Reload(&ra.spec.Spec)` in `reload`, so we can use template to build spec which can be used to modify request/response always in `Handle`

hhhcj: > It's a good idea. Not `reload` but `build`. I'll try it.
> 
> > maybe we can merge static spec into template before we call `ra.Builder.Reload(&ra.spec.Spec)` in `reload`, so we can use template to build spec which can be used to modify request/response always in `Handle`

I withdraw my last opinion. Neither `reload` nor `build` will make it, though it will make code more clean. I think `Reload` only accessed for template changed. `Build` is executed by reference and result in race condition, unless we do a deep copy of static spec.

pkg/supervisor/spec.go
```diff
@@ -105,7 +105,7 @@ func (s *Supervisor) NewSpec(config string) (spec *Spec, err error) {
 		panic(verr)
 	}
 
-	// Build final json config and raw spec.
+	// build final json config and raw spec.
```
samanhappy, line 108: redundant modification

pkg/filters/builder/responsebuilder_test.go
```diff
@@ -43,7 +43,7 @@ func getResponseBuilder(spec *ResponseBuilderSpec) *ResponseBuilder {
 	return rb
 }
 
-func setRequest(t *testing.T, ctx *context.Context, ns string, req *http.Request) {
+func setRequestWithNS(t *testing.T, ctx *context.Context, ns string, req *http.Request) {
```
samanhappy, line 46: renaming to `setRequestWithNS` is not so elegant, we can reserve only one `func setRequest(t *testing.T, ctx *context.Context, ns string, req *http.Request)` to support `ns` parameter with nil value

localvar:
hi @hhhcj , `path` must be handled, this may require redesigning the underlying structure of the template (that's not using the `Spec`).

hhhcj:
add another pa in template spec
> hi @hhhcj , `path` must be handled, this may require redesigning the underlying structure of the template (that's not using the `Spec`).

localvar:
> add another pa in template spec
> 
> > hi @hhhcj , `path` must be handled, this may require redesigning the underlying structure of the template (that's not using the `Spec`).

hi @hhhcj , please do not just solve the problem, we should try to solve it in the better or best way.
using the `pathadaptor.Spec` directly is not user-friendly, as I have mentioned:
> however, we don't need to (and should not) use `pathadaptor.Spec` for it, because in the template, we can easily do more than `pathadaptor.Spec` in a more straightforward way.

and

> this may require redesigning the underlying structure of the template (that's not using the `Spec`)

hhhcj:
I get your point now. For the path, we could use another field eg. `PathStr`, or rename `Path` to `PathSpec` which is not  backward compatible and add a field `Path string` to the spec. So template can shares the structure type with static field, and this will be more straightforward on path field.

But if we define that template to complete replace the static word, changing one of the field name to another or adding a new field that static field doesn't own may confuse the user that someone think the template field may work, but after trying, he found that the field has been renamed to another. We designed it to be straightforward and user-friendly but result to be unstraightforward and user-unfriendly. Is it worth to do this? Though we support the whole pathadaptor.Spec in template, but users can simply use pathadaptor.Spec.replace to modify the path by theirselves. @localvar 


> > add another pa in template spec
> > > hi @hhhcj , `path` must be handled, this may require redesigning the underlying structure of the template (that's not using the `Spec`).
> 
> hi @hhhcj , please do not just solve the problem, we should try to solve it in the better or best way. using the `pathadaptor.Spec` directly is not user-friendly, as I have mentioned:
> 
> > however, we don't need to (and should not) use `pathadaptor.Spec` for it, because in the template, we can easily do more than `pathadaptor.Spec` in a more straightforward way.
> 
> and
> 
> > this may require redesigning the underlying structure of the template (that's not using the `Spec`)

hhhcj:
I called to @samanhappy last thursday,  and knew that there will be two definitions on template, one is defining a glabal uniform structure for template or filter differenced template structure. If the latter one we choosed, the template should realize all of the static field directly.

localvar:
hi @hhhcj , after think it twice, I think your comments make sense.
however, please don't use the spec of the filter as the template, as many fields of the spec are not needed by the template. the better way is to define the template as a standalone structure and embed it into the spec.

pkg/filters/builder/requestadaptor.go
```diff
@@ -125,21 +126,23 @@ var signerConfigs = map[string]signerConfig{
...
+		Builder
 
 		pa     *pathadaptor.PathAdaptor
 		signer *signer.Signer
 	}
 
-	// Spec is HTTPAdaptor Spec.
-	Spec struct {
+	// RequestAdaptorBuilderSpec is HTTPAdaptor RequestAdaptorBuilderSpec.
+	RequestAdaptorBuilderSpec struct {
```
localvar, line 143: why rename this structure?

pkg/filters/builder/responseadaptor.go
```diff
@@ -31,46 +31,43 @@ import (
...
 	ResponseAdaptor struct {
-		spec *Spec
+		spec *ResponseAdaptorBuilderSpec
+		Builder
 	}
 
-	// Spec is HTTPAdaptor Spec.
-	Spec struct {
+	// ResponseAdaptorBuilderSpec is HTTPAdaptor ResponseAdaptorBuilderSpec.
+	ResponseAdaptorBuilderSpec struct {
```
localvar, line 68: ditto

pkg/filters/builder/requestadaptor.go
hhhcj: Because it named Spec which confilict with ResponseAdaptor which also named Spec.

localvar: But we can name it `RequestAdaptorSpec`, as all other filters.

hhhcj: According to the [samanhappy](https://github.com/samanhappy),
> we can move requestadaptor and responseadaptor files to the builder directory directly for now to avoid renaming too many variable and method names, and we can do it in the future refactor work if necessary

I moved requestadaptor and responseadaptor files to the builder directory directly, so it comes conflict with the same name `Spec`.

hhhcj: Actually some private fields if it has are not needed by the template indeed.  Define the template as a standalone structure and copy the definition of the common fields  may add redundancy to the code. Actually the template definition is the base structure to the `spec` for `RequestAdaptor` and `ResponseAdaptor`. If we would like to define a standalone structure for the template, I prefer `RequestAdaptorBuilderSpec` and `ResponseAdaptorBuilderSpec` inheriting the template structure to avoid redefinition such as
```golang
		Host       string                `json:"host" jsonschema:"omitempty"`
		Method     string                `json:"method" jsonschema:"omitempty,format=httpmethod"`
		Path       *pathadaptor.Spec     `json:"path,omitempty" jsonschema:"omitempty"`
		PathStr    string                `json:"pathStr,omitempty" jsonschema:"omitempty"`
		Header     *httpheader.AdaptSpec `json:"header,omitempty" jsonschema:"omitempty"`
		Body       string                `json:"body" jsonschema:"omitempty"`
		Compress   string                `json:"compress" jsonschema:"omitempty"`
		Decompress string                `json:"decompress" jsonschema:"omitempty"`
		Sign       *SignerSpec           `json:"sign,omitempty" jsonschema:"omitempty"`
```





> hi @hhhcj , after think it twice, I think your comments make sense.
> however, please don't use the spec of the filter as the template, as many fields of the spec are not needed by the template. the better way is to define the template as a standalone structure and embed it into the spec.

samanhappy: For me the request and response are different, we better not to couple them together, but the template spec and static spec are similar and we can use a inheritance for them like 

> define the template as a standalone structure and embed it into the spec.

doc/reference/filters.md
```diff
@@ -1688,3 +1696,20 @@ result YAML varies from filters and protocols.
   | statusCode | int | HTTP status code, default is 200.  | No |
   | headers | map[string][]string | Headers of the result request. | No |
   | body | string | Body of the result request. | No |
+
+* **Schema of RequestAdaptor**
+
+| Name | Type | Description | Required |
+  |------|------|-------------|----------|
```
samanhappy, line 1703:
```suggestion
|------|------|-------------|----------|
```

doc/reference/filters.md
```diff
@@ -1688,3 +1696,20 @@ result YAML varies from filters and protocols.
...
+| path       | [pathadaptor.Spec](#pathadaptorSpec)         | Rules to revise request path                                                                                                                                                                                        | No       |
+| header     | [httpheader.AdaptSpec](#httpheaderAdaptSpec) | Rules to revise request header                                                                                                                                                                                      | No       |
+| body       | string                                       | If provided the body of the original request is replaced by the value of this option. | No       |
+| host       | string                                       | If provided the host of the original request is replaced by the value of this option. | No       |
+
+* **Schema of ResponseAdaptor**
+
+| Name | Type | Description | Required |
+  |------|------|-------------|----------|
```
samanhappy, line 1713:
```suggestion
|------|------|-------------|----------|
```

pkg/filters/builder/requestadaptor.go
```diff
@@ -125,30 +126,37 @@ var signerConfigs = map[string]signerConfig{
...
+		Compress   string      `json:"compress" jsonschema:"omitempty"`
+		Decompress string      `json:"decompress" jsonschema:"omitempty"`
+		Sign       *SignerSpec `json:"sign,omitempty" jsonschema:"omitempty"`
+	}
+
+	RequestAdaptorTemplate struct {
+		Host    string                `json:"host" jsonschema:"omitempty"`
+		Method  string                `json:"method" jsonschema:"omitempty,format=httpmethod"`
+		Path    *pathadaptor.Spec     `json:"path,omitempty" jsonschema:"omitempty"`
+		PathStr string                `json:"pathStr,omitempty" jsonschema:"omitempty"`
```
localvar, line 157: why add this field? seems it has no usage.

doc/reference/filters.md
```diff
@@ -428,6 +428,10 @@ path:
 | decompress | string                                       | If provided, the request body is replaced by the value of decompressed body. Now support "gzip" decompress                                                                                                          | No       |
 | compress   | string                                       | If provided, the request body is replaced by the value of compressed body. Now support "gzip" compress                                                                                                              | No       |
 | sign   | [requestadaptor.SignerSpec](#requestadaptorsignerspec) | If provided, sign the request using the [Amazon Signature V4](https://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html) signing process with the configuration | No       |
+| template        | string | template to create request, the schema of this option must conform with `protocol`, please refer the [template](#template-of-builder-filters) for more information        | No       |
```
localvar, line 431: the description is not correct, please update according to the filter.

doc/reference/filters.md
```diff
@@ -553,6 +557,10 @@ header:
 | body   | string   | If provided the body of the original request is replaced by the value of this option. | No       |
 | compress | string | compress body, currently only support gzip | No |
 | decompress | string | decompress body, currently only support gzip | No |
+| template        | string | template to create request, the schema of this option must conform with `protocol`, please refer the [template](#template-of-builder-filters) for more information        | No       |
```
localvar, line 560: ditto

pkg/filters/builder/requestadaptor.go
```diff
@@ -125,30 +126,37 @@ var signerConfigs = map[string]signerConfig{
...
 
-		Host       string                `json:"host" jsonschema:"omitempty"`
-		Method     string                `json:"method" jsonschema:"omitempty,format=httpmethod"`
-		Path       *pathadaptor.Spec     `json:"path,omitempty" jsonschema:"omitempty"`
-		Header     *httpheader.AdaptSpec `json:"header,omitempty" jsonschema:"omitempty"`
-		Body       string                `json:"body" jsonschema:"omitempty"`
-		Compress   string                `json:"compress" jsonschema:"omitempty"`
-		Decompress string                `json:"decompress" jsonschema:"omitempty"`
-		Sign       *SignerSpec           `json:"sign,omitempty" jsonschema:"omitempty"`
+		*RequestAdaptorTemplate
```
localvar, line 147:
```suggestion
		*RequestAdaptorTemplate `json:",inline"`
```

pkg/filters/builder/responseadaptor.go
```diff
@@ -31,51 +31,52 @@ import (
...
 	}
 
-	// Spec is HTTPAdaptor Spec.
-	Spec struct {
+	// ResponseAdaptorSpec is HTTPAdaptor ResponseAdaptorSpec.
+	ResponseAdaptorSpec struct {
 		filters.BaseSpec `json:",inline"`
+		Spec             `json:",inline"`
+
+		*ResponseAdaptorTemplate
```
localvar, line 72:
```suggestion
		*ResponseAdaptorTemplate `json:",inline"`
```

hhhcj: Thanks for pointing out. As the validation goes, some fields of ResponseAdaptorTemplate should never be omitted, though some fields is empty or own a default value, but it not means the field could be omitted. If ResponseAdaptorTemplate is nil, it will be corruption. 

```golang
// Validate verifies that at least one of the validations is defined.
func (spec *RequestAdaptorSpec) Validate() error {
	if spec.Decompress != "" && spec.Decompress != "gzip" {
		return fmt.Errorf("RequestAdaptor only support decompress type of gzip")
	}
	if spec.Compress != "" && spec.Compress != "gzip" {
		return fmt.Errorf("RequestAdaptor only support decompress type of gzip")
	}
	if spec.Compress != "" && spec.Decompress != "" {
		return fmt.Errorf("RequestAdaptor can only do compress or decompress for given request body, not both")
	}
	if spec.Body != "" && spec.Decompress != "" {
		return fmt.Errorf("No need to decompress when body is specified in RequestAdaptor spec")
	}
	if spec.Sign == nil {
		return nil
	}
	s := spec.Sign
	if s.APIProvider != "" {
		if _, ok := signerConfigs[s.APIProvider]; !ok {
			return fmt.Errorf("%q is not a supported API provider", s.APIProvider)
		}
	}

	return nil
}
```

localvar (approved):

suchen-sci (approved):

samanhappy (approved):

PR #926, opened 2023-02-08
add tracing with Cloudflare

LokiWager:
Add edge networks tracing support for Easegress.

1. add tracing span when the traffic enters Cloudflare.
2. add "Usage with Cloudflare" section in "Distributed Tracing" document.


pkg/tracing/tracing.go
```diff
@@ -500,10 +502,13 @@ func (t *Tracer) NewSpan(ctx context.Context, name string) *Span {
 }
 
 // NewSpanWithStart creates a span with specify start time.
-func (t *Tracer) NewSpanWithStart(ctx context.Context, name string, startAt time.Time) *Span {
+func (t *Tracer) NewSpanWithStart(ctx context.Context, name string, startAt time.Time, req *http.Request) *Span {
```
localvar, line 505: according to the design, tracing is protocol independent (though there's a function `InjectHTTP`), this change will make it HTTP-specific.

pkg/tracing/tracing.go
```diff
@@ -512,6 +517,17 @@ func (t *Tracer) newSpanWithStart(ctx context.Context, name string, startAt time
 	return &Span{Span: span, ctx: ctx, tracer: t}
 }
 
+func (t *Tracer) newSpanWithCDN(ctx context.Context, name string, req *http.Request) *Span {
```
localvar, line 520: the function name is `SpanWithCDN`, but the implementation only supports cloudflare, which does not align with the implementation of `enableCDN` (currently, `enableCDN` only supports cloudflare too, but it is easy to see it will be extended to support other CDNs) .

localvar (approved):

pkg/object/httpserver/mux.go
```diff
@@ -250,7 +250,7 @@ func (mi *muxInstance) serveHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 
 	startAt := fasttime.Now()
 
-	span := mi.tracer.NewSpanWithStart(stdr.Context(), mi.superSpec.Name(), startAt)
+	span := mi.tracer.NewSpanForHTTP(stdr.Context(), mi.superSpec.Name(), stdr)
```
suchen-sci, line 253: Although in `NewSpanForHTTP` call the `fasttime.Now()`, but timestamp in `span` is now different with `startAt` (which is used in metric and accesslog below). Is this ok?

suchen-sci: yeah, and please help delete function `NewSpanWithStart`

suchen-sci (approved):

PR #929, opened 2023-02-13
Fix some issues with install.sh

godruoyi:
This pull request addresses several issues that arose during installation. 

Firstly, some platforms do not have the "abort" command, resulting in an error message:

```
/bin/bash: line 26: abort: command not found
```
We use `echo+exit` to replace abort and set a red color for it when install failed.

Secondly, even though the platform architecture is arm64, there was still an issue with unsupported architecture. 

```
$ uname -m
arm64

---

Unsupport CPU - arm64
```

Finally, when specifying the decompression directory manually, there was an error encountered with:

```
tar: could not chdir to 'easegress'
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -49,7 +49,6 @@ echo "Downloading the release file - \"${ARTIFACT}\" ..."
 curl -sL ${ARTIFACT_URL} -o ${DIR}/${ARTIFACT}
 echo "Downloaded \"${ARTIFACT}\""
 tar -zxf ${DIR}/${ARTIFACT} -C "${DIR}"
-rm ${DIR}/${ARTIFACT}
 echo "Extract the files successfully"
 
 # Fourth - configure the easegress
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+RED='\033[0;31m'
+NC='\033[0m'
 
 # First - check OS.
 OS="$(uname)"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -44,7 +48,8 @@ echo "Create the directory - \"${DIR}\" successfully."
 echo "Downloading the release file - \"${ARTIFACT}\" ..."
 curl -sL ${ARTIFACT_URL} -o ${DIR}/${ARTIFACT}
 echo "Downloaded \"${ARTIFACT}\""
-tar -zxf ${DIR}/${ARTIFACT} -C easegress 
+tar -zxf ${DIR}/${ARTIFACT} -C "${DIR}"
+rm ${DIR}/${ARTIFACT}
```
grootpiano, line 52: Is there need `-f` option to `rm` command?

Sometimes there will have `alias rm='rm -i'` configuration in system.

godruoyi: hi @grootpiano, perhaps it would be safer if we removed this command.

godruoyi: updated, PTAL

grootpiano: LGTM. However, I'm not Easegress's maintainer, yet.

So, please request for haoel's review for double check. 😄

localvar (approved):

scripts/install.sh
```diff
@@ -82,3 +86,5 @@ if [[ "${OS}" == "linux" ]]; then
     sleep 2
     systemctl status easegress
```
haoel, line 87:
```suggestion
    systemctl -q is-active easegress.service  && \
        echo "Easegress service is running!" || \
        systemctl status easegress.service
```

haoel (approved):
Overall, I am fine with this PR. Thanks for finding those errors.

PR #930, opened 2023-02-14
Refactoring the option.Parse method

godruoyi:
The `option.Parse` method should return only failure or success without other logic.

### DIFF
--- a/pkg/option/option.go
+++ b/pkg/option/option.go
@@ -32,7 +32,6 @@ import (
 
 	"github.com/megaease/easegress/pkg/common"
 	"github.com/megaease/easegress/pkg/util/codectool"
-	"github.com/megaease/easegress/pkg/version"
 )
 
 // ClusterOptions defines the cluster members.
@@ -159,7 +158,7 @@ func New() *Options {
 
 	opt.flags.IntVar(&opt.StatusUpdateMaxBatchSize, "status-update-max-batch-size", 20, "Number of object statuses to update at maximum in one transaction.")
 
-	opt.viper.BindPFlags(opt.flags)
+	_ = opt.viper.BindPFlags(opt.flags)
 
 	return opt
 }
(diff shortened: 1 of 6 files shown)


pkg/option/option.go
```diff
@@ -181,27 +180,19 @@ func (opt *Options) renameLegacyClusterRoles() {
...
 	if err != nil {
-		return "", err
-	}
-
-	if opt.ShowVersion {
-		return version.Short, nil
-	}
-
-	if opt.ShowHelp {
-		return opt.flags.FlagUsages(), nil
```
localvar, line 204: Removing these lines changes the logic, that it may fail to show the version or help information if the rest code returns an error.

pkg/option/option.go
```diff
@@ -159,7 +158,7 @@ func New() *Options {
 
 	opt.flags.IntVar(&opt.StatusUpdateMaxBatchSize, "status-update-max-batch-size", 20, "Number of object statuses to update at maximum in one transaction.")
 
-	opt.viper.BindPFlags(opt.flags)
+	_ = opt.viper.BindPFlags(opt.flags)
```
localvar, line 161: why make this change and the changes at L183, L187?

pkg/option/option.go
godruoyi: hi @localvar, I think it's okay for the program to return an error if the user includes some incorrect commands when starting the service while checking the version or help information.

pkg/option/option.go
godruoyi: this is a minor update because the IDE will display a warning if we don't handle the error.

localvar: that's ok, but may I know which IDE you are using? On my side, vs code never reported this warning, and this could not be a help as the error is still not handled.

pkg/option/option.go
godruoyi: I still think it's okay. Consider the following examples, 

```
# Same behavior
easegress-server --help
easegress-server some-not-exists-command --help

# Different behaviors
// Previously, help information was displayed, but now an error is returned instead.
easegress-server -f not_exists_file.yaml --help  
```

where only an incorrectly specified configuration file could cause different behavior between the two. 

If the user specifies an incorrect configuration file (or if the file content is incorrect), returning an error directly without showing help may is acceptable.

pkg/option/option.go
godruoyi: I'm using IntelliJ IDEA, and here's an example where a warning is displayed in IntelliJ IDEA:
image

pkg/option/option.go
localvar: No, if the user wants the version and usage, we need always show the information, we all know there are jokes that a user asks where the `any` key is, so we need to keep the help information as easier to get as possible.

and from the current code, not only an incorrect configuration file will return an error.

and from the maintenance view, the code will change from time to time, even if only an incorrect configuration file returns an error now, developers may add other error cases later, and it is very possible for them to update these codes at the same time, that's result in a bad user experience.

godruoyi: Do you have any good suggestions, or should we just close it?

localvar:
```go
	if opt.ShowVersion || opt.ShowHelp {
		return nil
	}
```

localvar (approved):

suchen-sci (approved):

PR #935, opened 2023-02-22
bug fix: init response object

LokiWager:
init response object


grootpiano:
This works for me to fix flash-sale demo.
I have been looking for this bug for two weeks. Thanks for saving my ass. 😭

PR #940, opened 2023-02-26
init object pool

sodaRyCN:


pkg/util/objectpool/withlock/objectpool.go
```diff
@@ -0,0 +1,179 @@
...
+
+// initializes the pool with initSize of objects
+func (p *Pool) init() error {
+	for i := 0; i < int(p.initSize); i++ {
+		iPoolObject, err := p.new()
+		if err != nil || !iPoolObject.HealthCheck() {
+			logger.Errorf("create pool object failed or object of pool is not healthy when init pool:, err %v", err)
+			continue
+		}
+		atomic.AddInt32(&p.size, 1)
```
localvar, line 109: no need to use atomic functions here.

pkg/util/objectpool/withoutlock/objectpool.go
```diff
@@ -0,0 +1,210 @@
...
+// if no free object, it will create a new one if the pool is not full
+// if the pool is full, it will block until an object is returned to the pool
+func (p *Pool) Get(ctx context.Context) (IPoolObject, error) {
+	backoff := 1
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case poolObject := <-p.store:
+			//fmt.Printf("[%d] id %d size %d got oj from channel\n", time.Now().UnixMicro(), goID(), atomic.LoadInt32(&p.size))
```
localvar, line 128: please remove these useless comments and keep the code clean.

pkg/util/objectpool/withlock/objectpool.go
```diff
@@ -0,0 +1,179 @@
...
+				if poolObject = p.createIPoolObject(); poolObject != nil {
+					return poolObject, nil
+				}
+				continue
+			}
+			return poolObject, nil
+		default:
+			if iPoolObject := p.createIPoolObject(); iPoolObject != nil {
+				return iPoolObject, nil
+			}
```
localvar, line 135: when the pool reaches max size, this default case will create a busy loop.

pkg/util/objectpool/withlock/objectpool.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+}
+
+func (p *Pool) createIPoolObject() IPoolObject {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	for i := 0; i < int(p.maxSize-p.size); i++ {
```
localvar, line 143: why use a loop here?

pkg/util/objectpool/withlock/objectpool.go
```diff
@@ -0,0 +1,179 @@
...
+		if iPoolObject, err := p.new(); err == nil && (!p.checkWhenGet || (p.checkWhenGet && iPoolObject.HealthCheck())) {
```
localvar, line 144: as a new created pool object, I think we don't need to call health check, the `new` function should always return a health object.

pkg/util/objectpool/withoutlock/objectpool.go
```diff
@@ -0,0 +1,210 @@
...
+			}
+		}
+	}
+}
+
+func (p *Pool) createIPoolObject() IPoolObject {
+	if atomic.AddInt32(&p.size, 1) > p.maxSize {
+		atomic.AddInt32(&p.size, -1)
+		return nil
+	}
```
localvar, line 161:
```suggestion
	for {
		size := p.size
		if size > p.maxSize {
			return nil
		}
		if atomic.CompareAndSwapInt32(&p.size, size, size + 1) {
			break
		}
	}
```

pkg/util/objectpool/withoutlock/objectpool.go
```diff
@@ -0,0 +1,210 @@
...
+				//fmt.Printf("[%d] id %d size %d created new oj when default case\n", time.Now().UnixMicro(), goID(), atomic.LoadInt32(&p.size))
+				return iPoolObject, nil
+			}
+			//fmt.Printf("[%d] id %d size %d would go backoff\n", time.Now().UnixMicro(), goID(), atomic.LoadInt32(&p.size))
+			for i := 0; i < backoff; i++ {
+				runtime.Gosched()
+			}
+			if backoff < maxBackoff {
+				backoff <<= 1
+			}
```
localvar, line 152: same issue as the with lock version.

pkg/util/objectpool/withlock/objectpool.go
sodaRyCN: If the `new` function returns err, I hope to retry in the `createIPoolObject` function instead of back to the `Get` function and waiting for the next time. The latter could have a higher probability of context switching. Of course, on the other hand, rapid retrying will still get an err, here I think both have advantages and disadvantages

pkg/util/objectpool/withoutlock/objectpool.go
sodaRyCN: I have considered like that, but in some scenarios, for example, the `new` function always return a unhealthy object, then `pool.size` could be a dynamic balance: `pool.size` always less than `pool.maxsize` and `pool.size` is always changing, so some goroutines will fall into this for loop. 

My code can at least guarantee that created ipoolobject num will not  more than pool.maxsize. Of course, its calculations will not be so accurate in turn, and it will miss the opportunity for it to enter the subsequent process, but at least it will have a chance in the next cycle. Overall in a race, I want to return to the case of context.timeout as much as possible, rather than being stuck in a loop.

sodaRyCN:
keep which one or both

pkg/util/objectpool/withoutlock/objectpool.go
localvar: > I have considered like that, but in some scenarios, for example, the new function always return a unhealthy object, then pool.size could be a dynamic balance: pool.size always less than pool.maxsize and pool.size is always changing, so some goroutines will fall into this for loop.

I'm not getting your point clearly, but I think what you worry about won't happen: using CAS in a loop is a common pattern in lock-free programming.

pkg/util/objectpool/withlock/objectpool.go
localvar: as you mentioned "rapid retrying will still get an err", I propose not using the loop.

localvar (commented):
Please keep the lock-free version.

pkg/util/objectpool/objectpool.go
```diff
@@ -0,0 +1,192 @@
...
+			if iPoolObject := p.createIPoolObject(); iPoolObject != nil {
+				return iPoolObject, nil
+			}
+			for i := 0; i < backoff; i++ {
+				runtime.Gosched()
+			}
+			if backoff < maxBackoff {
+				backoff <<= 1
+			}
+		}
```
localvar, line 144: this default case could still consume a lot of CPU, I think we can leverage `sync.Cond` to avoid it (but this may involve some code changes).

pkg/util/objectpool/objectpool.go
```diff
@@ -0,0 +1,192 @@
...
+	"fmt"
+	"github.com/megaease/easegress/pkg/logger"
+	"runtime"
+	"sync/atomic"
+)
+
+const maxBackoff = 16
+
+// IPoolObject is definition of object that managed by pool
+type IPoolObject interface {
```
localvar, line 32: `IPoolObject` is not a Go style name.

haoel (approved):

suchen-sci (approved):

PR #977, opened 2023-04-04
grpc filter use object pool

sodaRyCN:


pkg/util/objectpool/objectpool.go
```diff
@@ -32,53 +32,51 @@ type PoolObject interface {
...
 		New:          new,
 		CheckWhenGet: true,
 		CheckWhenPut: true,
-	})
+	}, ctx)
 }
 
 // NewWithSpec returns a new pool
-func NewWithSpec(spec Spec) *Pool {
+func NewWithSpec(spec *Spec, ctx context.Context) *Pool {
```
localvar, line 71: why do we need the `ctx` here?
I think the pool should not care about it, otherwise, the code will become too specific and not reusable.
can we find another solution to avoid it?

pkg/util/objectpool/objectpool.go
```diff
@@ -32,53 +32,51 @@ type PoolObject interface {
 }
 
 type (
+	// MultiPool manage multi Pool.
+	MultiPool struct {
```
localvar, line 36: `Pool` is designed to be a library or module, but the `MultiPool` is more tightly bound to the application using the `Pool`, so it should be put into the application code instead of here.

pkg/util/objectpool/objectpool.go
```diff
@@ -158,6 +163,41 @@ func (p *Pool) slowGet(ctx context.Context) (PoolObject, error) {
...
+	var ok bool
+	key := GetSeparatedKey(ctx)
+	if value, ok = m.pools.Load(key); !ok {
+		m.lock.Lock()
+		defer m.lock.Unlock()
+		if value, _ = m.pools.Load(key); !ok {
+			value = NewWithSpec(m.spec, ctx)
+			defer m.pools.Store(key, value)
+		}
+	}
```
localvar, line 197: please use `sync.Map.LoadOrStore` to refactor.

localvar (commented):
hi @sodaRyCN , please checkout my comments, basically, I think this PR brings too much couplings to the code.

pkg/util/objectpool/objectpool.go
sodaRyCN: In some scenarios, func `new` will require some input parameters to implement the business logic, and these parameters cannot be determined when the pool is created, such as the target address of the connection. so I envisage using ctx to pass variables, on the other hand, if the input params are not required in some scenarios, then input nil is ok. . Of course, we can also write anonymous functions to refer to context variables (I tried), but I think the code will be clearer.

pkg/util/objectpool/objectpool.go
sodaRyCN: I will try

pkg/util/objectpool/objectpool.go
sodaRyCN: I tried, but func `LoadOrStore` means that a pool will be created every time, and when the `initNum` of the pool's spec is non-zero, it will create some unused connections at the same time, then these connections need to be closed if  it is load instead of store. 
'LoadOrStore' cannot achieve an effect like this
func LoadOrStore(key interface{}, value func () interface{}) {
 if key exists  {
   return old;
 } else {
  newValue = value()
  store newValue
  .....
}
}

localvar: We can do it like this, though there's still possibility that an unnecessary object being created, but it should be fine now.

```suggestion
	if value, ok = m.pools.Load(key); !ok {
	        value = NewWithSpec(m.spec, ctx)
	        value, _ = m.pools.LoadOrStore(key, value)
	}
```

sodaRyCN: In the case of concurrency, more than one `Pool`  woule be created by L191, and the `initnum` of these pools may be 100、200, etc., and so many `PoolObject`  instances that will not be used will be created. 
L192, the variable `value` is reassigned, It is not good for the `PoolObjects` that need to explicitly call `destroy` or `pool.Close`, such as the underlying connection, variables will be GC, but the connection still exists.
Based on the above reasons, I have a tried like the follow before deciding to use `sync.Mutex`, but it have a problem：`newValue.Close` will always be called.
```
if value, ok = m.pools.Load(key); !ok {
	       newValue = NewWithSpec(m.spec, ctx)
	       if value, ok = m.pools.LoadOrStore(key, newValue) ; ok{
                  newValue.Close()
               }
}
```

pkg/filters/proxies/grpcproxy/proxy.go
```diff
@@ -76,12 +88,22 @@ type (
 
 		mainPool       *ServerPool
 		candidatePools []*ServerPool
+		pool           *MultiPool
+		poolSpec       *objectpool.Spec
```
localvar, line 92: please find a better name, it is too confusion as `mainPool` & `candidatePools` are already there.

pkg/filters/proxies/grpcproxy/proxy.go
```diff
@@ -159,6 +212,45 @@ func (p *Proxy) reload() {
 			p.candidatePools = append(p.candidatePools, pool)
 		}
 	}
+
+	p.borrowTimeout, _ = time.ParseDuration(p.spec.BorrowTimeout)
+	p.timeout, _ = time.ParseDuration(p.spec.Timeout)
+	p.connectTimeout, _ = time.ParseDuration(p.spec.ConnectTimeout)
+
+	if p.pool == nil {
```
localvar, line 220: this should always be true, so no need to check.

sodaRyCN: when updating the proxy, p.pool is not nil.

pkg/util/objectpool/objectpool.go
localvar: I checked the code, this has not been updated.

pkg/util/objectpool/objectpool.go
localvar: I failed to find a better solution, we can accept this for now, but please add a comment to say something about this, to let other developer know this may be changed later.

pkg/util/objectpool/objectpool.go
sodaRyCN: As I said, using `sync.Map` seems inappropriate in this case.
On the one hand it creates some unused object and needs to call `Destroy`, on the other hand, as shown in the code example, `LoadOrStore(key, value any) (actual any, loaded bool)`, **`loaded` always returns true, and `newValue.Close` always be called**. Of course,  we could solve this bug like the following code (use object equals instead of loaded):
```
if value, ok = m.pools.Load(key); !ok {
	       newValue = NewWithSpec(m.spec, ctx)
	       if value, _= m.pools.LoadOrStore(key, newValue) ; value != newValue  {
                  newValue.Close()
               }
}
```
For the reasons, I think it would be better to use `sync.Mutex` instead of `LoadOrStore`.

pkg/filters/proxies/grpcproxy/pool.go
```diff
@@ -87,7 +87,7 @@ func (m *MultiPool) Get(ctx stdcontext.Context) (objectpool.PoolObject, error) {
 	if value, ok = m.pools.Load(key); !ok {
 		m.lock.Lock()
 		defer m.lock.Unlock()
-		if value, _ = m.pools.Load(key); !ok {
+		if value, ok = m.pools.Load(key); !ok {
 			value = objectpool.NewWithSpec(m.spec, ctx)
 			defer m.pools.Store(key, value)
 		}
```
localvar: don't use `defer` here, as the defered function will be called at the exit of the current function.

```go
if value, ok = m.pools.Load(key); !ok {
		m.lock.Lock()
		if value, _ = m.pools.Load(key); !ok {
			value = objectpool.NewWithSpec(m.spec, ctx)
			m.pools.Store(key, value)
		}
		m.lock.Unlock()
}
```

localvar (approved):

suchen-sci (approved):

PR #1040, opened 2023-07-13
Roll rorate log

sodaRyCN:
Support roll all easegress logs include `stdout.log`,`etcd-client` and so on.

The reasons for not using [lumberjack](https://github.com/natefinch/lumberjack) directly are as follows:
- It does not support actively commits the current contents of the file to stable storage , so it may cause log contents loss in some scenarios, such as when the process exits gracefully. [lumberjack issue](https://github.com/natefinch/lumberjack/pull/178/files)
- Hard-coded definition log file access permissions are 0600, unless the log files are created in advance

### DIFF
--- a/doc/reference/controllers.md
+++ b/doc/reference/controllers.md
@@ -6,6 +6,7 @@
     - [TrafficController](#trafficcontroller)
     - [RawConfigTrafficController](#rawconfigtrafficcontroller)
       - [HTTPServer](#httpserver)
+      - [gRPCServer](#grpc-service-proxy)
     - [AccessLogVariable](#accesslogvariable)
       - [Pipeline](#pipeline)
     - [StatusSyncController](#statussynccontroller)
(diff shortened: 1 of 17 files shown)


suchen-sci:
Hi, excellent work. But this pr may drop the test coverage from 81.58% to 80.09%, main reason is some files in logger directory are not tested. Could you please help to add some tests? Please keep the test coverage above 81%. You check code coverage from [here](https://app.codecov.io/gh/megaease/easegress/pull/1040/tree)

sodaRyCN:
I will fix it

localvar (commented):
Overall, is there anything that this PR can resolve but the [logrotate](https://linux.die.net/man/8/logrotate) cannot?

By the design of easegress, it rely on `logrotate` for log rotation, so if this PR does not provide too much help, I think it is better to avoid the complexity it introduced and keep the simplicity.

sodaRyCN:
> 

As [described ](https://github.com/megaease/easegress/pull/1040#issue-1802154436) in the PR。

Of course, if this PR does not provide too much help, we can close the RP until it is really neede，it doesn't matter

PR #1042, opened 2023-07-19
Gateway api

localvar:


pkg/object/gatewaycontroller/gatewaycontroller.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Kind returns the kind of GatewayController.
+func (gc *GatewayController) Kind() string {
+	return Kind
+}
+
+// DefaultSpec returns the default spec of GatewayController.
+func (gc *GatewayController) DefaultSpec() interface{} {
+	return &Spec{}
```
xxx7xxxx, line 80: IMO, `~/.kube/config` could be the default value. And I think the GatewayController could support ClientSetInCluster mode[1].

[1] https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go

pkg/object/gatewaycontroller/translator.go
```diff
@@ -0,0 +1,660 @@
...
+	}
+	header.Del = f.Remove
+
+	b.respAdaptor.Header = header
+}
+
+func (b *pipelineSpecBuilder) addBackend(addr string, weight int) {
+	if b.proxy == nil {
+		b.proxy = &httpproxy.Spec{}
+		b.proxy.Pools = append(b.proxy.Pools, &httpproxy.ServerPoolSpec{ServerMaxBodySize: -1})
```
xxx7xxxx, line 247: A side question, why not make the default value 0 for unlimited body?

pkg/object/gatewaycontroller/k8s.go
```diff
@@ -0,0 +1,490 @@
...
+		sb := b.Subsets[i]
+		if isSubsetsChanged(sa, sb) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func isSubsetsChanged(a, b apicorev1.EndpointSubset) bool {
```
xxx7xxxx, line 82: Why `reflect.DeepEqual` can not cover this function.

pkg/object/gatewaycontroller/gatewaycontroller.go
localvar: the default is in cluster mode(when both `KubeConfig` and `MasterURL` are empty) as it will be deployed in a k8s cluster at most of the time. And ClientSetInCluster mode is used, please check L161 in k8s.go.

pkg/object/gatewaycontroller/k8s.go
```diff
@@ -0,0 +1,490 @@
...
+const (
+	resyncPeriod = 10 * time.Minute
+)
+
+func isResourceChanged(oldObj, newObj interface{}) bool {
+	if oldObj == nil || newObj == nil {
+		return true
+	}
+
+	if oldObj.(metav1.Object).GetResourceVersion() == newObj.(metav1.Object).GetResourceVersion() {
```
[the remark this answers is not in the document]
localvar: `apicorev1.Endpoints` is Object, but this line is only for other objects, for endpoints, we will do additional checks below.

however, just found some of the conditions and return values of this function is wrong, will fix them.

pkg/object/gatewaycontroller/k8s.go
localvar: `DeepEqual` can be used here.
But for performance consideration, explicit comparsion is used.

pkg/object/gatewaycontroller/translator.go
localvar: To support unlimited body, we have to enable the [stream](https://github.com/megaease/easegress/blob/main/doc/reference/stream.md) mode, but in this mode, some of the new features of v2.0 becomes unavailable. so the meaning of "set `ServerMaxBodySize` to zero" is a balance of different things, and we think using `-1` for unlimited body is better for more cases.

xxx7xxxx (approved):

pkg/object/gatewaycontroller/k8s.go
```diff
@@ -0,0 +1,490 @@
...
+
+	minor, err := strconv.Atoi(info.Minor)
+	if err != nil {
+		err = fmt.Errorf("unknown kubernetes minor version: %v", info.Minor)
+		return
+	}
+
+	if minor < 19 {
+		// Ingress version v1 has been added after kubernetes 1.19
+		panic(fmt.Errorf("kubernetes version [%v] is too low, GatewayController requires kubernetes v1.19+", info.GitVersion))
```
suchen-sci, line 218: In doc, the require of k8s is v1.18+, but here is v1.19...

suchen-sci: minor < 19 Panic, it actually is 1.20+?

suchen-sci (approved):

PR #1067, opened 2023-08-18
add egctl edit, egctl logs, update egctl create cmd

suchen-sci:


cmd/client/resources/customdatakind.go
```diff
@@ -0,0 +1,284 @@
...
+		return "", err
+	}
+
+	// reorder yaml, put name, kind in front of other fields.
+	lines := strings.Split(string(yamlBody), "\n")
+	var name string
+	var sb strings.Builder
+	sb.Grow(len(yamlBody))
+	for _, l := range lines {
+		if strings.HasPrefix(l, "name: ") {
```
localvar, line 166:
```suggestion
		if strings.HasPrefix(l, "name:") {
```

cmd/client/resources/object.go
```diff
@@ -90,6 +91,68 @@ func GetAllObject(cmd *cobra.Command) error {
...
+		return general.ErrorMsg(general.EditCmd, err, kind, args.Name)
+	}
+
+	// get old yaml and save it to a temp file
+	var oldYaml string
+	var err error
+	if oldYaml, err = getObjectYaml(args.Name); err != nil {
+		return getErr(err)
+	}
+	filePath := getResourceTempFilePath(kind, args.Name)
```
localvar, line 106: should move this line into `editResource`, there are other cases.

cmd/client/resources/object.go
```diff
@@ -90,6 +91,68 @@ func GetAllObject(cmd *cobra.Command) error {
...
+
+	_, newSpec, err := general.CompareYamlNameKind(oldYaml, newYaml)
+	if err != nil {
+		return getErr(editErrWithPath(err, filePath))
+	}
+	err = ApplyObject(cmd, newSpec)
+	if err != nil {
+		return getErr(editErrWithPath(err, filePath))
+	}
+	os.Remove(filePath)
```
localvar, line 123: move this line into `editResource` and use `defer` to ensure the file is deleted. there are other cases.

cmd/client/resources/object.go
suchen-sci: `filePath` and remove it are out of function `editResource` purposely. It follows the design of `kubectl edit`. The reason of this design is when edit failed, we want user to find the file and change it manually (for example, some typo, and resource is not updated, or they change the name or kind of resource). So the file is removed only if edit is succeeded and resource is updated as expect.

README.md
```diff
@@ -257,6 +257,13 @@ filters:
 The pipeline means it will forward traffic to 3 backend endpoints, using the
 `roundRobin` load balance policy.
 
+You can also create them using `egctl create httpproxy` command.
+```bash
+egctl create httpproxy demo --port 10080 \ 
+  --rule="/pipeline=http://127.0.0.1:9095,http://127.0.0.1:9096,http://127.0.0.1:9097"
+```
+this command will create `HTTPServer` `demo-server` and `Pipeline` `demo-pipeline-0` which work exactly same to `server-demo` and `pipeline-demo`. See more about [`egctl create httpproxy`](./doc/egctl-cheat-sheet.md#create-httpproxy).
+
```
localvar, line 266: propose not adding `-server` suffix to the name, user may get confused what he/she has created.
for the pipeline names, also propose not using the `pipeline-x` suffixes.

suchen-sci: good idea. I will update that!

cmd/client/resources/customdatakind.go
suchen-sci: updated!

cmd/client/commandv2/logs.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/megaease/easegress/v2/cmd/client/general"
+	"github.com/spf13/cobra"
+)
+
+// LogsCmd returns logs command.
+func LogsCmd() *cobra.Command {
+	var n int
+	var follow bool
+	examples := []general.Example{
+		{Desc: "Print the most recent 500 logs by default.", Command: "egctl logs"},
```
xxx7xxxx, line 35: Why not print all logs like other logs by default.

cmd/client/commandv2/edit.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/megaease/easegress/v2/cmd/client/general"
+	"github.com/megaease/easegress/v2/cmd/client/resources"
+	"github.com/spf13/cobra"
+)
+
+// EditCmd returns edit command.
+func EditCmd() *cobra.Command {
+	examples := []general.Example{
+		{Desc: "Edit a resource with name", Command: "egctl edit <resource> <name>"},
```
xxx7xxxx, line 32: Could we specify editor like vim, nano.

cmd/client/commandv2/create/createhttpproxy.go
```diff
@@ -0,0 +1,395 @@
...
+		},
+	}
+
+	cmd.Flags().IntVar(&o.Port, "port", -1, "Port of HTTPServer")
+	cmd.Flags().StringArrayVar(&o.Rules, "rule", []string{}, "Rule in format host/path=endpoint1,endpoint2. Paths containing the leading character '*' are considered as PathPrefix.")
+
+	cmd.Flags().BoolVar(&o.TLS, "tls", false, "Enable TLS")
+	cmd.Flags().BoolVar(&o.AutoCert, "auto-cert", false, "Enable auto cert")
+	cmd.Flags().StringVar(&o.CaCertFile, "ca-cert-file", "", "CA cert file")
+	cmd.Flags().StringArrayVar(&o.CertFiles, "cert-file", []string{}, "Cert file")
```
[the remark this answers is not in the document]
suchen-sci: it use as `--cert-file filename1 --cert-file filename2`... If use `--cert-files` it feels more like `--cert-files filename1,filename2,filename3`, and it should use `cmd.Flags().StringSliceVar`. But it may cause some problem, since filename may itself contains `,`.

cmd/client/commandv2/create/createhttpproxy.go
```diff
@@ -0,0 +1,395 @@
...
+	cmd.Flags().StringArrayVar(&o.KeyFiles, "key-file", []string{}, "Key file")
```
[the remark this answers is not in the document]
suchen-sci: same as `--cert-file` part...

cmd/client/commandv2/edit.go
suchen-sci: Yes, in `egctl-cheat-sheet.md`, mentioned use of env variable of `EGCTL_EDITOR` to change editor. I will add an example about that in examples.

cmd/client/commandv2/logs.go
suchen-sci: Yeah. Here are my concerns.

1. `kubectl logs` will print help information. If print logs `kubectl logs <pod-name>`. And actually `egctl create`, `egctl apply` will print help information. So, based on this, user may think `egctl logs` will print help information. They may think there are commands like `egctl logs std` or `egctl logs http-access` etc. So, by default, I choose `egctl logs` not print all logs, just a few.
2. easegress logs can be really really long. So print all logs may be not necessary... 

Just my naive opinions, I am ok to change it to print all logs.

xxx7xxxx: `git log | cat` will output all of them with `cat` to dismantle the pager effect. It just uses a pager to display logs, essentially it prints all of the logs. But `egctl logs | cat` outputs the last 500 lines. Well, I just provided another thought as I don't strongly insist on my point.  You can decide on the design.

suchen-sci: You are right. I am just a little afraid that in practice, when user run `egctl logs`, and terminal starts to print logs, then they press `CTRL+C`... (It happens to me all the time).
I am more prefer current design that provides a user-friendly default setting...

xxx7xxxx: Then it is.

xxx7xxxx (approved):

PR #1071, opened 2023-08-30
Limit 1 running object file

xxx7xxxx:


pkg/supervisor/object.go
```diff
@@ -310,10 +305,9 @@ func (or *ObjectRegistry) storeConfigInLocal(config map[string]string) {
 	}
 	buff.Write(configBuff)
 
-	configLocalFileName := or.getConfigLocalFileName()
-	err = os.WriteFile(configLocalFileName, buff.Bytes(), 0o644)
+	err = os.WriteFile(or.configLocalPath, buff.Bytes(), 0o644)
```
suchen-sci, line 308: Before write new `running_objects.json`, how about rename old one to `running_objects.bak.json`? 
Since we remove the time from file name, how about add a timestamp to `config`, or use `os.Chtimes(fileName, currentTime, currentTime)` to change to file create time?

xxx7xxxx: Updated. Json file does not support comment, we could just use `stat running_objects.json` to get the time.

suchen-sci (approved):

PR #1087, opened 2023-09-08
chore: consider perallocating 

zjx-ERROR:


pkg/object/meshcontroller/service/service.go
```diff
@@ -209,12 +209,12 @@ func (s *Service) ListServiceCerts() []*spec.Certificate {
 
 // ListAllIngressControllerInstanceCerts  gets the ingress controller cert.
 func (s *Service) ListAllIngressControllerInstanceCerts() []*spec.Certificate {
-	var certs []*spec.Certificate
 	values, err := s.store.GetPrefix(layout.AllIngressControllerInstanceCertPrefix())
 	if err != nil {
 		api.ClusterPanic(err)
 	}
 
+	certs := make([]*spec.Certificate, 0, len(values))
```
suchen-sci, line 217: Other changes look good. But for this one, you may consider to removes unused capacity from the slice, returning `s[:len(s):len(s)]`. Since it is returned.

zjx-ERROR: I agree with you.

suchen-sci (approved):

PR #1088, opened 2023-09-14
Make PersistentVolume local path configurable

codering:


helm-charts/easegress/templates/PersistentVolume.yaml
```diff
@@ -13,7 +13,7 @@ spec:
   persistentVolumeReclaimPolicy: Delete
   storageClassName: easegress-storage
   local:
-    path: /opt/easegress
+    path: {{ $.Values.cluster.volumeLocalPath }}
```
suchen-sci, line 16: suggest to remove the `$` sign and use `{{ .Values.cluster.volumeLocalPath }}` instead.

codering: In the range block,  `{{ .Values.xxx }}` doesn't work.  I change to define another varuable.

suchen-sci (approved):

helm-charts/easegress/templates/PersistentVolume.yaml
suchen-sci: Cool, lesson learned. Look good to me.

PR #1116, opened 2023-10-27
Add filter RedirectorV2

xxx7xxxx:
There's another feature left, which is ReplacePrefixMatch. Because there's no natural way to get the matching prefix in the router. So in the filter RedirectorV2 there's no way to get the real matching path prefix. It would be an intermediate change to httpserver/mux to do it. So let's hold it for the time being.

### DIFF
--- a/docs/07.Reference/7.02.Filters.md
+++ b/docs/07.Reference/7.02.Filters.md
@@ -70,9 +70,12 @@
 - [Redirector](#redirector)
   - [Configuration](#configuration-22)
   - [Results](#results-22)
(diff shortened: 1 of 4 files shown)


suchen-sci (approved):

suchen-sci:
It may be a little tricky. `HTTPServer` and `Pipeline` communicate with each other through `Context`. So we may add some specially data on `Context`? For example, some `HTTPServer` information like path prefix or path regexp etc.

xxx7xxxx:
Sure I'm doing right that, will propose it in next PR.

PR #1128, opened 2023-11-01
Expose route for context

xxx7xxxx:
Have tried to design a universal exposed route type system which is complicated and unnecessary, so I turned to the minimum way to implement the current requirement.

Will try to update Gateway API to v1.0.0(GA), there's tons of broken dependencies to fix...

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -23,7 +23,6 @@ import (
 	"runtime/debug"
 
 	"github.com/megaease/easegress/v2/pkg/logger"
-	"github.com/megaease/easegress/v2/pkg/object/httpserver/routers"
 	"github.com/megaease/easegress/v2/pkg/protocols"
 	"github.com/megaease/easegress/v2/pkg/tracing"
 )
@@ -74,7 +73,7 @@ type Context struct {
 
 	activeNs string
 
-	route     routers.Route
+	route     protocols.Route
 	requests  map[string]*requestRef
 	responses map[string]*responseRef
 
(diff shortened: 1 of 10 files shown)


pkg/filters/redirectorv2/redirectorv2.go
```diff
@@ -76,7 +77,6 @@ type (
 	Spec struct {
 		filters.BaseSpec `json:",inline"`
 
-		PathPrefix                       *string `json:"pathPrefix,omitempty"`
```
suchen-sci, line 79: Please update the doc in `7.02.filters.md`...

suchen-sci (dismissed):

pkg/filters/redirectorv2/redirectorv2.go
xxx7xxxx: Sure, updated

pkg/context/context.go
```diff
@@ -92,6 +94,16 @@ func New(span *tracing.Span) *Context {
 	return ctx
 }
 
+// SetRoute sets the route.
+func (ctx *Context) SetRoute(route routers.Route) {
```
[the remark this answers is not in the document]
xxx7xxxx: Updated

pkg/filters/redirectorv2/redirectorv2.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"github.com/megaease/easegress/v2/pkg/context"
 	"github.com/megaease/easegress/v2/pkg/filters"
+	httprouters "github.com/megaease/easegress/v2/pkg/object/httpserver/routers"
```
suchen-sci, line 28: Looks good to me. Just a little advice. How about add a type alias in `github.com/megaease/easegress/v2/pkg/protocols/httpprot` say:

```
var Route = github.com/megaease/easegress/v2/pkg/object/httpserver/routers.Route
```
then 
```
prefix := route.(httpprot.Route).GetPathPrefix()
```
The code produce exactly same result. But for future developers, it may a little bit easier for them to find right interface?

xxx7xxxx: That will cause an import cycle.

suchen-sci (approved):

PR #1132, opened 2023-11-07
Support cookie hash

### LINKED ISSUE #368: Load balancing by cookie
Sometimes we might want the same instance to serve all request from a user, during the whole session. This way all user or session specific data is loaded and cached to only one instance.

The reverse Proxy currently supports [load balancing](https://github.com/megaease/easegress/blob/main/doc/filters.md#proxyLoadBalance) using `roundRobin`, `random`, `weightedRandom`, `ipHash` or `headerHash` strategies. `ipHash` and `headerHash` enables routing clients to specific instances using ip or a header field as hashing key.  It would be nice if one could use `cookie` also to choose the instance, eg. `cookieHash` load balancing strategy.

While this can already be achieved by using `headerHash` and specifying Cookie/Set-Cookie header, 
```yaml
loadBalance:
  policy: headerHash
  headerHashKey: Set-Cookie
```
it would be more convenient to only specify
```yaml
loadBalance:
  policy: cookieHash
```
Then later in the future, to go even further with `cookieHash`, the cookie could be generated and added to response, when user sends the first request.

xxx7xxxx:
Fix #368

### DIFF
--- a/pkg/filters/proxies/loadbalance.go
+++ b/pkg/filters/proxies/loadbalance.go
@@ -112,7 +112,7 @@ func (glb *GeneralLoadBalancer) Init(
 		case LoadBalancePolicyHeaderHash:
 			lbp = &HeaderHashLoadBalancePolicy{spec: glb.spec}
 		case LoadBalancePolicyCookieHash:
-			lbp = &HeaderHashLoadBalancePolicy{spec: &LoadBalanceSpec{HeaderHashKey: "Set-Cookie"}}
+			lbp = &HeaderHashLoadBalancePolicy{spec: &LoadBalanceSpec{HeaderHashKey: "Cookie"}}
 		default:
 			logger.Errorf("unsupported load balancing policy: %s", glb.spec.Policy)
 			lbp = &RoundRobinLoadBalancePolicy{}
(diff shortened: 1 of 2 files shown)


pkg/filters/proxies/loadbalance.go
```diff
@@ -108,6 +111,8 @@ func (glb *GeneralLoadBalancer) Init(
 			lbp = &IPHashLoadBalancePolicy{}
 		case LoadBalancePolicyHeaderHash:
 			lbp = &HeaderHashLoadBalancePolicy{spec: glb.spec}
+		case LoadBalancePolicyCookieHash:
+			lbp = &HeaderHashLoadBalancePolicy{spec: &LoadBalanceSpec{HeaderHashKey: "Set-Cookie"}}
```
suchen-sci, line 115: `Set-Cookie` is in response header that server send to client. In this place, client sends requests to server. I think maybe `Cookie` header is more appropriate. 

As show in https://docs.oracle.com/en- us/iaas/Content/Balance/Reference/sessionpersistence.htm#cook__HowItWorks

```
The Load Balancer service calculates a hash of the configured cookie and other request parameters, 
and sends that value to the client in a cookie. 
The value stored in the cookie enables the service to route subsequent client requests to the correct backend server. 
If your backend servers change any of the defined cookies, 
the service recomputes the cookie's value and resends it to the client.
```

My understand for this is that load balancer use `Set-Cookie` header to set some information to client. And in later requests, get this information from `Cookie` header and use this information to route to correct server? 

And for cookie hash, do we support to route based on a specific cookie or route based on all cookies? When use `header.Get()` is always return first cookie...

xxx7xxxx: Updated

suchen-sci (approved):

Issue #884, opened 2022-12-28
Misuse of jsonschema in specs. 


suchen-sci:
Currently, we use jsonschema in Spec this way:
```go
type Spec struct {
    Field1  string  `json:"field1" jsonschema:"required"`
    Field2  string  `json:"field2" jsonschema:"omitempty,enum=1,enum=2"`
```

We want in our yaml file must contain `field1` and may or may not contain `field2`, if contain `field2` it must be one of ["1", "2"]. 
But in practice, I find write jsonschema in this way not work as expected. 

The right way should be like:
```go
type Spec struct {
    Field1  *string  `json:"field1" jsonschema:"required"`
    Field2  string  `json:"field2,omitempty" jsonschema:"omitempty,enum=1,enum=2"`
```
For `fileld1`, the reason is simple, `field1` is required, and default value of *stirng nil is not acceptable (default value of string "" is acceptable in previous case). 

For `field2` we need set `omitempty` in `json` part too, only in `jsonschema` part is not good enough.

I think we should consider to update these influenced Specs... Any idea here?

PR #1136, opened 2023-11-13
Fix misuse of omitempty in jsonschema

xxx7xxxx:
Try to fix the massive misuse of `omitempty` in https://github.com/megaease/easegress/issues/884 . Now the fields without specifying `omitempty` would be required by default.

As for the required misuse, I don't think there is a simple and brute solution like `emitempty`. Since `required` needs to change the type to its pointer type, which would cause a lot of huge changes. Another way is to change the vendor to do it, which will bring more complexity. I suggest letting it be and changing it in the next development step by step.


pkg/object/httpserver/spec.go
```diff
@@ -32,38 +32,38 @@ import (
...
-		IPFilter *ipfilter.Spec `json:"ipFilter,omitempty" jsonschema:"omitempty"`
-		Rules    routers.Rules  `json:"rules" jsonschema:"omitempty"`
+		IPFilter *ipfilter.Spec `json:"ipFilter,omitempty"`
+		Rules    routers.Rules  `json:"rules,omitempty"`
 
-		GlobalFilter string `json:"globalFilter,omitempty" jsonschema:"omitempty"`
+		GlobalFilter string `json:"globalFilter,omitempty"`
 
-		AccessLogFormat string `json:"accessLogFormat" jsonshema:"omitempty"`
+		AccessLogFormat string `json:"accessLogFormat,omitempty" jsonshema:"omitempty"`
```
suchen-sci, line 66:
```suggestion
		AccessLogFormat string `json:"accessLogFormat,omitempty"`
```

pkg/object/nacosserviceregistry/nacosserviceregistry.go
```diff
@@ -85,17 +85,17 @@ type (
...
 	// ServerSpec is the server config of Nacos.
 	ServerSpec struct {
-		Scheme      string `json:"scheme" jsonschema:"omitempty,enum=http,enum=https"`
-		ContextPath string `json:"contextPath" jsonschema:"omitempty"`
-		IPAddr      string `json:"ipAddr" jsonschema:"required"`
-		Port        uint16 `json:"port" jsonschema:"required"`
+		Scheme      string `json:"scheme,omitempty" jsonschema:"enum=http,enum=https"`
+		ContextPath string `json:"contextPath,omitempty"`
+		IPAddr      string `json:"ipAddr,omitempty" jsonschema:"required"`
+		Port        uint16 `json:"port,omitempty" jsonschema:"required"`
```
suchen-sci, line 98:
```suggestion
		IPAddr      string `json:"ipAddr" jsonschema:"required"`
		Port        uint16 `json:"port" jsonschema:"required"`
```

suchen-sci (dismissed):

suchen-sci:
@xxx7xxxx looks good to me, just two little typo...

suchen-sci (approved):

PR #1140, opened 2023-11-21
Support easegress filters and resilience in k8s gateway via extensionRef 

suchen-sci:


pkg/object/gatewaycontroller/translator.go
```diff
@@ -97,6 +114,15 @@ func (b *pipelineSpecBuilder) jsonConfig() string {
 	if b.proxy != nil {
 		b.proxy.BaseSpec.MetaSpec.Name = "proxy"
 		b.proxy.BaseSpec.MetaSpec.Kind = httpproxy.Kind
+		for i := range b.proxy.Pools {
+			for _, r := range b.resilience {
```
xxx7xxxx, line 118: Don't we need to support retry?

xxx7xxxx (approved):

PR #1155, opened 2023-12-02
Use Validate Hook to control the total count of AutoCertManager

xxx7xxxx:
Based on the last pr #1144

### DIFF
--- a/pkg/object/autocertmanager/autocertmanager.go
+++ b/pkg/object/autocertmanager/autocertmanager.go
@@ -422,11 +422,21 @@ func (acm *AutoCertManager) HandleHTTP01Challenge(w http.ResponseWriter, r *http
 	w.Write(data)
 }
 
-func GetGlobalAutoCertManager() *AutoCertManager {
-	entity, exists := supervisor.GetGlobalSuper().GetSystemController(Kind)
-	if !exists {
-		return nil
+func GetGlobalAutoCertManager() (*AutoCertManager, bool) {
+	var acm *AutoCertManager
+
+	supervisor.GetGlobalSuper().WalkControllers(func(controller *supervisor.ObjectEntity) bool {
+		if controller.Spec().Kind() == Kind {
+			acm = controller.Instance().(*AutoCertManager)
+			return false
+		}
+
+		return true
+	})
+
+	if acm == nil {
+		return nil, false
 	}
 
-	return entity.Instance().(*AutoCertManager)
+	return acm, true
 }
--- a/pkg/object/autocertmanager/autocertmanager.go
+++ b/pkg/object/autocertmanager/autocertmanager.go
@@ -38,7 +38,7 @@ import (
 
 const (
 	// Category is the category of AutoCertManager.
-	Category = supervisor.CategorySystemController
+	Category = supervisor.CategoryBusinessController
 
 	// Kind is the kind of AutoCertManager.
 	Kind = "AutoCertManager"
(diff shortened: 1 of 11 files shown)


pkg/object/httpserver/spec.go
```diff
@@ -136,8 +136,11 @@ func (spec *Spec) tlsConfig() (*tls.Config, error) {
 		Certificates: certificates,
 		NextProtos:   []string{"acme-tls/1"},
 	}
+
+	acm := autocertmanager.GetGlobalAutoCertManager()
+
 	tlsConf.GetCertificate = func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
-		return autocertmanager.GetCertificate(chi, !spec.AutoCert /* tokenOnly */)
+		return acm.GetCertificate(chi, !spec.AutoCert /* tokenOnly */)
```
localvar, line 143: could `acm` be nil here?

pkg/supervisor/supervisor.go
```diff
@@ -137,6 +141,26 @@ func (s *Supervisor) initSystemControllers() {
...
+		panic(err)
+	}
+
+	// NOTE: The spec is already in cluster.
+	if value != nil {
+		return
+	}
+
+	err = s.cls.Put(s.cls.Layout().ConfigObjectKey(spec.Name()),
+		spec.JSONConfig())
```
localvar, line 161: per Go style guide.

```suggestion
	err = s.cls.Put(s.cls.Layout().ConfigObjectKey(spec.Name()), spec.JSONConfig())
```

pkg/object/httpserver/mux.go
```diff
@@ -204,8 +204,15 @@ func (m *mux) ServeHTTP(stdw http.ResponseWriter, stdr *http.Request) {
 	// don't know which HTTP server listen on this port (consider there's an
 	// nginx sitting in front of Easegress), so all HTTP servers need to
 	// handle HTTP-01 challenges.
+	acm := autocertmanager.GetGlobalAutoCertManager()
+	if acm == nil {
+		logger.Errorf("BUG: autocert manager not found")
+		stdw.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
```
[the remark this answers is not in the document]
xxx7xxxx: updated

pkg/object/httpserver/spec.go
xxx7xxxx: In real running mode, it won't be, but I add a line to do defensive programming.

pkg/supervisor/supervisor.go
xxx7xxxx: updated.

pkg/object/httpserver/spec.go
xxx7xxxx: We chose to make AutoCertManager to BusinessController because its properties do not seem like a system controller.

pkg/object/autocertmanager/autocertmanager.go
```diff
@@ -55,13 +53,34 @@ var aliases = []string{
...
+		return true
+	})
+
+	if len(acms) >= 1 {
+		return fmt.Errorf("only one AutoCertManager is allowed, existed: %v", acms)
+	}
+
+	return nil
+}
+
```
suchen-sci, line 83: Do we need to add a lock for this hook? I mean in a really specially case, if two users create `AutoCertManger` at same time, so they may call this function at same time, then they may both create it successfully. 

Of course, in practice, it may never happen, usually only one administrator creates or deletes Easegress objects. Just some defensive programming advice.

xxx7xxxx: Your concern has been fixed in API layer, we call validate hooks within a global lock. You can see the change in api/object.go

pkg/object/httpserver/spec.go
```diff
@@ -136,8 +136,16 @@ func (spec *Spec) tlsConfig() (*tls.Config, error) {
...
+
+	if spec.AutoCert {
+		acm, exists := autocertmanager.GetGlobalAutoCertManager()
+		if !exists {
+			return nil, fmt.Errorf("BUG: autocert manager is not initialized")
+		}
+
+		tlsConf.GetCertificate = func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			return acm.GetCertificate(chi, !spec.AutoCert /* tokenOnly */)
+		}
```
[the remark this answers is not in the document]
xxx7xxxx: No need to use atomic.Value because it is called infrequently. But you're right we need to get acm in real time in `tlsConf.GetCertificate`.

pkg/object/autocertmanager/autocertmanager.go
suchen-sci: see that. you are right

pkg/object/httpserver/spec.go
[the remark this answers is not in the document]
suchen-sci: `tlsConf. GetCertificate ` is used to get certificate based on client hello message. When a client connect to `https` server, they send a client hello message. And golang use this client hello info to get certificate from autocertmanager. I am not sure if golang use cache for this. If not, it will be called every time a client connect to server. That can be a lot...

Anyway, `atomic.Value` brings no harm at all. Only benefits of efficiency.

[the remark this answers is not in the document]
xxx7xxxx: Updated

localvar (approved):

Issue #1163, opened 2023-12-07
[Question]: New filter run panic


cyrnicolase:
I developed a new filter, it's really simple. I used `egbuilder` init one, and then add some log. 

### My filter code 
```go
package echo

import (
        "fmt"

        egCtx "github.com/megaease/easegress/v2/pkg/context"
        "github.com/megaease/easegress/v2/pkg/filters"
)


const (
        resultEchoFail = "EchoFail"
)

var Kind = &filters.Kind{
        CreateInstance: func(spec filters.Spec) filters.Filter {
                return &Echo{spec: spec.(*Spec)}
        },
        DefaultSpec: func() filters.Spec {
                return &Spec{}
        },
        Description: "Echo is my first filter plugin",
        Name:        "Echo",
        Results:     []string{resultEchoFail},
}

type Echo struct {
        spec *Spec
}

type Spec struct {
        filters.BaseSpec `json:",inline"`
}

var _ filters.Filter = (*Echo)(nil)
var _ filters.Spec = (*Spec)(nil)

func (s *Spec) Validate() error {
        return nil
}

func (e *Echo) Name() string {
    return "Echo"
}

func (e *Echo) Kind() *filters.Kind {
        return Kind
}

func (e *Echo) Spec() filters.Spec {
        return e.spec
}

func (e *Echo) Init() {
    fmt.Println("Call Echo Init func")
}

// Inherit inherits previous filter.
func (e *Echo) Inherit(previousGeneration filters.Filter) {
    fmt.Println("Call Echo Inherit func")
}

// Handle handles the request.
func (e *Echo) Handle(ctx *egCtx.Context) string {
    fmt.Println("Hello my first filter")

        return ""
}

// Status returns the status of the filter.
func (e *Echo) Status() interface{} {
    fmt.Println("Call Echo status func")
        return nil
}

// Close closes the filter.
func (e *Echo) Close() {}
```
### My configuration
```yaml
name: hs
kind: HTTPServer
port: 10180
http3: false
https: false
keepAlive: true
keepAliveTimeout: 60s
maxConnections: 1024
clientMaxBodySize: 1048576
rules:
  - host: example.org
    paths:
      - backend: pl.example.org

---

name: pl.example.org
kind: Pipeline
flow:
  - filter: echo
  - filter: proxy
filters:
  - name: echo
    kind: Echo

  - name: proxy
    kind: Proxy
    pools:
      - servers:
        - url: http://127.0.0.1:8181

```

### I get a panic
[...]

PR #1164, opened 2023-12-08
fix egbuilder bug, add integration tests for egbuilder

suchen-sci:
As described in issue #1163 , the generated code by using`egbuilder` contains a bug. This pr fix this bug and also add more integration tests for it.

### DIFF
--- a/build/test/client.go
+++ b/build/test/client.go
@@ -26,14 +26,18 @@ import (
 	"strings"
 )
 
-func egctlCmd(args ...string) *exec.Cmd {
+func egctlWithServer(server string, args ...string) *exec.Cmd {
 	egctl := os.Getenv("EGCTL")
 	if egctl == "" {
 		egctl = "egctl"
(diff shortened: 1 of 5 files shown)


build/test/client.go
```diff
@@ -26,16 +26,20 @@ import (
...
 		egctl = "egctl"
 	}
 	cmd := exec.Command(egctl, args...)
-	cmd.Args = append(cmd.Args, "--server", "http://127.0.0.1:12381")
+	cmd.Args = append(cmd.Args, "--server", server)
 	return cmd
 }
 
+func egctlCmd(args ...string) *exec.Cmd {
+	return egctlWithServer("http://127.0.0.1:12381", args...)
```
xxx7xxxx, line 40: Should we use the default option `http://127.0.0.1:2381` in server.

xxx7xxxx (approved):

Issue #1166, opened 2023-12-15
[Question]: Websocket handshake fail


cyrnicolase:
### This is not Bug report or Feature request

- [X] I confirm that this is not bug report or feature request.

### Is this question answered in [the documentation](https://github.com/megaease/easegress/blob/main/doc/README.md)?

- [X] I have searched the documentation

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Your question

Where I use easegress reverse proxy my weboscket service.  I get an error like this

```
ERROR   httpproxy/wspool.go:213 websocketproxy#wsproxy#main: dial to ws://10.0.0.114 failed: failed to WebSocket dial: expected handshake response status code 101 but got 404
```

## Websocket pipeline configure
```yaml
name: joyparty-pipeline-wsproxy-k8s
kind: Pipeline
filters:
  - name: wsproxy
    kind: WebSocketProxy
    serverMaxBodySize: -1
    pools:
      - servers:
        - url: ws://10.0.0.113          # my k8s cluster instance
        - url: ws://10.0.0.114
        - url: ws://10.0.0.117
        loadBalance:
          policy: roundRobin
```

## Architecture

```
client --> easegress --> k8s --> pod[backend]
```

If I don't use k8s, the weboscket reverse proxy is well .
```
client --> easegress --> backend(websocket server)    # this is ok 
```

suchen-sci:
What happens if `client --> k8s --> pod[backend]`? 
The error message seems from backend?

cyrnicolase:
> What happens if `client --> k8s --> pod[backend]`? The error message seems from backend?

`client --> k8s --> pod[backend]` works well.

suchen-sci:
Hi, if `client --> k8s --> pod[backend]` works and `client --> easegress --> backend` works. 
It seems some network problem from easegress to k8s? 
But `websocketproxy` can find the backend and got a status code of 404 (not like a network problem). 

Emmm, I guess we can use some tool like wireshark to catch some packets. To see what's the difference between `easegresss -> k8s` and `client -> k8s`. If they send the same packet, they should get same result.

Make sure you client is on same server of easegress. You can add this filter to your pipeline to diff the requests from your client and easegress.

```
kind: RequestAdaptor
name: request-adaptor
header:
  add: 
    X-Easegress: Test
```  

If they send different packets, we can find the problem of easegress.

For the information you have provided, I can't dig more...

xxx7xxxx:
[...]

PR #1168, opened 2023-12-19
fix wsproxy keepHost bug

cyrnicolase:
#1166


suchen-sci:
Hi, can you also help update this doc `https://github.com/megaease/easegress/blob/main/docs/02.Tutorials/2.6.Websocket.md` for future users to avoid this problem? Maybe add some simple example to it?

suchen-sci (approved):

Issue #1170, opened 2023-12-27
[Bug]: easegress-server --signal-upgrade=true panic


cyrnicolase:
- Easegress v2.6.4
- 22.04.1-Ubuntu x86_64

---

When I reload the easegress-server with command `easegress-server --signal-upgrade=true`. It will panic.

```
panic: close of closed channel

goroutine 1 [running]:
github.com/megaease/easegress/v2/pkg/api.(*dynamicMux).close(...)
        github.com/megaease/easegress/v2/pkg/api/dynamicmux.go:139
github.com/megaease/easegress/v2/pkg/api.(*Server).Close(0xc000401900, 0x0?)
        github.com/megaease/easegress/v2/pkg/api/server.go:131 +0x111
github.com/megaease/easegress/v2/cmd.RunServer()
        github.com/megaease/easegress/v2/cmd/server.go:142 +0x78a
main.main()
        github.com/megaease/easegress/v2/cmd/server/main.go:27 +0x17
```

suchen-sci:
Hi, thanks for you report, we will check what happens here...

xxx7xxxx:
Fixed

PR #1190, opened 2024-01-06
signal-upgrade close of closed channel

suchen-sci:
As discussed in issue #1170 , when use `signal-upgrade` to graceful update easegress, it will cause a panic with message "close of closed channel". Although it not influence the usage of easegress, but we'd better to fix it.

After some carefully digging, here is what happens. When use `signal-upgrade`, `easegress-server` receive a user signal, then graceful upgrade. What happens here is `easegress-server` gracefully closes api-server and starts a new process (like fork, start a new process with original environment and configuration). And when the new process is started, it will send a signal to old process to terminate, which will close the api-server again, cause the panic with message "close of closed channel".
Because both two close of api-server happens in original process, it won't influence the new process.

But since this may influence the core part of easegress. Any fix with a lot of code change may cause potential bugs. So, in this pr, I fix it by making api-server recloseable.

### DIFF
--- a/pkg/api/dynamicmux.go
+++ b/pkg/api/dynamicmux.go
@@ -139,6 +139,11 @@ func (m *dynamicMux) reloadAPIs() {
 }
 
 func (m *dynamicMux) close() {
+	// make sure to close channel only once.
+	// when use "signal-upgrade", easegress will start a new process gracefully,
+	// which may cause the old process be closed twice.
+	// here we use sync.Once to make sure the channel is closed only once.
+	// more discussion here: https://github.com/easegress-io/easegress/issues/1170
 	m.closeOnce.Do(func() {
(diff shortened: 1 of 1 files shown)


pkg/api/dynamicmux.go
```diff
@@ -136,5 +139,7 @@ func (m *dynamicMux) reloadAPIs() {
 }
 
 func (m *dynamicMux) close() {
-	close(m.done)
+	m.closeOnce.Do(func() {
```
xxx7xxxx, line 142: It's rare to support reclosable in Easegress, so it's better to add a comment here.

suchen-sci: updated! please check

xxx7xxxx (approved):

PR #1194, opened 2024-01-11
implement 'urlQueryEscape' and 'urlQueryUnescape' template functions in builtin extraFuncs

KevinQiangK:
This commit adds the 'urlQueryEscape' and 'urlQueryUnescape' functions to the template's extra functions ('extraFuncs'). These functions provide URL encoding and decoding capabilities.

### DIFF
--- a/docs/07.Reference/7.02.Filters.md
+++ b/docs/07.Reference/7.02.Filters.md
@@ -2071,6 +2071,10 @@ package, and extra functions defined by Easegress:
   also an object, its type must also be `map[string]interface{}`.
 * **jsonEscape**: escape a string so that it can be used as the key or value
   in JSON text.
+* **urlQueryEscape**: escapes the string so that it can be safely placed inside
+  a URL query, equivalent to the `urlquery` template function.
+* **urlQueryUnescape**: performs the inverse transformation of `urlQueryEscape`,
+  decoding a URL-encoded string back to its origin form.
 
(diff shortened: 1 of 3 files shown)


docs/07.Reference/7.02.Filters.md
```diff
@@ -2071,6 +2071,10 @@ package, and extra functions defined by Easegress:
   also an object, its type must also be `map[string]interface{}`.
 * **jsonEscape**: escape a string so that it can be used as the key or value
   in JSON text.
+* **urlQueryEscape**: escapes the string so that it can be safely placed inside
+  a URL query, equivalent to the `urlquery` template function.
```
suchen-sci, line 2075: Hi, what's the difference between `urlQueryEscape` and `urlquery`? 
And could you please tell me why we need to add `urlQueryEscape` if there are already `urlquery`?

KevinQiangK: urlquery is a built-in template function in the Go template package, and it has the same implementation as urlQueryEscape. The reason for adding both urlQueryEscape and urlQueryUnescape functions simultaneously serves two purposes:

1. To provide a clearer expression of functionality, as these functions are implemented using the queryEscape and queryUnescape methods from the url package.
2. To maintain consistency within Easegress by using symmetric function names.

suchen-sci (approved):

PR #1197, opened 2024-01-19
support egctl get describe different namespace

suchen-sci:
In Easegress, when using `IngressController`, `MeshController`, or `GatewayController`, resources such as `HTTPServers` and `Pipelines` are created in separate namespaces to prevent conflicts. To access these resources, you can utilize the `--namespace` or `--all-namespaces` argument for `egctl get` or `egctl describe` command.

> The use of these arguments is relevant only when employing `IngressController`, `MeshController`, or `GatewayController`. Easegress does not support creating resources in different namespaces by using apis or `egctl` command.

This pr also fix some previous bugs about get status. In easegress, status and config of traffic object are put into different namespace. 
For example, autocertmanager in default namespace, then its status is in default namespace too. But for httpserver (which is traffic object), if httpserver is in default namespace, then its status is in eg-traffic-default. When store status, traffic controller use a special prefix for it.

### DIFF
--- a/cmd/client/resources/object.go
+++ b/cmd/client/resources/object.go
@@ -22,8 +22,10 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -34,6 +36,12 @@ import (
 	"github.com/megaease/easegress/v2/pkg/util/codectool"
 	"github.com/spf13/cobra"
 )
+
+// GetObjectFlags contains the flags for get objects.
+type GetObjectFlags struct {
+	Namespace    string
+	AllNamespace bool
+}
 
 // ObjectAPIResources returns the object api resources.
 func ObjectAPIResources() ([]*api.APIResource, error) {
@@ -54,23 +62,29 @@ func defaultObjectNameSpace() string {
 	return cluster.TrafficNamespace(cluster.NamespaceDefault)
 }
 
-func httpGetObject(name string) ([]byte, error) {
-	url := func(name string) string {
+func httpGetObject(name string, flags *GetObjectFlags) ([]byte, error) {
+	objectURL := func(name string) string {
 		if len(name) == 0 {
 			return makePath(general.ObjectsURL)
 		}
 		return makePath(general.ObjectItemURL, name)
 	}(name)
-	return handleReq(http.MethodGet, url, nil)
+	if flags != nil {
+		values := url.Values{}
+		values.Add("namespace", flags.Namespace)
+		values.Add("all-namespaces", strconv.FormatBool(flags.AllNamespace))
+		objectURL = fmt.Sprintf("%s?%s", objectURL, values.Encode())
+	}
+	return handleReq(http.MethodGet, objectURL, nil)
 }
 
 // GetAllObject gets all objects.
-func GetAllObject(cmd *cobra.Command) error {
+func GetAllObject(cmd *cobra.Command, flags *GetObjectFlags) error {
 	getErr := func(err error) error {
 		return general.ErrorMsg(general.GetCmd, err, "resource")
 	}
 
-	body, err := httpGetObject("")
+	body, err := httpGetObject("", flags)
 	if err != nil {
 		return getErr(err)
 	}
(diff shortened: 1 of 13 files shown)


cmd/client/resources/object.go
```diff
@@ -47,30 +60,39 @@ func ObjectAPIResources() ([]*api.APIResource, error) {
 	if err != nil {
 		return nil, err
 	}
+	globalAPIResources = res
 	return res, nil
 }
 
-func defaultObjectNameSpace() string {
-	return cluster.TrafficNamespace(cluster.NamespaceDefault)
+// trafficObjectStatusNamespace return namespace of traffic object status.
```
xxx7xxxx, line 67: I can't understand this comment, what does it mean? And the following line also contains confusing grammar.

pkg/api/cluster.go
```diff
@@ -116,8 +116,18 @@ func (s *Server) _deleteObject(name string) {
 	}
 }
 
-func (s *Server) _getStatusObject(name string) map[string]interface{} {
-	prefix := s.cluster.Layout().StatusObjectPrefix(cluster.TrafficNamespace(cluster.NamespaceDefault), name)
+// _getStatusObject returns the status object of the name.
```
xxx7xxxx, line 119:
```suggestion
// _getStatusObject returns the status object with the specified name.
```

pkg/object/trafficcontroller/trafficcontroller.go
```diff
@@ -714,3 +706,24 @@ func (tc *TrafficController) Close() {
 		logger.Infof("delete namespace %s", name)
 	}
 }
+
+// ListAllNamespace lists all namespaces pipelines and traffic gates.
```
xxx7xxxx, line 710:
```suggestion
// ListAllNamespace lists pipelines and traffic gates in all namespaces.
```

pkg/object/trafficcontroller/trafficcontroller.go
```diff
@@ -99,12 +97,6 @@ var _ easemonitor.Metricer = (*TrafficObjectStatus)(nil)
 
 func init() {
 	supervisor.Register(&TrafficController{})
-	api.RegisterObject(&api.APIResource{
```
xxx7xxxx, line 102: Why delete this registration?

pkg/api/object.go
```diff
@@ -387,3 +431,66 @@ func (s *Server) listObjectAPIResources(w http.ResponseWriter, r *http.Request)
...
+		return nil
+	}
+	tc, ok := entity.Instance().(*trafficcontroller.TrafficController)
+	if !ok {
+		return nil
+	}
+	return tc
+}
+
+func (s *Server) _listAllNamespace() map[string][]*supervisor.Spec {
```
xxx7xxxx, line 447:
```suggestion
func (s *Server) _listAllNamespaces() map[string][]*supervisor.Spec {
```

pkg/api/object.go
```diff
@@ -387,3 +431,66 @@ func (s *Server) listObjectAPIResources(w http.ResponseWriter, r *http.Request)
...
+		specs := make([]*supervisor.Spec, 0, len(objects))
+		for _, o := range objects {
+			specs = append(specs, o.Spec())
+		}
+		res[namespace] = specs
+	}
+	return res
+}
+
+func (s *Server) _listNamespace(ns string) []*supervisor.Spec {
```
xxx7xxxx, line 464:
```suggestion
func (s *Server) _listNamespaces(ns string) []*supervisor.Spec {
```

cmd/client/resources/object.go
suchen-sci: for non-traffic objects, like autocertmanager, it is in namespace default, and its status is in namespace default. 

but for traffic objects, they are managed by traffic controller, and traffic controller use specially prefix for status. for example, if a pipeline in namesapce default, then its status in namespace eg-traffic-default.

pkg/object/trafficcontroller/trafficcontroller.go
suchen-sci: cycle import, i move this registration to api pkg. it can't be solved by other way... i try it several times.

xxx7xxxx (approved):

PR #1218, opened 2024-02-21
Upgrade Go version to 1.21

xxx7xxxx:
Try to fix openssl/libcrypto3 vulnerability. After this merged, we could bump a new release.


build/package/Dockerfile.builder
```diff
@@ -1,2 +1,2 @@
-FROM golang:1.21.0-alpine
+FROM golang:1.21.7-alpine
```
localvar, line 1: when building the easegress docker image, we are always using the latest version of Go,.

```suggestion
FROM golang:1.22.0-alpine
```

suchen-sci (approved):

PR #1248, opened 2024-03-26
optim egctl get httpserver to show port and https

suchen-sci:


cmd/client/resources/object.go
```diff
@@ -272,61 +282,90 @@ func unmarshalPrintNamespaceMetaSpec(body []byte, filter func(*supervisor.MetaSp
...
+		}
 	}
+
 	for namespace, metas := range metas {
 		if namespace == DefaultNamespace {
 			continue
 		}
 		for _, meta := range metas {
-			table = append(table, []string{meta.Name, meta.Kind, namespace, getAgeFromMetaSpec(meta)})
+			if meta.Kind == "HTTPServer" {
```
zhao-kun, line 336: It is common for each object to have its representation, so a well-designed framework is needed to handle the difference.

suchen-sci: updated, please check

cmd/client/resources/spec.go
```diff
@@ -0,0 +1,124 @@
...
+// TableRow is the row type.
+type TableRow []string
+
+// WithNamespace adds the namespace to the row.
+func (row TableRow) WithNamespace(namespace string) TableRow {
+	return append(row, namespace)
+}
+
+// ObjectSpec is the interface of the object spec.
+type ObjectSpec interface {
```
zhao-kun, line 64: It's best to give it a meaningful name to intuitively describe its usage

PR #1511, opened 2026-03-25
feat: Adds Qdrant Vector Search Provider

Anush008:
## Description

This PR adds support for using [Qdrant](https://qdrant.tech/) as a vector search provider in Easegress.

### Testing

I've unit tested this integration against a local Qdrant instance.

### Setup

You can run Qdrant with 

```bash
docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant
```

The dashboard is accessible at http://localhost:6333/dashboard

The Go client uses the gRPC API that's available at `host=localhost` and `port=6334`.

Ref: https://qdrant.tech/documentation/interfaces/

### Documentation

I've drafted documentation for the integration to be added to https://qdrant.tech/documentation/platforms/.


xxx7xxxx (changes requested):
Need to fix unittests

PR #1518, opened 2026-05-26
fix(grpcproxy): do not degrade pool on gRPC application-level errors (fixes #1517)

### LINKED ISSUE #1517: GRPCProxy: biTransportProxy degrades backend pool on valid gRPC application errors (e.g. NOT_FOUND), causing subsequent calls to stall
## Summary

`biTransportProxy()` in `pkg/filters/proxies/grpcproxy/pool.go` marks the backend connection as `resultServerError` for **any** non-`io.EOF` error returned from the backend stream — including well-formed gRPC status errors like `NOT_FOUND` (code 5), `INVALID_ARGUMENT`, etc.

This causes the connection pool to degrade a healthy backend connection. The next request then blocks indefinitely on `pool.borrow()` because no `BorrowTimeout` is configured by default, until the outer context deadline fires.

## Root Cause

```go
// pkg/filters/proxies/grpcproxy/pool.go — biTransport()
case s2cErr := <-s2cErrChan:
    ctx.resp.SetTrailer(grpcprot.NewTrailer(proxyAsClientStream.Trailer()))
    if s2cErr != io.EOF {
        // BUG: also fires for valid gRPC status errors
        return serverPoolError{status.Convert(s2cErr), resultServerError}
    }
    return nil
```

A `*status.Status` error returned by `grpc.ClientStream.RecvMsg()` means the backend responded correctly over a **healthy connection** with an application-level error. It is not a transport failure. Returning `resultServerError` for it is incorrect.

## Impact

Any service behind GRPCProxy that legitimately returns a non-OK gRPC status for some requests triggers this bug:

1. **Call A** → backend returns `NOT_FOUND` → `biTransport` marks pool entry `resultServerError` → pool exhausted
2. **Call B** → `pool.borrow()` with no timeout → **blocks indefinitely**
3. After 30 seconds the outer gRPC deadline fires → `context.Canceled` → downstream caller receives a generic service error

This is reproducible with any backend that mixes successful responses with `NOT_FOUND` / `INVALID_ARGUMENT` / etc. in the same stream lifecycle.

## Fix

Use `status.FromError()` to distinguish transport failures from gRPC application responses:

```go
case s2cErr := <-s2cErrChan:
    ctx.resp.SetTrailer(grpcprot.NewTrailer(proxyAsClientStream.Trailer()))
    if s2cErr != io.EOF {
        // A valid gRPC status error means the backend communicated an
        // application-level result over a healthy connection.
        // Only degrade the pool for genuine transport failures.
        if _, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
            return nil
        }
        return serverPoolError{status.Convert(s2cErr), resultServerError}
    }
    return nil
```

## Affected Versions

The bug exists in all versions — the relevant code in `pool.go` has been unchanged since the initial gRPC proxy implementation (commit `c3f1a8c`, 2024-03-10). Confirmed present in v2.7.3 and the current `main` branch.

## Reproduction

1. Configure a GRPCProxy pool pointing to any gRPC service
2. Make two back-to-back calls where the first returns a non-OK status (e.g. `NOT_FOUND`)
3. Observe the second call hanging for the full context deadline duration

A fix with tests is available in PR #[to be linked].

akpradheeph:
## Problem

Fixes #1517.

`biTransportProxy()` in `pkg/filters/proxies/grpcproxy/pool.go` treats every non-`io.EOF` error from the backend stream as a transport failure and calls `svr.close(resultServerError)`, degrading the backend connection pool.

This fires incorrectly for **valid gRPC status errors** (e.g. `NOT_FOUND`, `INVALID_ARGUMENT`, `INTERNAL`) that the backend returned over a perfectly healthy connection. The result:

1. Call A → backend returns `NOT_FOUND` → pool entry marked `resultServerError` → pool exhausted
2. Call B → `pool.borrow()` with no `BorrowTimeout` → **blocks indefinitely**
3. Outer 30-second deadline fires → `context.Canceled` → caller gets a generic service error

## Root Cause

```go
// Before (buggy)
if s2cErr != io.EOF {
    return serverPoolError{status.Convert(s2cErr), resultServerError}  // fires for NOT_FOUND too
}
```

`grpc.ClientStream.RecvMsg()` returns a `*status.Status`-wrapped error for application-level responses. `status.FromError()` returns `ok=true` for these — they indicate the backend sent a valid gRPC response, not that the connection is broken.

## Fix

```go
// After
if s2cErr != io.EOF {
    // A valid gRPC status error means the backend communicated an
    // application-level result over a healthy connection.
    // Only degrade the pool for genuine transport failures.
    if _, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
        return nil
    }
    return serverPoolError{status.Convert(s2cErr), resultServerError}
}
```

## Changes

- **`pkg/filters/proxies/grpcproxy/pool.go`** — one `status.FromError()` guard in `biTransport()`
- **`pkg/filters/proxies/grpcproxy/pool_test.go`** — `TestBiTransportDoesNotDegradePoolOnGRPCAppError` covering:
  - `NOT_FOUND` → returns `nil` (no pool degradation)
  - `INVALID_ARGUMENT` → returns `nil`
  - `INTERNAL` → returns `nil`
  - `io.EOF` → returns `nil` (existing happy path)
  - plain transport error (`io.ErrUnexpectedEOF`) → returns `serverPoolError{resultServerError}` (unchanged)

## Testing

```
=== RUN   TestBiTransportDoesNotDegradePoolOnGRPCAppError
    --- PASS: .../NOT_FOUND_is_an_application_error (0.00s)
    --- PASS: .../INVALID_ARGUMENT_is_an_application_error (0.00s)
    --- PASS: .../INTERNAL_is_a_server-side_gRPC_error (0.00s)
    --- PASS: .../io.EOF_is_the_happy-path_completion (0.00s)
    --- PASS: .../plain_transport_error (0.00s)
PASS — full grpcproxy suite: 0 regressions
```

### DIFF
--- a/pkg/filters/proxies/grpcproxy/pool.go
+++ b/pkg/filters/proxies/grpcproxy/pool.go
@@ -373,7 +373,11 @@ func (sp *ServerPool) biTransport(ctx *serverPoolContext, proxyAsClientStream gr
 				// subsequent calls to stall waiting on pool.borrow(). Only return
 				// resultServerError for transport-level failures (non-gRPC errors)
 				// where the connection itself is broken.
-				if _, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
+				if st, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
+					// Preserve the backend's gRPC status so the caller receives the
+					// correct code (e.g. NOT_FOUND, INVALID_ARGUMENT). The connection
+					// itself is healthy so we still return nil to avoid pool degradation.
+					ctx.resp.SetStatus(st)
 					return nil
 				}
 				return serverPoolError{status.Convert(s2cErr), resultServerError}
--- a/pkg/filters/proxies/grpcproxy/pool.go
+++ b/pkg/filters/proxies/grpcproxy/pool.go
@@ -366,6 +366,16 @@ func (sp *ServerPool) biTransport(ctx *serverPoolContext, proxyAsClientStream gr
 			ctx.resp.SetTrailer(grpcprot.NewTrailer(proxyAsClientStream.Trailer()))
 			// c2sErr will contain RPC error from client code. If not io.EOF return the RPC error as server stream error.
 			if s2cErr != io.EOF {
+				// A well-formed gRPC status error means the backend communicated an
+				// application-level result (e.g. NOT_FOUND, INVALID_ARGUMENT) over a
+				// healthy connection. Marking the pool entry as resultServerError in
+				// that case degrades a perfectly usable backend connection and causes
+				// subsequent calls to stall waiting on pool.borrow(). Only return
+				// resultServerError for transport-level failures (non-gRPC errors)
+				// where the connection itself is broken.
+				if _, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
+					return nil
+				}
 				return serverPoolError{status.Convert(s2cErr), resultServerError}
 			}
 			return nil
--- a/pkg/filters/proxies/grpcproxy/pool_test.go
+++ b/pkg/filters/proxies/grpcproxy/pool_test.go
@@ -253,10 +253,10 @@ type biTransportClientStream struct {
 	recvErr error
 }
 
-func (m *biTransportClientStream) Header() (metadata.MD, error) { return metadata.New(nil), nil }
-func (m *biTransportClientStream) Trailer() metadata.MD         { return metadata.New(nil) }
-func (m *biTransportClientStream) CloseSend() error             { return nil }
-func (m *biTransportClientStream) Context() context.Context     { return context.Background() }
+func (m *biTransportClientStream) Header() (metadata.MD, error)  { return metadata.New(nil), nil }
+func (m *biTransportClientStream) Trailer() metadata.MD          { return metadata.New(nil) }
+func (m *biTransportClientStream) CloseSend() error              { return nil }
+func (m *biTransportClientStream) Context() context.Context      { return context.Background() }
 func (m *biTransportClientStream) SendMsg(msg interface{}) error { return nil }
 func (m *biTransportClientStream) RecvMsg(msg interface{}) error { return m.recvErr }
 
@@ -268,11 +268,11 @@ type biTransportServerStream struct {
 	done <-chan struct{}
 }
 
-func (m *biTransportServerStream) Context() context.Context      { return context.Background() }
-func (m *biTransportServerStream) SetHeader(md metadata.MD) error { return nil }
+func (m *biTransportServerStream) Context() context.Context        { return context.Background() }
+func (m *biTransportServerStream) SetHeader(md metadata.MD) error  { return nil }
 func (m *biTransportServerStream) SendHeader(md metadata.MD) error { return nil }
-func (m *biTransportServerStream) SetTrailer(md metadata.MD)      {}
-func (m *biTransportServerStream) SendMsg(msg interface{}) error  { return nil }
+func (m *biTransportServerStream) SetTrailer(md metadata.MD)       {}
+func (m *biTransportServerStream) SendMsg(msg interface{}) error   { return nil }
 func (m *biTransportServerStream) RecvMsg(msg interface{}) error {
 	<-m.done
 	return io.EOF
--- a/pkg/filters/proxies/grpcproxy/pool_test.go
+++ b/pkg/filters/proxies/grpcproxy/pool_test.go
@@ -287,29 +287,34 @@ func TestBiTransportDoesNotDegradePoolOnGRPCAppError(t *testing.T) {
 	sp := &ServerPool{}
 
 	cases := []struct {
-		name    string
-		err     error
-		wantNil bool
+		name     string
+		err      error
+		wantNil  bool
+		wantCode codes.Code // non-OK: assert spCtx.resp carries this status; OK means skip (e.g. io.EOF)
 	}{
 		{
-			name:    "NOT_FOUND is an application error — pool must not degrade",
-			err:     status.Error(codes.NotFound, "resource not found"),
-			wantNil: true,
+			name:     "NOT_FOUND is an application error — pool must not degrade",
+			err:      status.Error(codes.NotFound, "resource not found"),
+			wantNil:  true,
+			wantCode: codes.NotFound,
 		},
 		{
-			name:    "INVALID_ARGUMENT is an application error — pool must not degrade",
-			err:     status.Error(codes.InvalidArgument, "bad request"),
-			wantNil: true,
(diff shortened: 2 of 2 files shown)


pkg/filters/proxies/grpcproxy/pool.go
```diff
@@ -366,6 +366,16 @@ func (sp *ServerPool) biTransport(ctx *serverPoolContext, proxyAsClientStream gr
...
 			if s2cErr != io.EOF {
+				// A well-formed gRPC status error means the backend communicated an
+				// application-level result (e.g. NOT_FOUND, INVALID_ARGUMENT) over a
+				// healthy connection. Marking the pool entry as resultServerError in
+				// that case degrades a perfectly usable backend connection and causes
+				// subsequent calls to stall waiting on pool.borrow(). Only return
+				// resultServerError for transport-level failures (non-gRPC errors)
+				// where the connection itself is broken.
+				if _, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
+					return nil
```
xxx7xxxx, line 381: This keeps the pool healthy, but it also loses the status from the backend. `sp.handle` treats a nil error as success, and the gRPC server later returns `resp.Err()`, so a backend `NOT_FOUND` or `INVALID_ARGUMENT` can become an OK response. Please set `ctx.resp` from `s2cErr` before taking this path.

pkg/filters/proxies/grpcproxy/pool_test.go
```diff
@@ -241,3 +244,106 @@ name: grpcforwardproxy
...
+			spCtx := &serverPoolContext{
+				stdr: serverStream,
+				stdw: serverStream,
+				resp: grpcprot.NewResponse(),
+			}
+
+			result := sp.biTransport(spCtx, clientStream)
+
+			if tc.wantNil {
+				assert.Nil(t, result,
```
xxx7xxxx, line 343: This also needs to check the response status. As written, the test passes even if `NOT_FOUND` is turned into an OK response; asserting `spCtx.resp.GetStatus().Code()` for the gRPC status cases would catch that.

pkg/filters/proxies/grpcproxy/pool.go
akpradheeph: Good catch — fixed in 0e69ba3b.

Capturing the `*status.Status` return value from `status.FromError` and calling `ctx.resp.SetStatus(st)` before returning `nil`:

```go
if st, isGRPCStatus := status.FromError(s2cErr); isGRPCStatus {
    ctx.resp.SetStatus(st)
    return nil
}
```

This preserves the backend’s NOT_FOUND / INVALID_ARGUMENT / etc. on the response so `handle()` propagates it correctly to the caller.

pkg/filters/proxies/grpcproxy/pool_test.go
akpradheeph: Fixed in 0e69ba3b. Added a `wantCode codes.Code` field to each test case and assert `spCtx.resp.GetStatus().Code()` for every gRPC status variant:

```go
if tc.wantCode != codes.OK {
    assert.Equal(t, tc.wantCode, spCtx.resp.GetStatus().Code(),
        "backend gRPC status must be propagated to the response for %q", tc.name)
}
```

All three gRPC status cases (NOT_FOUND, INVALID_ARGUMENT, INTERNAL) now carry a `wantCode` assertion. The io.EOF happy-path leaves `wantCode` as `codes.OK` and skips the check. The test would have caught the silent-OK regression before this fix.

pkg/filters/proxies/grpcproxy/pool_test.go
```diff
@@ -241,3 +244,117 @@ name: grpcforwardproxy
...
+	grpc.ClientStream
+	recvErr error
+}
+
+func (m *biTransportClientStream) Header() (metadata.MD, error) { return metadata.New(nil), nil }
+func (m *biTransportClientStream) Trailer() metadata.MD         { return metadata.New(nil) }
+func (m *biTransportClientStream) CloseSend() error             { return nil }
+func (m *biTransportClientStream) Context() context.Context     { return context.Background() }
+func (m *biTransportClientStream) SendMsg(msg interface{}) error { return nil }
+func (m *biTransportClientStream) RecvMsg(msg interface{}) error { return m.recvErr }
```
xxx7xxxx: `gofmt` still catches this block. This should be the only change here.

```suggestion
func (m *biTransportClientStream) Header() (metadata.MD, error)  { return metadata.New(nil), nil }
func (m *biTransportClientStream) Trailer() metadata.MD          { return metadata.New(nil) }
func (m *biTransportClientStream) CloseSend() error              { return nil }
func (m *biTransportClientStream) Context() context.Context      { return context.Background() }
func (m *biTransportClientStream) SendMsg(msg interface{}) error { return nil }
func (m *biTransportClientStream) RecvMsg(msg interface{}) error { return m.recvErr }
```

pkg/filters/proxies/grpcproxy/pool_test.go
```diff
@@ -241,3 +244,117 @@ name: grpcforwardproxy
...
+type biTransportServerStream struct {
+	grpc.ServerStream
+	done <-chan struct{}
+}
+
+func (m *biTransportServerStream) Context() context.Context      { return context.Background() }
+func (m *biTransportServerStream) SetHeader(md metadata.MD) error { return nil }
+func (m *biTransportServerStream) SendHeader(md metadata.MD) error { return nil }
+func (m *biTransportServerStream) SetTrailer(md metadata.MD)      {}
+func (m *biTransportServerStream) SendMsg(msg interface{}) error  { return nil }
```
xxx7xxxx: Same `gofmt` alignment here.

```suggestion
func (m *biTransportServerStream) Context() context.Context        { return context.Background() }
func (m *biTransportServerStream) SetHeader(md metadata.MD) error  { return nil }
func (m *biTransportServerStream) SendHeader(md metadata.MD) error { return nil }
func (m *biTransportServerStream) SetTrailer(md metadata.MD)       {}
func (m *biTransportServerStream) SendMsg(msg interface{}) error   { return nil }
```

pkg/filters/proxies/grpcproxy/pool_test.go
akpradheeph: Fixed in 869121bc — ran `gofmt -w` on the file, which aligns both blocks as suggested.

pkg/filters/proxies/grpcproxy/pool_test.go
akpradheeph: Fixed in 869121bc — same `gofmt -w` pass covers this block too.

xxx7xxxx (approved):

